PR #2, opened 2020-09-08
Initial skeleton

levb:
First implementation

    - General plugin skeleton: commands, configuration
    - Skeleton Registry, Proxy, Subscription entities
    - Skeleton OnUserJoinedChannel implementation
    - Basic `Expander`


levb:
@hanzei @jfrerich Any idea what's wrong with the lint (git check)?

jfrerich:
@levb that's from `go mod tidy`.  Run and it to remove those unused packages.

server/cloudapps/expand.go
```diff
@@ -0,0 +1,166 @@
...
+}
+
+func NewExpander(mm *pluginapi.Client, configurator configurator.Configurator) *Expander {
+	return &Expander{
+		mm:           mm,
+		configurator: configurator,
+	}
+}
+
+func (e *Expander) Expand(expand *Expand, actingUserID, userID, channelID string) (expanded *Expanded, err error) {
```
larkox, line 63: Should expansions be scoped? i.e. Do we want to limit what certain apps can expand and how much can they expand it?

server/cloudapps/proxy.go
```diff
@@ -0,0 +1,76 @@
...
+	go func() {
+		defer pipew.Close()
+		encodeErr := json.NewEncoder(pipew).Encode(msg)
+		if encodeErr != nil {
+			// <><> TODO log
+			return
+		}
+	}()
+
+	u := path.Join(app.RootURL, "notify", string(SubjectUserJoinedChannel))
```
larkox, line 63: Should we force the path or leave the app to handle its own paths? e.g. An app may want to handle the wishes at `root/wish/certainWish`but handle all the notifications at `root/notifications`.

server/command/command.go
```diff
@@ -0,0 +1,102 @@
...
+
+func NewCommand(configurator configurator.Configurator, mm Mattermost) Command {
+	return &command{
+		Mattermost:   mm,
+		Configurator: configurator,
+	}
+}
+
+// Register should be called by the plugin to register all necessary commands
+func (c *command) Init(buildConfig *configurator.BuildConfig) error {
```
larkox, line 41: Since apps will be able to register commands, we should put that logic somewhere. Not sure if this package is the right place or not.

larkox (approved):
A few considerations and questions, but in general LGTM

server/cloudapps/on_user_joined_channel.go
```diff
@@ -0,0 +1,47 @@
...
+// See License for license information.
+
+package cloudapps
+
+import (
+	"github.com/mattermost/mattermost-server/v5/model"
+	"github.com/mattermost/mattermost-server/v5/plugin"
+)
+
+type UserJoinedChannelNotification struct {
```
jfrerich, line 11: The [Mattermost Cloud Apps](https://docs.google.com/document/d/1r52PPwVfcyCjOgnq5ulmrLdeOVgDSGuTbhqu9Cjq9GE/edit#heading=h.qksj0bkjfz31) describes these as Notification Subjects.  The list of subjects is a considerable number with the potential to grow.  I am wondering if we'd benefit from creating another package (subjects?) inside `cloudapps`?

jfrerich (approved):

server/cloudapps/placeholder_structs.go
```diff
@@ -0,0 +1,112 @@
...
+	Callable
+	Name    string
+	Title   string
+	Header  string
+	Footer  string
+	IconURL string
+}
+
+type Widget struct {
+	Callable
```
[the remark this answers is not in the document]
levb: Please ignore for now, this file is a scratchpad from the past; will remove before merging.

server/cloudapps/expand.go
levb: Added a note to the doc, 1/5 out of scope for this project. But the exact consent model is TBD. E.g. admin consents that the app has access to "UserJoinedChannel", but maybe the channel admin's consent is required to Subscribe to any specific channel?

server/cloudapps/proxy.go
levb: Right. Ack the dilemma. For the POC I'm 1/5 on hardcoding forced paths to try to work out a convention.

server/command/command.go
levb: A future PR, a registration API for all things. Probably best elsewhere (in cloudapps[/commands]?)

server/constants/constants.go
```diff
@@ -0,0 +1,15 @@
+// Copyright (c) 2019-present Mattermost, Inc. All Rights Reserved.
+// See License for license information.
+
+package constants
+
+const (
+	Repository     = "mattermost-plugin-cloudapps"
+	CommandTrigger = "apps"
```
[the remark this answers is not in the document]
levb: This is just a placeholder.

server/cloudapps/app_install.go
```diff
@@ -0,0 +1,30 @@
...
+
+package cloudapps
+
+import (
+	"errors"
+
+	"github.com/mattermost/mattermost-plugin-cloudapps/server/utils/md"
+)
+
+type InInstallApp struct {
```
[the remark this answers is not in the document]
levb: The idea is that each API defines its own Out type so we have very specific Go types. If some end up overlapping almost entirely I'd support merging them, otherwise keep them separate.

server/cloudapps/on_user_joined_channel.go
levb: maybe at some point it maybe worth moving the entire notification proxy into its own package, but 0/5 no reason to do it just yet.

server/cloudapps/app_install.go
```diff
@@ -0,0 +1,30 @@
...
+
+type OutInstallApp struct {
+	md.MD
+}
+
+func (r *registry) InstallApp(in *InInstallApp) (*OutInstallApp, error) {
+	if in.App.AppID == "" {
+		return nil, errors.New("app ID must not be empty")
+	}
+	r.apps[in.App.AppID] = &in.App
```
iomodo, line 24: Won't we get race condition here?
Also don't we need UninstallApp method as well?

levb: yyy, a total placeholder for now - will mark as such

server/command/command.go
```diff
@@ -0,0 +1,102 @@
...
+
+type Command interface {
+	Init(*configurator.BuildConfig) error
+	Handle(pluginContext *plugin.Context, commandArgs *model.CommandArgs) (out md.MD, err error)
+}
+
+type command struct {
+	Mattermost
+	configurator.Configurator
+}
```
iomodo, line 29: Is this responsible only for current slash commands or other commands as well? Could we have better naming?
If it's just a slash command I'd argue that structure should be more detailed, with subcommands and etc. To combine autocomplete and executing of the slash commands in the same code.

levb: Correct, just the `apps` command itself (install, list, subscribe, etc).

I am working on base structs for Forms, Modals, and Autocomplete now, under `./cloudapps`. The naming will hopefully become clearer then.

Agree on the parsing etc - leaving it for a future enhancement PR, the goal here is just to get the `Hello` going.

server/cloudapps/on_user_joined_channel.go
jfrerich: sure. I'm also 0/5 and was just throwing it out there.  Non-blocking nit

levb:
I am inclined to merge "as is", so that incremental PRs can be made, and we get a public godoc to link in the spec.

@iomodo @catalintomai @hanzei any objections?

iomodo (approved):

hanzei (approved):

catalintomai (approved):

PR #3, opened 2020-09-09
Initial data structs for forms, autocomplete, and modals

levb:


larkox (approved):

server/cloudapps/form.go
```diff
@@ -0,0 +1,71 @@
...
+	// RefreshOnChangeTo indicates that changes to the listed fields must reload
+	// the list (and reset the current value if the old one is not available).
+	RefreshOnChangeTo []string `json:"refresh_on_change_to,omitempty"`
+	RefreshURL        string   `json:"url,omitempty"`
+}
+
+type textElementProps struct {
+	Subtype   string `json:"subtype,omitempty"`
+	MinLength int    `json:"min_length,omitempty"`
+	MaxLength int    `json:"max_length,omitempty"`
```
mickmister, line 53: 👍

server/cloudapps/autocomplete.go
```diff
@@ -0,0 +1,38 @@
+package cloudapps
+
+type Autocomplete struct {
+	Form
+}
+
+type autocompleteElementProps struct {
+	FlagName string `json:"flag_name"`
```
mickmister, line 8: Can we have this be `FieldName` or something similar? Mainly because I want to include positional arguments here

mickmister (approved):

server/cloudapps/form.go
```diff
@@ -0,0 +1,71 @@
...
+
+const (
+	ElementTypeCommand       = ElementType("command")
+	ElementTypeText          = ElementType("text")
+	ElementTypeStaticSelect  = ElementType("static_select")
+	ElementTypeDynamicSelect = ElementType("dynamic_select")
+	// ElementTypeTime          = ElementType("time")
+	ElementTypeBool    = ElementType("bool")
+	ElementTypeUser    = ElementType("user")
+	ElementTypeChannel = ElementType("channel")
```
mickmister, line 23: We also need a `textarea` type that is already supported by the interactive dialogs

server/cloudapps/autocomplete.go
levb: `elementProps` already has a `Name` - which is the JSON name - is it not sufficient? Or you want display names for positional arguments (not sure if it makes sense - where would they be displayed?)

server/cloudapps/form.go
```diff
@@ -0,0 +1,71 @@
...
+	Type ElementType `json:"type"`
+
+	// Name is the name of the JSON field to use when submitting
+	Name string `json:"name"`
+
+	// Description is long description, can be used in modals in addition to
+	// Label, and in Autocomplete as Help
+	Description string `json:"description,omitempty"`
+
+	IsRequired bool `json:"is_required,omitempty"`
```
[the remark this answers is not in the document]
levb: What exactly would the semantics be? Also, should it be `IsNotClearable` (better defaulting? depends on the semantics. Is this a React thing, or common-sense UX (asking in case there are naming implications)

server/cloudapps/form.go
levb: I _thought_ it was a subtype or some other flag - as a data type, only `text` makes sense. 2/5 separate presentation from encoding.

server/cloudapps/form.go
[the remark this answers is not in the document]
mickmister: None of the select boxes in the interactive dialog are clearable at the moment. 0/5 on how the developer can convey they want it clearable or not, but it needs to be supported. Maybe we should have them be clearable and not have the config field for it.

server/cloudapps/autocomplete.go
mickmister: > elementProps already has a Name - which is the JSON name - is it not sufficient?

The `Name` field works :+1:

PR #4, opened 2020-09-10
POC Authorization flow

larkox:
#### Summary
First draft on a POC over the Authorization Flow for Apps.

There is a first request to install an app from an URL (be it marketplace, or external), which call "InstallApp".
This will get the manifest of the app, and ask the sys admin to authorize that manifest (mainly the permissions).
The Sys admin will receive somehow (in the current implementation an interactive post) a screen to authorize the app.
When the app is authorized, a new "service account" (or bot account, or however we define it) is created, and the secret for that account is generated.
That secret is sent to the app, and the app returns their secret (so we can send events to the app).
Whenever the app wants to use the API as a user, it has to ask for the user token.
To do that, it makes a request signed with JWT, asking for the token of certain user. The server verifies the JWT, and create a token with the permissions defined in the manifest, and send it to the app.

#### Ticket Link
None

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -3,6 +3,7 @@ module github.com/mattermost/mattermost-plugin-starter-template
 go 1.12
 
 require (
+	github.com/mattermost/mattermost-plugin-api v0.0.11
 	github.com/mattermost/mattermost-server/v5 v5.3.2-0.20200723144633-ed34468996e6
 	github.com/pkg/errors v0.9.1
 	github.com/stretchr/testify v1.6.1
--- a/server/authorization.go
+++ b/server/authorization.go
@@ -1,0 +1,298 @@
+package main
+
+import (
+	"bytes"
+	"encoding/json"
+	"errors"
+	"fmt"
+	"net/http"
+
+	"github.com/mattermost/mattermost-plugin-api/experimental/common"
+	"github.com/mattermost/mattermost-server/v5/model"
+)
+
+const ContextState = "state"
+const StateKey = "s_"
+const AppKey = "app_"
+
+type Manifest struct {
+	Name        string
+	Description string
+	URL         string
+	Permissions string
+}
+
+func (m *Manifest) ParsePermissions() string {
+	return m.Permissions
+}
+
+func (p *Plugin) InstallApp(url, adminID string) error {
+	manifestURL := url + "/manifest"
+	r, err := http.Get(manifestURL)
+	if err != nil {
+		return err
+	}
+
+	var manifest Manifest
+	if r.StatusCode != 200 {
+		return errors.New("cannot fetch manifest")
+	}
+
+	err = json.NewDecoder(r.Body).Decode(&manifest)
+	if err != nil {
+		return err
+	}
+
+	err = p.AskAuthorization(manifest, adminID)
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func (p *Plugin) AskAuthorization(manifest Manifest, adminID string) error {
+	id := model.NewId()
+	state := fmt.Sprintf("%s_%s", manifest.Name, id)
+	key := fmt.Sprintf("%s_%s", StateKey, state)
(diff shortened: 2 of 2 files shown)


catalintomai:
"Whenever the app wants to use the API as a user, it has to ask for the user token." - does that mean that every server API requests requires an extra request to the plugin? No app(s) caching/storage?

larkox:
I have not read the document before. I was mainly basing my work on how it is done for Google Cloud (https://developers.google.com/identity/protocols/oauth2/service-account).

The current approach does not consider at the moment per-user permissions. I personally thought that the complexity was not needed. But I can engineer it into the solution.
For security and some simplicity for the app (less storage), I thought the approach taken by the link shared before was interesting. Whenever you want to act as a user, you ask for a short-lived token. Do all you need in that short time with the same token, and then let the token expire. The next time you need to act on behalf of that user, just do it again. That means more calls to the API, but solves the problem of storage on the app.

larkox:
@catalintomai :point_up:

catalintomai:
@larkox , the current PM spec https://docs.google.com/document/d/1K7LIGRIx6Sf_6G45NXdTomu11gOooxFRRXCEwO2tktg/edit# looks to have the per-user permissions requirement, @aaronrothschild can confirm. Though it also mentions "Improvements in Permissions are needed to limit what apps are visible to particular groups of users.", don't think we talked about/addressed this requirement.

larkox:
In other plugins and situations, we are trying to get away from "user level authorization". Having an end-user authorization seems in certain contexts "less secure" than having a general authorization given by the admin. In other words, some organizations prefer the admin to give the authorization for all users instead of each user having to go through the authorization process. @aaronrothschild probably can give more insight into this.

App visibility for group of users is complex in the design, mainly to define what do we mean with "group of users". If we use mattermost groups, we could add another layer of scoping where "if the app is trying to act on behalf of user A, but user A does not belong to group B, return an unauthorized error". I can work with that too.

levb:
$my .02: 👍  to what @larkox said, with a slight twist.

I think we should default to the user-level authorization, but maybe try to tie in the UX experience with `connect`-ing to the 3rd party in the same flow if appropriate.

In some cases, the sysadmins should have the capability of declaring the app "trusted", so user-level authorization is skipped.

This is what I also discussed with @crspeller

aaronrothschild:
@catalintomai @larkox I think it's worth having a conversation with the Enterprise (?) team about their plans for security in the future.  A switch from RBAC to ABAC may make implementing this easier in the future (in which case we should wait?) or we may need to add Permissions granularity to plugins.  It's a conversation not a requirement.

Primary Use Case for permissions: 
- I have an LDAP "Dev team" and "Support team" and "CSM Team"... 
   - I want everyone who has a Zendesk account to be able to see the Zendesk message action "button" after they have authenticated using `/zendesk connect`
  - Or, I want only the support team to see it at all.  Only people in "Support Team" should be able to connect and see the button in Mattermost. 

I think there is also a bigger question about plugins and our security model in the future.  We should try to stay away from hard coding Security assumptions into plugins.   I would prefer where in the plugin config it could specify a mattermost group for example.

levb:
> Whenever the app wants to use the API as a user, it has to ask for the user token.
> To do that, it makes a request signed with JWT, asking for the token of certain user. The server verifies the JWT, and create a token with the permissions defined in the manifest, and send it to the app. 

@larkox, the plan of record is to use the standard OAuth2 flows to get the user API token, and store it itself. Is this implementation a suggested replacement, or just a POC/partial implementation to get the flow in place?

larkox:
@levb This part a suggestion based on this:
https://developers.google.com/identity/protocols/oauth2/service-account

levb:
@larkox do you want to keep this PR open, or re-work the service (bot) account into master?

larkox:
I think it is better to close this one and start from scratch (using this as reference when needed) for the service account.

PR #5, opened 2020-09-14
MM-28442: JWT, Incorporated much of Install/Authorize POC, cleanup

levb:
Fixes https://mattermost.atlassian.net/browse/MM-28442

- Reworked #4 to match the rest of the code base, used an Interactive Dialog
- Added a simple JWT to the outgoing change notifications
- Added a simple Hello App hosted in the plugin itself for now

Currently implemented as a command, 
image

Pops up:
image

Upon pressing "Approve and Install", takes the user to the DM channel, with the updates
image


server/cloudapps/proxy.go
```diff
@@ -60,14 +84,19 @@ func (p *proxy) SendChangeNotification(s *Subscription, msg interface{}) {
...
+	}
+	req.Header.Set("Authorization", "Bearer "+token)
+	resp, err := client.Do(req)
 	if err != nil {
 		// <><> TODO log
+		// TODO ticket: progressive backoff on errors
 		return
 	}
-	defer resp.Body.Close()
+	resp.Body.Close()
```
larkox, line 99: Is there any reason to keep the body open in any branch (mainly the error handling)? If not, we can use the defer at the end of the function.

server/http/dialog/install.go
```diff
@@ -0,0 +1,150 @@
...
+	noUserConsentForOAuth2 := false
+	if consentValue == "require" {
+		noUserConsentForOAuth2 = true
+	}
+
+	v = dialogRequest.Submission["secret"]
+	secret, _ := v.(string)
+
+	var manifest apps.Manifest
+	err = json.Unmarshal([]byte(dialogRequest.State), &manifest)
```
larkox, line 132: We are using the manifest that comes with the request. This might be a security issue if someone was able to change the content of the manifest on the client side, then making the user accept some permissions while the underlying manifest has others.

In order to solve this, in the original PR I did some countermeasures inspired by OAuth:
- I create a unique ID for the dialog, and create a "state" variable in the dialog.
- I add a register in the KVStore with the ID as key, and the manifest as value
- When the dialog returns, I check the state, and check if that state is in the KVStore.
- If it is in the KVStore, I authorize the manifest I have in the store, since it is the one I used to create the dialog

Probably it is not a perfect solution, and I may be making things more secure than they need, but I thought it was a nice addition.

The relevant code is here:
https://github.com/mattermost/mattermost-plugin-cloudapps/blob/05a96c274a064f0ca66701b1f64f533908282500/server/authorization.go#L55-L67
https://github.com/mattermost/mattermost-plugin-cloudapps/blob/05a96c274a064f0ca66701b1f64f533908282500/server/authorization.go#L81-L86
https://github.com/mattermost/mattermost-plugin-cloudapps/blob/05a96c274a064f0ca66701b1f64f533908282500/server/authorization.go#L131-L150

server/cloudapps/install_app.go
```diff
@@ -0,0 +1,43 @@
...
+	App *App
+}
+
+func (r *registry) InstallApp(in *InInstallApp) (*OutInstallApp, error) {
+	if in.Manifest.AppID == "" {
+		return nil, errors.New("app ID must not be empty")
+	}
+	// TODO check if acting user is a sysadmin
+
+	// TODO remove mock, implement for real
```
larkox, line 30: Regarding real implementation (to make sure we are all aligned):

We already got the app secret during the dialog, so we can send messages to the app, but we still need to inform the app of two things:
  - We are installed
  - What is our secret so they can communicate with us (or the OAuth token, or whichever communication we want to set in place)

I understand this will be done by a simple HTTP call to the app, or opening a new window to trigger the OAuth flow.

larkox (approved):
Awesome work Lev :tada:

Just some considerations on details of the implementation, but it is looking really good.

server/cloudapps/install_app.go
levb: Yes.

server/http/dialog/install.go
levb: If a remote user session is compromised, it can just execute necessary APIs to install whatever it wants, I am kinda puzzled with the need for the extra verification steps here.

0/5 we mark this (obviously! regardless of how we verify the steps) for a review with the security team once all of the Auth flows are in place?

server/cloudapps/proxy.go
levb: Later if we need to... KISS for now. This came up before when the error messages were of unanticipated types, better NOT happen to us here!

server/cloudapps/install_app.go
levb: > What is our secret so they can communicate with us (or the OAuth token, or whichever communication we want to set in place)

To use the REST API as a user, the App must go through the OAuth2 loop. To use the Bot-level token - TBD, but I think we discussed the general outline - provision the bot upon install, then use the bot account.

server/apps/install_app.go
```diff
@@ -0,0 +1,44 @@
...
+
+import (
+	"errors"
+
+	"github.com/mattermost/mattermost-plugin-apps/server/utils/md"
+)
+
+type InInstallApp struct {
+	ActingMattermostUserID string
+	NoUserConsentForOAuth2 bool
```
mickmister, line 14: Why is this a negative boolean as opposed to `NeedsUserConsentForOAuth2` or something similar?

mickmister (approved):
LGTM! Just one question on the naming of `NoUserConsentForOAuth2`

server/apps/install_app.go
levb: default == nil value, so negate the name instead?

server/cloudapps/install_app.go
larkox: In theory, the OAuth protocol should handle re-validation, but I have to dig deeper into that.

PR #7, opened 2020-09-18
Add OAuth App creation

larkox:
#### Summary
Add OAuth App creation

#### Ticket Link
None


server/apps/install_app.go
```diff
@@ -5,40 +5,120 @@ package apps
...
 	out := &OutInstallApp{
-		MD:  md.Markdownf("Installed %s (%s)", in.Manifest.DisplayName, in.Manifest.AppID),
+		MD:  md.Markdownf("Installed %s (%s)\n%s", in.Manifest.DisplayName, in.Manifest.AppID, extraInfo),
 		App: app,
 	}
 	return out, nil
 }
+
+func (r *registry) createBot(manifest *Manifest, sessionToken string) (*model.Bot, *model.UserAccessToken, error) {
+	client := model.NewAPIv4Client(r.configurator.GetConfig().MattermostSiteURL)
```
larkox, line 74: All APIv4Client on this file will not work because there is no logged user to the client. I am not sure how to get an authenticated client in this context.

levb: 0/5 seems the most logical choice is to either create the new needed p.API methods, or to find a way to create a token from the sessionID that we have.

levb: More confused... We have a `p.API.GetSession` and it has the token - why not use it?

levb (approved):
This looks good, but presumably we should not merge this until we work out the bot/OAuth creation issue?

server/apps/install_app.go
larkox: Maybe I am missing something. This is done we hit a button in a modal dialog. Therefore, this comes from the HTTP handler. The HTTP handler only have access to the request and the response writer, neither of which contains the session ID nor anything similar.

The base ServeHTTP function does have the plugin.Context object that contains the sessionID, but it is "lost" when calling the ServeHTTP of the router. I will dig deeper into how to pass this information (maybe adding an extra header in the request?).

levb: :+1

server/http/service.go
```diff
@@ -33,5 +33,6 @@ func NewService(router *mux.Router, apps *apps.Service, initf ...func(*mux.Route
 
 // Handle should be called by the plugin when a command invocation is received from the Mattermost server.
 func (s *service) ServeHTTP(c *plugin.Context, w http.ResponseWriter, r *http.Request) {
+	r.Header.Add("MM_SESSION_ID", c.SessionId)
```
levb, line 36: I 👎  Gorilla for this - no easy way to pass extra data into the handlers. They recommend using `Context` but 3/5 it's a wrong thing to do.

levb (approved):

larkox:
I made a small change to not check the existing User Access Tokens, since you can see they exist, but you cannot get the Token. We may want to revisit how we handle this (delete the other present tokens, delete them during uninstall...).

I also changed how the information is shown to the user (which I thought was already done). Here a screenshot of the last step of the install process.
Screenshot from 2020-09-21 17-43-11

This information is shown here to allow the person installing the app fill this information "in the other side". This is not the safest way to do it, since the token for the bot is supposed to be "read & burn", and we are just posting it in a message. We could write that in an ephemeral, which would improve the security, but it is still something I would worry about.

There is also the solution to send this information to an endpoint of the app (i.e. `http://myapp.com/mattermost/installComplete`), but again I worry about the security (in this case, probably over worrying).

levb:
Yeah, we don't want to display secrets in the UI (screen captures), best to (https) send to the app.

larkox:
@levb In order to send directly the information to the app, I had to do a big refactor. I may have done too much in a single step, so I am OK reverting this commit and doing most of the refactor on a different PR if you find it better.

jfrerich (approved):

jfrerich:
I actually like the refactoring and the introduction of the `appmodel` package.  At first I was thinking it should be `model`, but suppose that would require import naming/aliasing because of the server models.

larkox:
Per discussion with @levb , I reverted the last two commits since we will tackle the client problem in a different PR.

PR #9, opened 2020-09-23
Change notifications and subscriptions

jfrerich:
#### Summary
This is a draft PR for the following:

I've created a doc (linked to from the spec) that contains more in-depth discussion, findings, thoughts. https://docs.google.com/document/d/1SXE6IopMqsOTDmhwn0bokBOJNj31-EFSDCNO5XIFXus/edit#

**Subscriptions:**
- Add a subscription (API available through http)
- Delete a subscription (API available through http)

**Change Notifications handled:**
- UserHasBeenCreated
- UserHasJoinedChannel
- UserHasLeftChannel
- UserHasJoinedTeam
- UserHasLeftTeam
- MessageHasBeenPosted
- ChannelHasBeenCreated

**Expanded working for:**
- config
- channel
- team
- user
- acting_user

Needs Expand for:
- ParentPost
- Post
- RootPost
- Mentioned

#### Ticket Link
n/a


server/apps/subscriptions.go
```diff
@@ -4,31 +4,149 @@
...
+	}
+
+	// check if sub exists
+	var newSubs []*Subscription
+	for _, s := range savedSubs {
+		if s.SubscriptionID == subID {
+			continue
+		}
+		newSubs = append(newSubs, s)
+	}
```
larkox, line 116:
```suggestion
	for i, s := range savedSubs {
		if s.SubscriptionID == subID {
			newSubs = append(newSubs, savedSubs[i+1:]...)
			break
		}
		newSubs = append(newSubs, s)
	}
```

What about something like this?

server/apps/subscriptions.go
```diff
@@ -4,31 +4,149 @@
...
+
+	// sub was deleted. update and save updated subs
+	// TODO check for following:
+	//   - don't need to save if sub was not deleted?
+	//   - if delete the last subscription for the channel, delete the key also
+	_, err := subs.mm.KV.Set(key, newSubs)
```
larkox, line 122: Should we worry about the lack of atomicity of this? We might end up with a race condition if two subscriptions where to be deleted nearly at the same time.

server/http/api/api.go
```diff
@@ -0,0 +1,79 @@
...
+		errD, _ := json.MarshalIndent(err, "", "    ")
+		fmt.Printf("err = %+v\n", string(errD))
+		// return respondErr(w, http.StatusInternalServerError, err)
+		return
+	}
+
+	subs := apps.NewSubscriptions(a.mm, a.configurator)
+
+	switch req.Method {
+	case http.MethodPost:
```
larkox, line 68: Does it make sense to add a MethodGet for listing the subscriptions? Probably only used for debugging or similar, but I feel it could be useful.

larkox: And a method Method Patch or Put to update a subscription? Again, not sure of the usecase, but might be handy.

larkox (commented):
Sorry for the drive-by review :car: 
It looks very promising. I just made a few comments, but in general, it looks good to me.

server/apps/on_post_created.go
```diff
@@ -0,0 +1,57 @@
...
+	PostID         string
+	RootID         string
+	UserID         string
+	// TeamID         string
+	Expanded *Expanded
+}
+
+// OnUserHasBeenCreated sends a change notification when a new user has
+// joined a team.
+func (p *proxy) OnPostHasBeenCreated(ctx *plugin.Context, post *model.Post) {
```
levb, line 25: This can be worked into a function shared for all subscriptions, with a handful of parameters.

server/apps/on_post_created.go
```diff
@@ -0,0 +1,57 @@
...
+	subs, err := p.Subscriptions.GetSubscriptionsForChannelOrTeam(SubjectPostCreated, "")
+	if err != nil {
+		// p.Logger.Debugf("OnPostHasBeenCreated: failed to get subscriptions: %s %s: ",
+		// 	SubjectPostCreated, user.UserId, err)
+		return
+	}
+	if len(subs) == 0 {
```
levb, line 32: 2/5 make `GetSubscriptionsForChannelOrTeam` return `utils.ErrNotFound` when there are 0 subs, will simplify the logic.

server/apps/on_post_created.go
```diff
@@ -0,0 +1,57 @@
...
+		return
+	}
+
+	expander := NewExpander(p.mm, p.configurator)
```
levb, line 36: 1/5 At least this loop can be worked into an `ExpandNotification`-ish generic function

server/plugin/plugin.go
```diff
@@ -19,6 +19,7 @@ import (
 	"github.com/mattermost/mattermost-plugin-apps/server/configurator"
 	"github.com/mattermost/mattermost-plugin-apps/server/constants"
 	myhttp "github.com/mattermost/mattermost-plugin-apps/server/http"
+	"github.com/mattermost/mattermost-plugin-apps/server/http/api"
```
levb, line 22: 1/5 recommend package name either `restapi` or `httpapi` to avoid collisions and reveal the intention. This package name should be used rarely (`plugin.go` only?) so making it (even more?) verbose is ok.

server/plugin/plugin.go
```diff
@@ -64,6 +65,7 @@ func (p *Plugin) OnActivate() error {
 	p.http = myhttp.NewService(mux.NewRouter(), p.apps,
 		dialog.Init,
 		helloapp.Init,
+		api.Init,
```
levb, line 68: `restapi.Init,` would be more intention-revealing

server/plugin/plugin.go
```diff
@@ -90,6 +92,22 @@ func (p *Plugin) ServeHTTP(c *plugin.Context, w http.ResponseWriter, req *http.R
 	p.http.ServeHTTP(c, w, req)
 }
 
+func (p *Plugin) UserHasBeenCreated(pluginContext *plugin.Context, user *model.User) {
```
levb, line 95: 2/5 We should define the right interface for the `OnXXX` methods (`Hooks`?) and embed it in the Plugin struct directly, avoid these extra proxy methods. Separate PR is ok.

server/http/api/api.go
levb: 1/5 Let's not patch. Idempotent changes only, KIS for now?

server/http/api/api.go
```diff
@@ -0,0 +1,79 @@
...
+}
+
+func Init(router *mux.Router, apps *apps.Service) {
+	a := api{
+		mm:           apps.Mattermost,
+		configurator: apps.Config,
+	}
+
+	subrouter := router.PathPrefix(constants.APIPath).Subrouter()
+	subrouter.HandleFunc(SubscribePath, a.handleSubscribe).Methods("POST", "DELETE")
```
levb, line 41: 0/5 not sure what the Gorilla-idiomatic style is, but seems like having separate handlers defined for the methods would be cleaner. The bolierplate of handling requests can be worked into helper function(s).

server/apps/subscriptions.go
```diff
@@ -4,31 +4,149 @@
...
+	pluginapi "github.com/mattermost/mattermost-plugin-api"
 	"github.com/mattermost/mattermost-plugin-apps/server/configurator"
+	"github.com/pkg/errors"
 )
 
+const SubsPrefixKey = "sub_"
+
 type Subscriptions interface {
-	GetSubscriptionsForChannel(subj SubscriptionSubject, channelID string) ([]*Subscription, error)
+	GetSubscriptionsForChannelOrTeam(subj SubscriptionSubject, channelOrTeamID string) ([]*Subscription, error)
```
levb, line 15: - 1/5 s/`Subscriptions`/`Subs`/g, use `subs *Subs` for variable name, consistently.
- 1/5 Keep `Subscription`. Reason: typography.
- 2/5 use `sub *Subscription` except in obvious context (member functions) `s *Subscription` is acceptable.
- 2/5 `GetHookSubs(..., channelID, teamID, etc...)` and `GetAppSubs`(appID)`

server/apps/subscriptions.go
```diff
@@ -4,31 +4,149 @@
...
+	// }
+	//
+	// go p.SendChangeNotification(s, msg)
+	return nil
+}
+
+// GetSubsKey returns the KVstore Key for a subsject. If teamOrChannelID
+// provided, the value is appended to the subject key making it unique to the
+// channelID or teamID
+func GetSubsKVkey(subj SubscriptionSubject, teamOrChannelID string) string {
```
levb, line 140: 2/5 hardcode the key prefixes for simpler readability (and fewer lines of code!), 0/5 hardcode them as function 1-liners if it makes it shorter. I think we can reduce the number of lines of code and indirections here.

server/apps/subscriptions.go
levb: 1/5 I specifically asked @jfrerich not to worry concurrency yet. I fail to imagine this to be the final storage implementation. We just need a POC that works functionally. Having said that, _simple_ improvements would be welcome.

levb (changes requested):

levb:
Also, lint failing?

jfrerich:
Thanks for the very early feedback @levb and @larkox.  I'll continue to work on this next week and incorporate your suggestions.

server/apps/subscriptions.go
jfrerich: That will work nicely and allow exiting the loop earlier, if possible.  Lev and I discussed building out a POC with minimal effort on efficiency concerns.  Thanks for this enhancement

jfrerich: done.

server/plugin/plugin.go
jfrerich: I like restapi for the package naming.  committed the change.

server/plugin/plugin.go
jfrerich: Done

server/apps/subscriptions.go
jfrerich: @levb, what is your take on `GetSubsForChannelOrTeam` vs `GetChannelOrTeamSubs`?

jfrerich: For each of the notification structs.  Do you prefer the shortened field name `SubsID`?
```go
type SubscriptionDeletedNotification struct {
	SubscriptionID SubscriptionID
	Subject        SubscriptionSubject
	Expanded       *Expanded
}

```

jfrerich: See [`036f2bb` (#9)](https://github.com/mattermost/mattermost-plugin-apps/pull/9/commits/036f2bb0eb151409ace05a2cd5b2e06d38813f6e)

server/apps/subscriptions.go
jfrerich: Are you suggesting to scrap the `GetSubsKVkey` function altogether and hardcode the prefixes in each `KV.Get()` call?

server/plugin/plugin.go
jfrerich: I see you have hooks.go in https://github.com/mattermost/mattermost-plugin-apps/pull/13/files.  I'll review and hopefully merge that PR soon and then merge into this branch

server/http/api/api.go
jfrerich: I picked this implementation because the endpoint is the same and the function is determined by a switch statement based on the method type.  I don't see many variations of `subscribe` path so chose this instead of two handlers (handleSubscribeStore, handleSubscribeDelete).  I'm also not positive if this is idiomatic Gorilla. I just saw the option to pass in a string slice

server/apps/subscriptions.go
levb: Please ignore the comment, I misunderstood the structure of the key

server/http/restapi/api.go
```diff
@@ -0,0 +1,74 @@
...
+
+	// actingUserID := req.Header.Get("Mattermost-User-Id")
+	// fmt.Printf("actingUserID = %+v\n", actingUserID)
+	// if actingUserID == "" {
+	// 	// err = errors.New("user not logged in")
+	// 	status = http.StatusUnauthorized
+	// 	return
+	// }
+
+	body, _ := ioutil.ReadAll(req.Body)
```
levb, line 54: json.NewDecoder(req.Body).Decode(&subRequest)

server/apps/expand.go
```diff
@@ -101,7 +103,23 @@ func (e *expander) collectConfig(expand *Expand) error {
 	return nil
 }
 
-func (e *expander) collectChannel(channelID string) func(*Expand) error {
+func (e *expander) collectTeam(teamID string) func(*Expand) error {
```
jfrerich, line 106: @levb I added this function back into the expander because leaving/joining a team, does not have a "channel".  

However, I left the `collectChannelAndTeam` function because, given a channel, you can retrieve the team.

server/apps/api_notify.go
```diff
@@ -0,0 +1,76 @@
...
+	if tm != nil {
+		msg.TeamID = tm.TeamId
+		msg.UserID = tm.UserId
+		channelOrTeamID = tm.TeamId
+	}
+	if post != nil {
+		msg.PostID = post.Id
+		msg.ParentID = post.ParentId
+		msg.RootID = post.RootId
+	}
```
jfrerich, line 52: FYI, I hate this code, but wanted to get this PR to you.  This code determines how to get IDs for the `Expand()` functions.  

I'm not sure of the difference between actingUser and User.  It's very possible I'm not handling that correctly.  See line 30 where I set UserID: actingUserID

server/apps/api.go
```diff
@@ -3,7 +3,20 @@
...
 
+import (
+	"github.com/mattermost/mattermost-plugin-apps/server/constants"
+	"github.com/mattermost/mattermost-server/v5/model"
+)
+
 type API interface {
 	Call(Call) (*CallResponse, error)
 	InstallApp(InInstallApp) (*OutInstallApp, error)
+	Notify(
```
levb, line 14: @larkox I will refactor the parameters soon

server/apps/api_notify.go
levb: Right. In the case when no obvious "actingUser" is available, let's set it to nil.

levb: `UserID` should be the subject of the action, aka UserJoinedChannel - would be the userID that joined. For a Post, it's the UserId on the Post. It should rarely (if ever?) be set to actingUserId

levb (approved):
Besides the UserID comment, LGTM.

I'm going to fix the Expand/Context soon, so it's ok to have stub methods there for the time being.

server/apps/api_notify.go
jfrerich: @lev, maybe the API docs are what confused me (and why I had set `userID` to `actingUserID`.  Looking at [Plugin server reference](https://developers.mattermost.com/extend/plugins/server/reference/) all the `UserHas...` hooks have the `model.User` in the function signature set named as `actor`


image

jfrerich: Is the actor not the user that left/joined each associated team/channel?

levb:
Absorbed in #20

PR #10, opened 2020-09-24
Integrate Oauther with helloapp

larkox:
#### Summary
Draft of integration with OAuther. Everything should be working, but still needs to be connected to real values and processes.

#### Ticket Link
None

### DIFF
--- a/server/http/helloapp/helloapp.go
+++ b/server/http/helloapp/helloapp.go
@@ -31,6 +31,7 @@ func Init(router *mux.Router, apps *apps.Service) {
 
 	subrouter := router.PathPrefix(constants.HelloAppPath).Subrouter()
 	subrouter.HandleFunc("/mattermost-app.json", a.handleManifest).Methods("GET")
+	subrouter.PathPrefix("/oauth2").HandlerFunc(a.handleOAuth).Methods(http.MethodGet, http.MethodPost)
 }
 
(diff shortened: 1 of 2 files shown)


levb (approved):
WFM.

In the case of "install" flow, the app may need that installing sysadmin's privileges, so there should be a way to "connect" _him/her_ within the install flow. cc @aaronrothschild

larkox:
As any OAuth we know, in order to connect any user through OAuth we need:
- To send a specific URL to the user
- The user has to open that URL that will open in a new window (we could use wish responses for this, but if popups are blocked the window may not be opened)
- (This happens automatically, here just for completion) That URL goes to the App, which does all needed preparations (including saving the state) and redirects to the OAuth provider (in this case Mattermost)
- (We have no control over this, this is the normal OAuth flow, here just for completion) The new window will show a screen from Mattermost, where we may have to log in, and accept the permissions requested.
- (This happens automatically, here just for completion) When accepted the permissions, this will redirect to a URL of the App. Here the app stores the token, and usually returns a website saying "Authorization complete. You can close this window" and try to close the window automatically.
- At this point, the app can do whatever it wants to notify the user or even start using the token.

levb:
We should close this PR in favor of #13 - I merged it there.

larkox:
Closing as per levb comment

PR #13, opened 2020-09-27
Hello App Install Wish with Bot token and OAuth

levb:
Implements the first part of the install process for Hello App:
- OAuth2 and Bot PAT (personal access token) are provisioned
- the `Install` Wish from the manifest is executed, it requests the user to OAuth-connect
- Upon connection, `connected_install` wish is executed, currently just requests the number of teams

Includes https://github.com/mattermost/mattermost-plugin-apps/pull/10
Depends on https://github.com/mattermost/mattermost-plugin-api/pull/76 (and then some go.mod manipulation)


server/http/helloapp/oauth.go
```diff
@@ -0,0 +1,128 @@
...
+	}
+
+	return &oauth2.Config{
+		ClientID:     id,
+		ClientSecret: secret,
+		Endpoint: oauth2.Endpoint{
+			AuthURL:  conf.MattermostSiteURL + "/oauth/authorize",
+			TokenURL: conf.MattermostSiteURL + "/oauth/access_token",
+		},
+		RedirectURL: h.AppURL(PathOAuth2Complete),
```
larkox, line 53: RedirectURL is not needed. It is handled here: https://github.com/mattermost/mattermost-plugin-api/blob/8b14e62b421ca2816901165e9ff750895451e621/experimental/oauther/oauth2.go#L102

server/http/helloapp/oauth.go
```diff
@@ -0,0 +1,128 @@
...
+		return
+	}
+
+	// TODO 2/5 we should wrap the OAuther for the users as a "service" so that
+	//  - startOAuth2Connect is a Call
+	//  - payload for finish should be a Call
+	//  - a Wish can check the presence of the acting user's OAuth2 token, and
+	//    return Call startOAuth2Connect(itself)
+	// for now hacking access to apps object and issuing the call from within
+	// the app.
```
larkox, line 84: This makes sense. It would alleviate the needs of the App in terms of storage (it no longer has to store the tokens, just ask for the token, and if it doesn't receive it, understand it is not authorized by that user).

This will also make things slightly more secure, since the app information stays in mattermost instead of residing in the app.

The only problem is then how does the App identifies itself to ask for the tokens? Are we sharing yet another secret?

server/http/helloapp/helloapp.go
```diff
@@ -10,94 +10,83 @@ import (
...
 	}
 
 	subrouter := router.PathPrefix(constants.HelloAppPath).Subrouter()
-	subrouter.HandleFunc("/mattermost-app.json", a.handleManifest).Methods("GET")
+	subrouter.HandleFunc(PathManifest, h.handleManifest).Methods("GET")
+
+	subrouter.HandleFunc(PathWishInstall, wish(h.handleInstall)).Methods("POST")
+	subrouter.HandleFunc(PathWishConnectedInstall, wish(h.handleConnectedInstall)).Methods("POST")
+
+	subrouter.PathPrefix(PathOAuth2).HandlerFunc(h.handleOAuth).Methods(http.MethodGet, http.MethodPost)
```
larkox, line 46: This was a mistake in my PR. Only GET is needed. No need for POST.

larkox, line 46: This is the part I like the less about how to handle URLs within the package. The idea to do it like this was to remove dependencies to gorilla mux. But after trying to use it here, I personally didn't like it much. Any idea to make this work better is welcomed.

server/http/helloapp/oauth.go
```diff
@@ -0,0 +1,128 @@
...
+
+func (h *helloapp) InitOAuther() error {
+	oauth2Config, err := h.GetOAuthConfig()
+	if err != nil {
+		return err
+	}
+	h.OAuther = oauther.NewFromClient(h.apps.Mattermost,
+		*oauth2Config,
+		h.finishOAuth2Connect,
+		logger.NewNilLogger(),
```
larkox, line 24: This was a quick implementation. We may want to put a proper logger here.

larkox (changes requested):
Pretty good in general. A few questions and some small changes.

server/http/helloapp/oauth.go
levb: We don't have a Logger yet. I have a branch where I have some changes to make it (loggers) work to my satisfaction, but it's not ready yet. I'll add a TODO

server/apps/wish.go
```diff
@@ -0,0 +1,11 @@
+package apps
+
+type Wish struct {
+	URL string
```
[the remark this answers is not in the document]
levb: I don't want to introduce it to the struct until it's clear how to use it. 0/5 There is no case yet. FWIW, if we have a convention, then ${AppURL}/wish/${NAME} is the URL and they become interchangeable.

server/http/helloapp/oauth.go
levb: I think the App should _store_ the tokens and the app credentials, as it does now. I just want the wire the flow of InstallWish -> ConnectWish -> ...link/browser... -> CompleteCallbackWish (aka connected install)

levb:
@jfrerich @hanzei Did CI stop running for the plugin-apps repo?

server/http/helloapp/helloapp.go
```diff
@@ -43,7 +43,7 @@ func Init(router *mux.Router, apps *apps.Service) {
 	subrouter.HandleFunc(PathWishInstall, wish(h.handleInstall)).Methods("POST")
 	subrouter.HandleFunc(PathWishConnectedInstall, wish(h.handleConnectedInstall)).Methods("POST")
 
-	subrouter.PathPrefix(PathOAuth2).HandlerFunc(h.handleOAuth).Methods(http.MethodGet, http.MethodPost)
+	subrouter.PathPrefix(PathOAuth2).HandlerFunc(h.handleOAuth).Methods("POST")
```
larkox, line 46: Maybe I was not clear on my last comment. What we need is GET. We do NOT need POST.
Also, not sure why instead of using the constant in the http package you are just putting the string yourself.

server/http/helloapp/oauth.go
```diff
@@ -50,11 +42,19 @@ func (h *helloapp) GetOAuthConfig() (*oauth2.Config, error) {
 			AuthURL:  conf.MattermostSiteURL + "/oauth/authorize",
 			TokenURL: conf.MattermostSiteURL + "/oauth/access_token",
 		},
-		RedirectURL: h.AppURL(PathOAuth2Complete),
+		// RedirectURL: h.AppURL(PathOAuth2Complete), - not needed, OAuther will configure
```
larkox, line 45: Since we no longer use PathOauth2Complete we may consider removing the declaration of this constant.

larkox (changes requested):
A small misunderstanding, and a potential cleanup.

server/http/helloapp/oauth.go
levb: We use it in the manifest (CallbackURL). Maybe we shouldn't - should come from oAuther somehow.

larkox: Understood. Yes, I think the oauther should be able to give that information. For the time being it can continue as it is.

larkox (approved):

jfrerich:
@levb, it appears ci did stop running.  Can you give me permissions to this repo and I'll investigate.

levb:
@jfrerich done

jfrerich:
Were there some merges that contained the repo name change?  I don't see anything obvious right now.  The webhooks from circleCI are working for other branches for `plugin-cloud-apps`, but they just stopped working for yours

server/apps/api_call.go
```diff
@@ -0,0 +1,16 @@
+package apps
+
+import (
+	"github.com/pkg/errors"
+)
+
+func (s *Service) Call(call Call) (*CallResponse, error) {
+	switch {
+	case call.Wish != nil && call.Modal == nil:
```
jfrerich, line 9: If one of these and only one of these must be specified, can we reduce to one field that is a bool?

jfrerich (approved):
@levb, LGTM!  I have the one question to see if we could have only one field to replace Wish and Modal. 

Regarding the circleCI, I'm am not positive why it stopped for this PR.  I suggest running `make` locally, fixing errors and we merge to master with you fixing the make checks locally.  I might suggest creating some new branch off of master and pushing to see if CI picks up for you again.

PR #18, opened 2020-10-08
Add locations

larkox:
#### Summary
Add all needed to handle locations.

#### Ticket Link
None


server/command/main.go
```diff
@@ -54,3 +57,12 @@ func runSubcommand(
...
 	return normalOut(params, md.MD("<><> TODO"), nil)
 }
+
+func (s *service) executeDebugLocations(params *params) (*model.CommandResponse, error) {
+	locations, err := s.apps.API.GetLocations(params.commandArgs.UserId, params.commandArgs.ChannelId)
+	if err != nil {
+		return normalOut(params, md.MD("error"), err)
+	}
+	bytes, _ := json.Marshal(locations)
+	return normalOut(params, md.MD(string(bytes)), nil)
```
mickmister, line 67: For better formatting
```suggestion
	return normalOut(params, md.JSONBlock(locations), nil)
```

mickmister (changes requested):
This is awesome! Great work @larkox 

I have a few requests/questions

server/apps/location.go
```diff
@@ -1,24 +1,86 @@
...
-	Type() LocationType
+type LocationInt interface {
+	GetType() LocationType
 }
 
-type location struct {
+type Location struct {
 	LocationType LocationType
 	LocationID   string
+	Wish         Wish
```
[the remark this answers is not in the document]
mickmister: Or possibly on the `Wish` struct?

server/http/helloapp/locations.go
```diff
@@ -0,0 +1,112 @@
...
+		if err != nil {
+			httputils.WriteBadRequestError(w, err)
+			return
+		}
+
+		f(w, req)
+	}
+}
+
+func ExtractUserAndChannelID(f func(http.ResponseWriter, *http.Request, string, string)) func(http.ResponseWriter, *http.Request) {
```
mickmister, line 40: Can we move this function and `CheckAuthentication` into a different file(s)?

Could we actually place this code in `mattermost-plugin-apps/server/apps`, to be shareable to other Go programs? And to have the parsing logic right by the marshaling logic.

server/apps/client.go
```diff
@@ -124,3 +153,45 @@ func (s *Service) GetManifest(manifestURL string) (*Manifest, error) {
...
+	if err != nil {
+		return nil, errors.Wrap(err, "error getting app")
+	}
+
+	url, err := url.Parse(app.Manifest.LocationsURL)
+	if err != nil {
+		return nil, errors.Wrap(err, "error parsing the url")
+	}
+	q := url.Query()
+	q.Add("userID", userID)
```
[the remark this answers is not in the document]
levb: +1 for snakes in all external appearances

server/apps/location.go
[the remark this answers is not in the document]
levb: I started now, yes now we should.

levb (approved):
I'm ok with this, subject to further refactoring.

I sent @larkox a private PR to sync up with #20. It has some of my feedback.

server/apps/location.go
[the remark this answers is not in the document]
larkox: We probably do, since we will performing the wish with the client for that app. I will think it further to see whether it should go in the Wish or the Location.

server/http/helloapp/locations.go
larkox: Since these are App related, I am not sure whether we should move them back to apps. Any app can be written in any language, not just Go, and they may or may not use the same pattern as we do. Therefore might not be useful.

Also, I am 0/5 on whether a Go app should import this repo to access those utils. Maybe a different utils repo? Something like mattermost-plugins-apps-utils? Nevertheless, I know there is some thoughts against utils repos :sweat_smile:

mickmister: It seems unorthodox to do what I suggested, but I still think it's not a bad idea. Of course apps can be written in other languages, but I think it's fine for a plugin or app written in go to import the `mattermost-plugin-apps` repo, just as we import `mattermost-server` in the plugins. Different scenario of course, but I still like how both sides of the JWT conversation would be right next to each other. We are already importing things from the `apps` package in the `helloapp` package, so I think it's not a big deal here.

server/command/main.go
```diff
@@ -64,5 +64,5 @@ func (s *service) executeDebugLocations(params *params) (*model.CommandResponse,
 		return normalOut(params, md.MD("error"), err)
 	}
 	bytes, _ := json.Marshal(locations)
-	return normalOut(params, md.MD(string(bytes)), nil)
+	return normalOut(params, md.JSONBlock(string(bytes)), nil)
```
mickmister, line 67: You should just have to do `JSONBlock(locations)`, and it will format it properly with JSON indenting
```suggestion
	return normalOut(params, md.JSONBlock(string(bytes)), nil)
```

server/http/helloapp/locations.go
```diff
@@ -78,9 +32,7 @@ func (h *helloapp) HandleLocations(w http.ResponseWriter, req *http.Request, use
 		&apps.ChannelHeaderIconLocation{
 			Location: apps.Location{
 				LocationType: apps.LocationChannelHeaderIcon,
-				Wish: apps.Wish{
-					URL: h.AppURL(PathWishSample),
-				},
+				Wish:         *apps.NewWish(AppID, h.AppURL(PathWishSample)),
```
mickmister, line 35: Maybe make the `Wish` field a `*Wish` so it is consistent with the rest of the data model?

server/http/helloapp/locations.go
```diff
@@ -36,23 +36,23 @@ func (h *helloapp) HandleLocations(w http.ResponseWriter, req *http.Request, use
 			},
 			DropdownText: user.Username,
 			AriaText:     user.Username,
-			Icon:         icon.String(),
+			Icon:         "https://www.wizcase.com/wp-content/uploads/2020/06/Zoom-Logo.jpg",
```
mickmister, line 39: Can we have this constantized?

mickmister (changes requested):

mickmister:
Also heads up about the large amount of conflicts due to another PR 🙂

mickmister (approved):
LGTM :+1:

PR #20, opened 2020-10-11
MM-29348: Finish the install flow POC

levb:
Fixes:
MM-29348 - finish the install flow (POC)
MM-29352 - simplify context, expand, etc.

I also subsumes https://github.com/mattermost/mattermost-plugin-apps/pull/9 and fixes 
MM-28436, MM-28437 - at least, adds basic persistence.


server/apps/call_test.go
```diff
@@ -0,0 +1,39 @@
+package apps
```
levb, line 1: nm this simplistic test - there were silly issues with JSON unmarshaling, solved by restructuring. The test is not useful, but somewhat illustrative.

server/apps/expand.go
```diff
@@ -1,182 +1,209 @@
...
+		user, err := e.s.Mattermost.User.Get(e.UserID)
 		if err != nil {
-			return err
+			return nil, errors.Wrapf(err, "failed to expand user %s", e.UserID)
 		}
-		mmuser.SanitizeProfile(nil)
+		e.User = user
+	}
+
+	clone.expandedContext = expandedContext{
```
jfrerich, line 99: Interesting implementation. Send in the full model to each strip method return cloned model, or reduced cloned model if `store.ExpandSummary`.  I like this. 👍

jfrerich (approved):
@levb this looks awesome. I only have one `nit` regarding variable naming.

I also focussed on the notification changes and see that you usurped my PR into this PR. Thanks.

Approving so that I am non-blocking while out on vacation this week.  great work

server/constants/constants.go
```diff
@@ -9,7 +9,7 @@ const (
 )
 
 const (
-	BotUserName    = "appsbot"
+	BotUsername    = "appsbot"
```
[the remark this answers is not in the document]
levb: 0/5 my thinking is that the first letter of each _term_ should be capitalized, and FWIW, `username` is a term that has expectations beyond "the name of the user". (the expectation of being a unque-ish key, for instance). Also, the snake_case for it should be `username`, not `user_name`.

https://godoc.org/github.com/mattermost/platform/model#User,  `Username`

server/apps/context.go
```diff
@@ -0,0 +1,115 @@
+package apps
+
+import (
+	"github.com/mattermost/mattermost-plugin-apps/server/store"
+	"github.com/mattermost/mattermost-server/v5/model"
+)
+
+type Context struct {
+	AppID        store.AppID       `json:"-"`
```
mickmister, line 9: If `AppID` is ``json:"-"`` then the frontend will not be supplying it in the context. Should we change it so it can be taken from the frontend this way? Or should we instead have the `AppID` as part of the URL or part of the incoming `Wish`?

server/http/helloapp/install.go
```diff
@@ -39,21 +52,106 @@ func (h *helloapp) handleInstall(w http.ResponseWriter, req *http.Request, claim
...
+			team = teams[0]
+
+			// Ensure "Hallo სამყარო" channel
+			channel, _ = mmclient.GetChannelByName(AppID, team.Id, "")
+			if channel != nil {
+				// TODO DM to user that the channel has been found
+				if channel.DeleteAt != 0 {
+					return errors.Errorf("TODO unarchive channel %s \n", channel.DisplayName)
+				}
+				h.DM(data.Context.ActingUserID, "Found eisting ~%s channel.", AppID)
```
mickmister, line 81:
```suggestion
				h.DM(data.Context.ActingUserID, "Found existing ~%s channel.", AppID)
```

server/http/helloapp/install.go
```diff
@@ -39,21 +52,106 @@ func (h *helloapp) handleInstall(w http.ResponseWriter, req *http.Request, claim
...
+
+				h.DM(data.Context.ActingUserID, "Created ~%s channel.", AppID)
+			}
+
+			// Add the Bot user to the team and the channel.
+			_, api4Resp = mmclient.AddTeamMember(team.Id, data.Context.App.BotUserID)
+			if api4Resp.Error != nil {
+				return api4Resp.Error
+			}
+			_, api4Resp = mmclient.AddChannelMember(channel.Id, data.Context.App.BotUserID)
```
mickmister, line 103: Do bots need to be channel members? I feel like that crowds the channel members list if that's the case.

mickmister (approved):
Functionality looks good :+1: Just some questions and a few typos

server/apps/context.go
levb: That's an interesting point, I was not considering the client apps, yet.

We may want to choose communicating the AppID differently client->server->plugin, as a path element for instance. There is no point sending the AppID to the App itself - it knows it. Kinda silly/embarrassing/wasteful.

server/store/app.go
```diff
@@ -1,41 +1,39 @@
...
+	RootURL              string      `json:"root_url"`
 }
 
 type App struct {
-	Manifest *Manifest
+	Manifest *Manifest `json:"manifest"`
 
 	// Secret is used to issue JWT
-	Secret string
+	Secret string `json:",omitempty"`
```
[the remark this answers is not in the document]
levb: This was before we were putting name tags on them. I'll add the name now, just meant do not translate the name, but omitempty still.

server/http/helloapp/auth.go
```diff
@@ -83,46 +85,76 @@ func (h *helloapp) finishOAuth2Connect(userID string, token oauth2.Token, payloa
...
 	})
 }
 
+const AppCredentialsKey = "key_app_credentials"
+
+type AppCredentials struct {
+	BotAccessToken     string
+	BotUserID          string
+	OAuth2ClientID     string
+	OAuth2ClientSecret string
```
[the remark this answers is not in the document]
levb: This is in the Hello app... who cares? :)

server/http/helloapp/install.go
levb: Bot (accounts/access tokens) have to be added to the Teams and Channels to have access to them.

server/apps/client.go
```diff
@@ -7,41 +7,42 @@ import (
...
+func (s *service) PostNotification(n *NotificationRequest) error {
+	app, err := s.Store.GetApp(n.Context.AppID)
 	if err != nil {
-		// <><> TODO log
-		return
+		return err
 	}
 
-	resp, err := s.post(app, "", path.Join(app.Manifest.RootURL, "notify", string(sub.Subject)), msg)
+	resp, err := s.post(app, "", app.Manifest.RootURL+"/notify/"+string(n.Subject), n)
```
larkox, line 32: Nit: Why not use the "path.Join"? I would even prefer a `fmt.Sprintf` than just the `+`. That way probably the string cast can be removed.

server/apps/context.go
```diff
@@ -0,0 +1,115 @@
...
+type MattermostConfig struct {
+	SiteURL string `json:"site_url"`
+}
+
+type Thread struct {
+	ChannelID  string `json:"channel_id"`
+	RootPostID string `json:"root_post_id"`
+}
+
+func NewChannelContext(ch *model.Channel) *Context {
```
larkox, line 41: I wonder how useful will be these functions, since the context will come mainly from the client, so it will send only the ids, not the whole objects.

server/store/apps.go
```diff
@@ -0,0 +1,29 @@
+// Copyright (c) 2020-present Mattermost, Inc. All Rights Reserved.
+// See License for license information.
+
+package store
+
+import (
+	"github.com/mattermost/mattermost-plugin-apps/server/utils"
+)
+
+func (s *store) StoreApp(app *App) error {
```
larkox, line 10: We need a way to get "all installed apps". Do we have that in mind?

server/store/wish.go
```diff
@@ -0,0 +1,5 @@
+package store
+
+type Wish struct {
```
larkox, line 3: Most of the things in store feel that would make more sense in a "model" package, since they feel more definitions of the objects than information about how to store them.

larkox (approved):
As a broad it looks really good :smile:
There are a few comments, but are minimal considerations.

server/store/wish.go
levb: I thought about it, not quite ready for it; my thinking was - call it `model`, and define a KV interface in the package, with SaveToKV/LoadFromKV for the relevant structs. Separate PR?

server/store/apps.go
levb: yes, I filed https://mattermost.atlassian.net/browse/MM-29585 to address this

server/apps/context.go
levb: These are useful in notifications, the clients (should? must?) create the contexts for user-initiated actions.

PR #21, opened 2020-10-14
Wish endpoint POC

mickmister:
#### Summary

In this PR, the user must pass in the app id in the call requests context. The request is then forwarded to the specified app.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-29650

### DIFF
--- a/server/apps/client.go
+++ b/server/apps/client.go
@@ -8,6 +8,7 @@ import (
 	"io"
 	"net/http"
 	"path"
+	"strings"
 	"time"
 
 	"github.com/dgrijalva/jwt-go"
@@ -41,7 +42,16 @@ func (s *Service) PostWish(call Call) (*CallResponse, error) {
 	if err != nil {
 		return nil, err
 	}
-	resp, err := s.post(app, call.Data.Context.ActingUserID, call.Wish.URL, call.Data)
+
+	url := call.Wish.URL
+	if !strings.HasPrefix(url, "https://") && !strings.HasPrefix(url, "http://") {
+		if !strings.HasPrefix(url, "/") {
+			url += "/"
+		}
+		url = app.Manifest.RootURL + url
+	}
+
+	resp, err := s.post(app, call.Data.Context.ActingUserID, url, call.Data)
 	if err != nil {
 		return nil, err
 	}
--- a/server/http/wish/wish.go
+++ b/server/http/wish/wish.go
@@ -1,0 +1,96 @@
+package wish
+
+import (
+	"encoding/json"
+	"net/http"
+
+	"github.com/gorilla/mux"
+	"github.com/pkg/errors"
+
+	"github.com/mattermost/mattermost-plugin-apps/server/apps"
+	"github.com/mattermost/mattermost-plugin-apps/server/constants"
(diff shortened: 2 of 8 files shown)


server/apps/client.go
```diff
@@ -41,7 +42,16 @@ func (s *Service) PostWish(call Call) (*CallResponse, error) {
...
 		return nil, err
 	}
-	resp, err := s.post(app, call.Data.Context.ActingUserID, call.Wish.URL, call.Data)
+
+	url := call.Wish.URL
+	if !strings.HasPrefix(url, "https://") && !strings.HasPrefix(url, "http://") {
+		if !strings.HasPrefix(url, "/") {
+			url += "/"
+		}
+		url = app.Manifest.RootURL + url
```
mickmister, line 51: If the url passed in is a relative URL, we must mount it onto the app's root URL

server/http/wish/wish.go
```diff
@@ -0,0 +1,96 @@
...
+	w := wish{
+		apps: apps,
+	}
+
+	subrouter := router.PathPrefix(constants.WishPath).Subrouter()
+	subrouter.HandleFunc("", w.handleWish).Methods("POST")
+}
+
+func (wish *wish) verifyCallContext(actingUserID string, ctx apps.CallContext) error {
+	if ctx.ChannelID == "" {
```
larkox, line 29: A channel ID is not always needed IMO. For example, a element on the main menu is not associated to any channel.

server/http/wish/wish.go
```diff
@@ -0,0 +1,96 @@
...
+
+func (wish *wish) handleWish(w http.ResponseWriter, req *http.Request) {
+	var err error
+
+	call := apps.Call{}
+	defer req.Body.Close()
+
+	err = json.NewDecoder(req.Body).Decode(&call)
+	if err != nil {
+		err = errors.Wrap(err, "Failed to unmarshal Call struct")
```
larkox, line 50: As a go standard, errors should not start with uppercase.

larkox (approved):
In general looks good to me. I will adapt my implementation to fit this.

server/http/wish/wish.go
mickmister: @larkox It's likely not needed in the case of the main menu but we still have it available to provide in all cases, and it allows us to know what team they're currently viewing, may not be needed for a given request either of course. I think it's better to provide as much information as possible (like the channel id) for all calls, especially when it's not expensive.

larkox: My worry is that since here we are requiring it, if we are in some context that does not make sense having one (the backstage for example, if we ever use it), it will create problems. I am all in for if we have a channelID, to check if we are part of that channel, but not requiring the channelID to be different to "".

levb:
Absorbed into #23

server/apps/client.go
mickmister: @levb I see this fix didn't make it into master. I also realized that I have a typo here, as it should have been `url = "/" + url`

How do we handle the case when a relative URL is used in the `call.FormURL`?

levb:
@mickmister RE: relative URLs. 1/5 we stick with FQ URLs for now, and consider shortcuts once we reach a bit more stability. Is there a use case that requires relative URLs presently?

PR #22, opened 2020-10-16
Handle embedded forms

larkox:
#### Summary
Handle embedded forms from the proxy

#### Ticket Link
None

#### Related Pull Requests
Redux: https://github.com/mattermost/mattermost-redux/pull/1262
Webapp: https://github.com/mattermost/mattermost-webapp/pull/6804


server/http/helloapp/embedded_wish.go
```diff
@@ -0,0 +1,196 @@
...
+					Default:     "default text",
+					Placeholder: "placeholder",
+					HelpText:    "This a regular input in an interactive dialog triggered by a test integration.",
+				},
+			},
+		},
+	}
+}
+
+func (h *helloapp) getDialogFullSample() model.OpenDialogRequest {
```
mickmister, line 82: It looks like this function is not called. Is this here to toggle between the two dialogs by editing the source here?

server/http/restapi/embedded_form.go
```diff
@@ -0,0 +1,166 @@
...
+	if dialogRequest.Submission[embeddedSubmissionAppIDKey] == nil {
+		writeDialogError(w, "No App ID provided")
+		return
+	}
+
+	appID, ok := dialogRequest.Submission[embeddedSubmissionAppIDKey].(string)
+	if !ok {
+		writeDialogError(w, "App ID not recognized")
+		return
+	}
```
mickmister, line 49: We can get rid of the first block here as well

server/http/restapi/embedded_form.go
```diff
@@ -0,0 +1,166 @@
...
+			if errors, ok := resp.Data["errors"].(map[string]interface{}); ok {
+
+				for key, value := range errors {
+					if svalue, ok := value.(string); ok {
+						dialogResponse.Errors[key] = svalue
+					}
+				}
+			}
+		}
+		dialogResponse.Error = resp.Error
```
mickmister, line 90: Do we want to return an error code in the response here, or does the response need to be 200?

server/http/restapi/embedded_form.go
```diff
@@ -0,0 +1,166 @@
...
+	} else {
+		if resp.Data["post"] != nil {
+			if updatePost, parseErr := postFromInterface(resp.Data["post"]); parseErr == nil {
```
mickmister, line 93:
```suggestion
			if updatedPost, parseErr := postFromInterface(resp.Data["post"]); parseErr == nil {
```

server/http/restapi/embedded_form.go
```diff
@@ -0,0 +1,166 @@
...
+
+	return &post, nil
+}
+
+func (a *api) UpdatePost(postID string, post *model.Post) error {
+	// If the updated post does contain a replacement Props set, we still
+	// need to preserve some original values, as listed in
+	// model.PostActionRetainPropKeys. remove and retain track these.
+	remove := []string{}
+	retain := map[string]interface{}{}
```
mickmister, line 132: Can you link to the source of this comment here? https://github.com/mattermost/mattermost-server/blob/20491c2585475c2218f964e0a882c65deac570a5/app/integration_action.go#L57

mickmister (approved):
Awesome work, just some non-blocking questions/suggestions on cleaning up code

server/http/helloapp/embedded_wish.go
larkox: Exactly. It is in case you want to trigger a fuller example.

server/http/restapi/embedded_form.go
larkox: I believe this is expected to work with a 200 response, not an error response. The thinking behind is "the transaction was succesful, and you have to show an error message".

mickmister (approved):
LGTM great work :+1:

server/http/helloapp/embedded_wish.go
```diff
@@ -0,0 +1,197 @@
...
+)
+
+func (h *helloapp) handleCreateEmbedded(w http.ResponseWriter, req *http.Request, claims *apps.JWTClaims, data *api.Call) (int, error) {
+	err := h.asBot(func(mmclient *model.Client4, botUserID string) error {
+		post := &model.Post{
+			Message:   "Debug form",
+			ChannelId: data.Context.ChannelID,
+			UserId:    botUserID,
+		}
+		post.AddProp("appID", AppID)
```
levb, line 19: constants.AppID

levb: A hypothetical question, can we envision multiple embedded forms added to a post? Should AppID be embedded in the Form?

Either way I don't like having `AppID` as a top-level prop, because it relates that a post "belongs to" an App, which should not be the case.

server/http/helloapp/embedded_wish.go
```diff
@@ -0,0 +1,197 @@
...
+		post.AddProp("dialog", h.getDialogSmallSample())
```
levb, line 20: we should decide on the term and be consistent. Perhaps, `"embedded_form"`? also, `constants.XXX`

server/http/restapi/embedded_form.go
```diff
@@ -0,0 +1,166 @@
...
+)
+
+const (
+	embeddedSubmissionPostIDKey = "mm_postID"
+	embeddedSubmissionAppIDKey  = "mm_appID"
+)
+
+func (a *api) handleEmbeddedForm(w http.ResponseWriter, req *http.Request, userID string) {
+	var dialogRequest model.SubmitDialogRequest
+	err := json.NewDecoder(req.Body).Decode(&dialogRequest)
```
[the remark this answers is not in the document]
levb: No need to close req.Body, ever, the http server takes care of that. In fact, it may be a counter-pattern, messing with connection management.

levb (changes requested):
I am going to hold off a more detailed review until master is merged

server/http/helloapp/embedded_wish.go
larkox: I do not like much the idea of several plugins displaying a dialog in the same message. Also, I think the post does indeed belong to the app, since it is the one posting it as a form, and will be the one modifying it.

server/http/helloapp/embedded_wish.go
larkox: It is the same as an interactive dialog. So we should unify the two concepts. Here some proposals:
- Interactive Dialog - Embedded Dialog
- Modal Form - Embedded Form

levb (approved):
I still have a bunch of questions, but this is a big step forward.

server/api/call.go
```diff
@@ -25,12 +25,12 @@ const (
 )
 
 type CallResponse struct {
-	Type CallResponseType
+	Type CallResponseType `json:"type,omitempty"`
 
-	Markdown md.MD       `json:"markdown,omitempty"`
-	Data     interface{} `json:"data,omitempty"`
+	Markdown md.MD                  `json:"markdown,omitempty"`
+	Data     map[string]interface{} `json:"data,omitempty"`
```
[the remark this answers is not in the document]
larkox: I personally prefer a map[string]interface, since it is easier to handle. I understand that for this PR, i could have added the Post or the Error list directly as the Data, but I still feel more comfortable using map[string]interface{}. Nevertheless, I am open to revert back that change.

server/api/embedded.go
```diff
@@ -0,0 +1,9 @@
+package api
+
+const (
+	EmbeddedPostPropAppID  = "appID"
```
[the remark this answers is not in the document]
larkox: Since this one is strictly used on embedded forms, I wonder if it makes sense to have it separated. But I am open to unify.

server/api/call.go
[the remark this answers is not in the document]
levb: IMO, `Data` is to return any data the app wants to return from a successful invocation, usually some sort of a JSON result. The pointer here should be as generic as possible, thus interface{}.

For other use-cases (returning the Form, for example) I am inclined to add typed, dedicated fields. Anyway, we can decide on this later. Do a separate cleanup when we have more use-cases implemented.

server/api/embedded.go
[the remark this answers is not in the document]
levb: Gotcha. 0/5 if it walks like a duck, if it quacks like a duck, it is an AppID (aka app_id). Not an essential issue, obviously.

server/http/helloapp/embedded_wish.go
levb: 0/5 log as a separate issue to discuss/resolve later?

server/http/helloapp/embedded_wish.go
levb: Why should we ultimately support the old-style interactive dialogs? I thought it was temporary "stub", to be replaced later?

PR #23, opened 2020-10-20
Call endpoint (aka Wish proxy, restructured)

levb:
#21 re-worked

- Replaced `Wish` with `FormURL`
- Combined `Call` and `CallRequest`
- Added `API.Call` and `Client.PostCall`
- Wrapped `Client` implementation into `client` to avoid namespace collisions
- Simplified/tweaked some names
- Added HTTP `/proxy/call` endpoint (should be `/proxy/form`?)


server/http/helloapp/auth.go
```diff
@@ -70,25 +70,24 @@ func (h *helloapp) startOAuth2Connect(userID string, callOnComplete apps.Call) (
...
 	//  - startOAuth2Connect is a Call
 	//  - payload for finish should be a Call
-	//  - a Wish can check the presence of the acting user's OAuth2 token, and
+	//  - a Call can check the presence of the acting user's OAuth2 token, and
 	//    return Call startOAuth2Connect(itself)
 	// for now hacking access to apps object and issuing the call from within
 	// the app.
 
-	cr, _ := h.apps.Client.PostWish(&call)
+	cr, _ := h.apps.Client.PostCall(call)
```
larkox, line 87: I know this is no new code, but wondering now. Why the helloapp cast a call when it finishes the OAuth connect? Wouldn't it simpler to just "do something", instead of going through the process to perform an http call to itself?

server/http/helloapp/locations.go
```diff
@@ -37,7 +36,7 @@ func (h *helloapp) HandleLocations(w http.ResponseWriter, req *http.Request, use
 		&apps.ChannelHeaderIconLocation{
 			Location: apps.Location{
 				LocationType: apps.LocationChannelHeaderIcon,
-				Wish:         store.NewWish(AppID, h.AppURL(PathWishPing)),
+				FormURL:      h.AppURL(PathPing),
```
larkox, line 39: Since wishes no longer hold the AppID, we have to think where to include the AppID for the locations, in order for the client to know how to fill up the context of the call. (i.e. the appID)

larkox (approved):
LGTM, just some questions and one consideration.

server/http/helloapp/locations.go
levb: Will add `AppID` elsewhere, tbd.

server/http/helloapp/auth.go
levb: @larkox and I discussed offline how it should be put in the OAuther, as a `Call`

mickmister (approved):
LGTM :+1:

PR #26, opened 2020-10-24
Support slash command forms

mickmister:
Slash commands are declared in the locations response. The definitions for the subcommands are then fetched at runtime when the user types the base command.

#### Related PRs

https://github.com/mattermost/mattermost-webapp/pull/6912


server/http/helloapp/auth.go
```diff
@@ -157,3 +157,21 @@ func (h *helloapp) DM(userID string, format string, args ...interface{}) {
 		Message:   fmt.Sprintf(format, args...),
 	})
 }
+
+func (h *helloapp) Ephemeral(userID, channelID string, format string, args ...interface{}) {
```
larkox, line 161: You can use the AsBot function to do this, so you don't have to deal with the tokens.

larkox: I am moving all these functions to a `client.go` file. I will send you the link when I have that finished.

server/http/helloapp/command.go
```diff
@@ -0,0 +1,153 @@
...
+	AppID       string        `json:"app_id"`
+	Pretext     string        `json:"pretext"`
+	Description string        `json:"description"`
+	SubCommands []SubCommand  `json:"sub_commands"`
+	Args        []interface{} `json:"args"`
+	FormURL     string        `json:"form_url"`
+}
+
+func (h *helloapp) handleHelloCommandSubmission(w http.ResponseWriter, req *http.Request, claims *apps.JWTClaims, call *apps.Call) (int, error) {
+	var e = func(err error, msg string) (int, error) {
```
larkox, line 25: Wonder how useful is this function, so we have it outside of this function and named. If we are going to create more commands handlers may be interesting (even if we are just using them in the sample helloapp).

server/http/helloapp/command.go
```diff
@@ -0,0 +1,153 @@
...
+		httputils.WriteJSON(w, apps.CallResponse{
+			Type:     apps.CallResponseTypeError,
+			Markdown: md.MD(errors.Wrap(err, msg).Error()),
+		})
+		return http.StatusInternalServerError, err
+	}
+
+	username := call.Values.Get("user")
+	message := call.Values.Get("message")
+	u, err := h.apps.Mattermost.User.GetByUsername(username)
```
larkox, line 35: We shouldn't use the mattermost api in the helloapp (we should use the way the app would to get this info), but I understand we can leave it like this for now.

server/http/helloapp/command.go
```diff
@@ -0,0 +1,153 @@
...
+	var e = func(err error) (int, error) {
+		h.DM(call.Context.ActingUserID, err.Error())
+		httputils.WriteJSON(w, apps.CallResponse{
+			Type:  apps.CallResponseTypeError,
+			Error: err,
+		})
+		return http.StatusInternalServerError, err
+	}
+
+	users, err := h.apps.Mattermost.User.ListInChannel(call.Context.ChannelID, "username", 0, 50)
```
larkox, line 120: Do we always just want to look for users in the channel? I cannot run for example `/todo send @someone something` from a DM with the bot?

larkox (commented):
Several comments.

server/http/helloapp/auth.go
mickmister: Sure sounds good :+1:

server/http/helloapp/command.go
mickmister: Yes this was for quick debugging, and should be removed from this function

server/http/helloapp/command.go
mickmister: This was just a demonstration to show that the context is being properly passed in and used. It's not really a practical feature of course.

server/http/helloapp/command.go
```diff
@@ -0,0 +1,153 @@
...
+	data := SubCommand{
+		Pretext: "/hello",
+		AppID:   AppID,
+		SubCommands: []SubCommand{
+			{
+				AppID:       AppID,
+				Pretext:     "sendto",
+				FormURL:     PathCommandSubmission,
+				Description: "Send a message to someone in this channel",
+				Args: []interface{}{
```
mickmister, line 62: @levb This should probably be [`Autocomplete`](https://github.com/mattermost/mattermost-plugin-apps/blob/640b40537497565ab42778e2f0679afb7724e55e/server/apps/autocomplete.go#L3) or [`Form`](https://github.com/mattermost/mattermost-plugin-apps/blob/640b40537497565ab42778e2f0679afb7724e55e/server/apps/form.go#L3) instead of just an array of the form arguments.

mickmister:
Outdated

PR #27, opened 2020-10-26
Locations -> Bindings

levb:
See https://github.com/mattermost/mattermost-plugin-apps/pull/27/files#diff-0009600598e7a8e6fad5e528e92ffd816900bb89aae2ef72ab822c00ffb0de95 for what the app's `Binding`s look like. https://github.com/mattermost/mattermost-plugin-apps/pull/27/files#diff-ccfd1477cb4624d9e6785dc801779accb4c19f5695acc61c377313dc7ca3f8c4R14 for the `Form` (Modal/Autocomplete combined).

- type `api.Binding` is used to bind functions/forms to known top-level locations, and to define sub-locations (commands, menu items, etc.) It is a single, combined struct for the union of all use-cases. It binds a `LocationID` to an `api.Call`, or a list of sub-locations.
- type `api.Function` is introduced, for the time being. May prove unnecessary, the handling (caching?) of per-function `Expand` will be reconsidered as serving and updating Forms is implemented.
- type `api.Field` represents fields in a modal or autocomplete. It is a single, combined struct for the union of all use-cases and field types. 
- all `Call` values now must be `string`s. 
- `api.GetBindings()` returns consolidated bindings for all Apps (`GET /api/v1/bindings`)
- Combined Fiel
- `/install` and `/bindings` are now a convention, no need in the manifest
- temporarily removed `From` from `Call` until we figure it out, the concept still appears useful
- style

cc @larkox


larkox (approved):

server/apps/bindings.go
```diff
@@ -0,0 +1,65 @@
...
+	for _, b := range bb {
+		b.AppID = appID
+		if len(b.Bindings) != 0 {
+			setAppID(b.Bindings, appID)
+		}
+	}
+}
+
+// This and registry related calls should be RPC calls so they can be reused by other plugins
+func (s *service) GetBindings(cc *api.Context) ([]*api.Binding, error) {
```
levb, line 42: TODO: Make sure top-level elements can not be overwritten, maybe switch around the defaulting - first one wins, no overrides, only merging children

server/api/form.go
```diff
@@ -1,71 +1,19 @@
...
-	// options - encoding, regexp, etc.
-}
-
-type SelectOption struct {
-	Label string `json:"label"`
-	Value string `json:"value"`
+	// DependsOn is the list of field names that when changed force reloading
+	// the form. Values of the fields with values that are not included in the
+	// refreshed form are lost.
+	DependsOn        []string `json:"depends_on,omitempty"`
```
levb, line 12: Maybe unnecessary on `Form` since the intent is to always fetch it before displaying.

server/apps/bindings.go
```diff
@@ -0,0 +1,65 @@
+package apps
```
levb, line 1: Resolve: the app's `/command/X` should translate into `/command/appCommand/X`

server/apps/call.go
```diff
@@ -3,15 +3,19 @@ package apps
 import "github.com/mattermost/mattermost-plugin-apps/server/api"
 
 func (s *service) Call(call *api.Call) (*api.CallResponse, error) {
-	var err error
+	// TODO <><> get a cached function
```
levb, line 6: Really, Expand will likely end up served with the form definition, back into `Call`?

server/api/binding.go
```diff
@@ -0,0 +1,45 @@
...
+	RoleID           string `json:"role_id,omitempty"`
+	DependsOnTeam    bool   `json:"depends_on_team,omitempty"`
+	DependsOnChannel bool   `json:"depends_on_channel,omitempty"`
+	DependsOnUser    bool   `json:"depends_on_user,omitempty"`
+	DependsOnPost    bool   `json:"depends_on_post,omitempty"`
+
+	// A Binding is either to a Call, or is a "container" for other locations -
+	// i.e. menu sub-items or subcommands.
+	Call     *Call      `json:"call,omitempty"`
+	Bindings []*Binding `json:"bindings,omitempty"`
```
mickmister, line 44: 💯

server/http/helloapp/bindings.go
```diff
@@ -0,0 +1,60 @@
...
+						},
+					},
+				},
+			}, {
+				LocationID: api.LocationPostMenu,
+				Bindings: []*api.Binding{
+					{
+						LocationID:  "message",
+						Description: "message a user",
+						Call:        &modal,
```
mickmister, line 54: In this binding, `Call` has the value of
```go
{
	URL: "/message"
    AsModal: true,
}
```

I could be completely wrong, but is this supposed to tell client, "you need to open a modal, and the definition can be fetched with this URL." If this is the case, the level of granularity is pretty low, rather than a string that indicates to open a modal. I have a feeling I'm just misunderstanding.

What is expected to happen when `AsModal = false`?

server/http/helloapp/bindings.go
```diff
@@ -0,0 +1,60 @@
...
+		[]*api.Binding{
+			{
+				LocationID: api.LocationCommand,
+				Bindings: []*api.Binding{
+					{
+						LocationID:  "message",
+						Hint:        "[--user] message",
+						Description: "send a message to a user",
+						Call:        call,
+					}, {
```
mickmister, line 29: If I understand correctly, what's going on here is we are telling the client:

"Load this sub command. If `/hello message` it is typed, fetch the form from the hello app from the `/message` endpoint." Which is the same endpoint that is used for the modal defined below.

mickmister (approved):
This is great, I like it a lot. No issues found, just some questions for a better understanding.

server/apps/bindings.go
mickmister: I'm not sure what you mean by this. Are you saying there's an endpoint with that structure?

server/apps/bindings.go
levb: Filed https://mattermost.atlassian.net/browse/MM-30036 to cover separately, in sync with Autocomplete PRs

server/apps/bindings.go
levb: https://mattermost.atlassian.net/browse/MM-30036

server/http/helloapp/bindings.go
levb: Discussed offline, still TBD

server/http/helloapp/bindings.go
levb: Discussed offline, basically yes, but still TBD on the details. Looking forward to the first implementation of Autocomplete or Modal to tie it all together.

PR #29, opened 2020-10-29
Small updates to Binding/Form to support client-side commands

mickmister:
#### Summary

`Field.Position` has been added, which is a one-based positional argument indicator
`SelectOption.IconData` has been added

Hello app is updated to use these changes

#### Related Pull Requests

https://github.com/mattermost/mattermost-webapp/pull/6912

### DIFF
--- a/server/api/binding.go
+++ b/server/api/binding.go
@@ -17,11 +17,11 @@ type Binding struct {
 	// For PostMenu, ChannelHeader locations specifies the icon.
 	Icon string `json:"icon,omitempty"`
 
-	// Label is the (usually short) primary text to display at the location.
+	// Name is the (usually short) primary text to display at the location.
 	// - For LocationPostMenu is the menu item text.
 	// - For LocationChannelHeader is the dropdown text.
 	// - For LocationCommand is the name of the command
-	Label string `json:"text,omitempty"`
+	Name string `json:"name,omitempty"`
 
 	// Hint is the secondary text to display
 	// - LocationPostMenu: not used
--- a/server/api/binding.go
+++ b/server/api/binding.go
@@ -17,11 +17,11 @@ type Binding struct {
 	// For PostMenu, ChannelHeader locations specifies the icon.
 	Icon string `json:"icon,omitempty"`
 
-	// Name is the (usually short) primary text to display at the location.
+	// Label is the (usually short) primary text to display at the location.
 	// - For LocationPostMenu is the menu item text.
 	// - For LocationChannelHeader is the dropdown text.
 	// - For LocationCommand is the name of the command
-	Name string `json:"name,omitempty"`
+	Label string `json:"label,omitempty"`
 
 	// Hint is the secondary text to display
 	// - LocationPostMenu: not used
--- a/server/api/field.go
+++ b/server/api/field.go
@@ -33,7 +33,7 @@ type Field struct {
 	// number.
 	AutocompleteLabel string `json:"autocomplete_label"`
 	AutocompleteHint  string `json:"hint"`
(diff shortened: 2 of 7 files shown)


server/api/field.go
```diff
@@ -32,6 +33,7 @@ type Field struct {
 	// number.
 	AutocompleteLabel string `json:"autocomplete_label"`
 	AutocompleteHint  string `json:"hint"`
+	Positional        bool   `json:"positional"`
```
levb, line 36: Right, so my idea of setting `Name` to $1 obviously won't work because we still need a Name for JSON. What about having a `PositionalAlias="$1"`?

3/5 we should not rely on the order of `Field`s in the array for positional args, the order is already used for Modal, and may not be a match. Explicit assignment for positional args would be more accomodating,

server/api/binding.go
```diff
@@ -17,11 +17,11 @@ type Binding struct {
 	// For PostMenu, ChannelHeader locations specifies the icon.
 	Icon string `json:"icon,omitempty"`
 
-	// Label is the (usually short) primary text to display at the location.
+	// Name is the (usually short) primary text to display at the location.
```
levb, line 20: 0/5. I chose `Label` because `Name` is used in `Field` as the JSON field (ID). Also, if we rename `Label`->`Name` here we should rename `Field.AutocompleteLabel` accordingly (`Field.AutocompleteFlagName`?) since it means the same thing. FWIW, I _like_ label since it's more indicative of the "Display Name" nature of it to me.

server/api/field.go
```diff
@@ -12,8 +12,9 @@ const (
 )
 
 type SelectOption struct {
-	Label string `json:"label"`
-	Value string `json:"value"`
+	Label    string `json:"label"`
+	Value    string `json:"value"`
+	IconData string `json:"icon_data"`
```
levb, line 17: OT: Can we get a recommendation from the (Web, Mobile)Apps team as to the best practice of handling icons - 1/5 we should adopt either embedded data, or URLs, consistently in (most) contexts.

server/apps/call.go
```diff
@@ -9,13 +9,19 @@ func (s *service) Call(call *api.Call) (*api.CallResponse, error) {
 		return nil, err
 	}
 
+	if call.AsModal {
```
levb, line 12: :) 👍

levb (changes requested):

server/api/field.go
mickmister: @levb FYI we pass the data to the React component the same way whether it's data or a URL, and it "just works". i.e.

```jsx
<img src={'data:iojriqjwiqj0iefjwe...'}/>

<img src={'https://images.com/image.png'}/>
```

server/api/field.go
mickmister: @levb @larkox I'm thinking we should go with ints, as they are indexes in the "array" of positional arguments. So I'm proposing we have a `Position int` field.

server/api/binding.go
mickmister: Yeah `Label` sounds good

levb (approved):
wondering why, but sure. FWIW, I am making "OK" to be ""

mickmister:
> wondering why, but sure. FWIW, I am making "OK" to be ""

@levb I'm not sure what this means. Assuming you are referring to the placement of `Form`:

Having `Form` on the top level of the `CallResponse` was something we recently discussed in passing on a zoom call, but it has not been solidified. I had to move the `Form` out of `Data`, because that field is now `map[string]string` and cannot support the form.

PR #28, opened 2020-10-26
Hello app with locations

larkox:
#### Summary
Here is a first approach to the hello app with some locations registered.

It still need to merge with the latest changes from Lev.

In order to use, both POCApps and EmbeddedForms branches have to be merged in webapp and redux repos.

PR #32, opened 2020-11-07
Restructured Call/Form, adjusted the app (aka Hello App with Locations)

levb:
Related PRs:
https://github.com/mattermost/mattermost-webapp/pull/7017
https://github.com/mattermost/mattermost-redux/pull/1287

This PR takes in (most?) elements of https://github.com/mattermost/mattermost-plugin-apps/pull/28 along with some re-structuring.
- Removed the current Interactive Dialog-based stubs, to be replaced with `Modal`s soon
- Eliminated `Function` and `GetFunctionMetadata` - now use `Call[Response]TypeForm`

Call changes
- Added `CallType` as `Submit`, `Form`, `Cancel`
- Added `CallResponseTypeForm` to `CallResponseType`. When returned in response to a `Submit` request, indicates the intent to open an interactive form (Modal).
- Moved `Expand` back into `Call`
- Added `LocationID` to `Context` (previously known as `From`, for now just a single value
- Added `LocationInPost` to available top-level locations (for embedded forms)
- Made all `Call`s go through the `API` layer (were direct invocations of `Cient.Call`)

Form/Field changes:
- `s/AutocompleteLabel/Label` (JSON: `label`) in `Field`
- Added `SubmitButtons string` that can point to a static or dynamic select (by name) to use as submit buttons, "" use OK
- Added `CancelButton bool` that adds a default "Cancel" button
- Added `SubmitOnCanel bool`

Embedded form changes:
- Defined `app_bindings` Prop on Posts, as `[]*Binding`
- Added `Form` to `Binding` to support embedding forms in posts


jfrerich (approved):

server/api/call.go
```diff
@@ -7,36 +7,57 @@ import (
 	"github.com/mattermost/mattermost-plugin-apps/server/utils/md"
 )
 
+type CallType string
+
+const (
+	// CallTypeSubmit (default) indicates the intent to take action.
+	CallTypeSubmit = CallType("")
+	// CallTypeForm retrieves the form definition for the current set of falues,
```
mickmister, line 15:
```suggestion
	// CallTypeForm retrieves the form definition for the current set of values,
```

server/api/call.go
```diff
@@ -7,36 +7,57 @@ import (
...
+	CallTypeForm = CallType("form")
+	// CallTypeCancel is used for for the (rare?) case of when the form with
+	// SubmitOnCancel set is dismissed by the user.
+	CallTypeCancel = CallType("cancel")
+)
+
 type Call struct {
 	URL        string            `json:"url,omitempty"`
+	Type       CallType          `json:"type,omitempty"`
 	Values     map[string]string `json:"values,omitempty"`
```
mickmister, line 26: Something to keep in mind with using `map[string]string`, we will have multiselect values which will probably need to be sent as `[]string`. What are the pros of going with `map[string]string` as the container for the values, versus `map[string]interface{}`?

server/http/helloapp/bindings.go
```diff
@@ -12,51 +12,76 @@ import (
...
+					Hint:        "[--user] message",
+					Description: "send a message to a user",
+					Call:        sendSurvey,
+				}, {
+					LocationID:  "manage",
+					Hint:        "subscribe | unsubscribe ",
+					Description: "manage channel subscriptions to greet new users",
+					Bindings: []*api.Binding{
+						{
+							Label:       "subscribe",
```
mickmister, line 69: Here we are using `Label` instead of `LocationID`. Is this intentional?

server/http/helloapp/f_survey.go
```diff
@@ -0,0 +1,54 @@
...
+	// userID := c.GetValue(fieldUserID, c.Context.ActingUserID)
+	message := c.GetValue(fieldMessage, "default hello message")
+
+	switch c.Type {
+	case api.CallTypeForm:
+		out = h.newSurveyFormResponse(message)
+
+	case api.CallTypeSubmit:
+		// TODO post something somewhere; for embedded form - what do we do?
+		out = &api.CallResponse{}
```
mickmister, line 42: Per your comment here, maybe need a way to clean up / close an embedded form if appropriate?

mickmister (approved):
A few comments, looks good :+1:

server/apps/client.go
```diff
@@ -101,6 +100,7 @@ func (c *client) post(toApp *api.App, fromMattermostUserID string, url string, m
 		return nil, err
 	}
 	req.Header.Set(OutgoingAuthHeader, "Bearer "+jwtoken)
+	req.Header.Set("Content-Type", "application/json")
```
jfrerich, line 103: @levb, I added this commit to the PR.  This is needed for clients receiving the post with `json` data structures

server/apps/bindings.go
```diff
@@ -56,7 +58,7 @@ func (s *service) GetBindings(cc *api.Context) ([]*api.Binding, error) {
 
 		// TODO eliminate redundant AppID, just need it at the top level? I.e.
 		// group by AppID instead of top-level LocationID
-		setAppID(bb, appID)
+		setAppID(bb, appID, true)
```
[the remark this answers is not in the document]
levb: 0/5 Let's maybe file a separate ticket? It can be just a plugin-apps/redux change, right? Also, has no impact on the Apps themselves

server/api/call.go
levb: ack. A separate change when we need it?

server/http/helloapp/bindings.go
levb: 0/5 for subcommands LocationID and Label are always the same, so it'd be nice to avoid duplication, and only require 1 field. My original thinking was to make the Label optional and default it to LocationID. Then we discussed it offline, and flipped it so the LocationID defaults to label. We should probably discuss the semantics of the binding props offline with @larkox.

server/http/helloapp/f_survey.go
levb: There's a ticket for @aaronrothschild and @abhijit-singh to define the embedded form submit behavior. 2/5 embedded forms should be immutable *including no support for `RefreshOnChange`, and collapse upon a successful submission. Results should be displayed as a separate message.

server/apps/bindings.go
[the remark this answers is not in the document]
mickmister: https://mattermost.atlassian.net/browse/MM-30461

server/http/helloapp/f_send_survey.go
```diff
@@ -0,0 +1,89 @@
...
+func (h *helloapp) sendSurvey(userID, message string) {
+	p := &model.Post{
+		Message: "Please respond to this survey",
+	}
+	p.AddProp(apps.PropAppBindings, []*apps.Binding{
+		{
+			LocationID: "survey",
+			Form:       h.newSurveyForm(message),
+		},
+	})
```
mickmister, line 87: When the front processes the props and sees `location_id: "survey"` in the bindings, what does this convey to the frontend? Is it assumed this embedded binding in the post be rendered as an embedded form because `form` is present?

server/http/helloapp/f_send_survey.go
```diff
@@ -0,0 +1,89 @@
...
+					Type:             apps.FieldTypeText,
+					IsRequired:       true,
+					Description:      "Text to ask the user about",
+					Label:            "message",
+					AutocompleteHint: "Anything you want to say",
+					ModalLabel:       "Text",
+					TextMinLength:    2,
+					TextMaxLength:    1024,
+				},
+			},
```
mickmister, line 40: `Field.Value` is currently how we set the default value of a field. Are we planning on having that property be the source of truth value on a refresh? Or are we going to have a separate `map[string]string` sent to the client for the form refresh?

server/http/helloapp/bindings.go
levb: Why don't we stick to them being separate things, Location is required, and _maybe_ default Label to Location if it's empty.

server/http/helloapp/f_send_survey.go
levb: It is only a "reference" to send back to the App in context.Location

But... it may be useful to map out all UX form containers as Locations, may come useful later.

server/http/helloapp/f_send_survey.go
levb: I was thinking, at least for now - the former (ack that we may need to extend it)

levb:
@mickmister @larkox any objections to merging this now? This is the de-facto master ATM even if imperfect...

larkox:
I haven't got the chance to properly review it, but I trust you and @mickmister and @jfrerich 's reviews, so it is ok for me :D

PR #33, opened 2020-11-07
Style

levb:
- Moved things around, no functional changes
  - /api -> apps
  - /apps -> apps/impl
  - /store -> apps/store
- Small style fixes


server/http/helloapp/n_user_joined_channel.go
```diff
@@ -3,12 +3,11 @@ package helloapp
 import (
 	"net/http"
 
-	"github.com/mattermost/mattermost-plugin-apps/server/api"
 	"github.com/mattermost/mattermost-plugin-apps/server/apps"
 )
 
-func (h *helloapp) handleUserJoinedChannel(w http.ResponseWriter, req *http.Request,
-	claims *apps.JWTClaims, n *api.Notification) (int, error) {
+func (h *helloapp) nUserJoinedChannel(w http.ResponseWriter, req *http.Request,
```
jfrerich, line 9: It's pretty intuitive using "f_xx", but I'm not sure "n_" is.  just a note and that we might consider.  Non-blocking

jfrerich (commented):
LGTM!

server/http/helloapp/n_user_joined_channel.go
levb: these are "functions" and "notifications", probably worth expanding a bit 👍

server/apps/api.go
```diff
@@ -0,0 +1,55 @@
...
+
+type JWTClaims struct {
+	jwt.StandardClaims
+	ActingUserID string `json:"acting_user_id,omitempty"`
+}
+
+type InInstallApp struct {
+	GrantedPermissions Permissions
+	AppSecret          string
+	OAuth2TrustedApp   bool
```
mickmister, line 48: Just an aside: If these are serialized, should we have json tags on the fields?

mickmister (approved):
LGTM :+1:

server/http/helloapp/n_user_joined_channel.go
levb: Added a comment

server/apps/api.go
levb: done

PR #34, opened 2020-11-09
WIP Apps in Config, tune GetBindings

levb:


mickmister (commented):
No issues found upon first review. Ack that the PR is in draft

server/apps/app.go
```diff
@@ -32,46 +33,77 @@ type App struct {
...
+	m.AppID = AppID(appID)
+	m.DisplayName, _ = c["display_name"].(string)
+	m.Description, _ = c["description"].(string)
+	m.OAuth2CallbackURL = c["oauth2_callback_url"].(string)
+	m.HomepageURL = c["homepage_url"].(string)
+	m.RootURL = c["root_url"].(string)
+	m.RequestedPermissions = permissionsFromConfigArray(c["requested_permissions"])
+	m.RequestedLocations = locationsFromConfigArray(c["requested_locations"])
+	return m
+}
```
larkox: @levb Any strong reason to not do this as something like:
```
b, _ := json.Marshal(in)
var m Manifest
_ := json.Unmarshal(b, &m)
return m
```
It is simpler at needs less maintenance (if we were to add or remove any field to the manifest, we have to change also this).

larkox: A similar approach can be done for all the functions in this file.

larkox: Another solution would be to create our own function using reflection to check the json tag and build the object, but why to reinvent the wheel?

server/apps/impl/bindings.go
```diff
@@ -70,21 +66,26 @@ func (s *service) scanAppBindings(app *apps.App, bindings []*apps.Binding, locPr
...
 			b.AppID = app.Manifest.AppID
 		}
 
 		if len(b.Bindings) != 0 {
-			scanned, err := s.scanAppBindings(app, b.Bindings, fql)
-			if err != nil {
-				return nil, err
+			scanned := s.scanAppBindings(app, b.Bindings, fql)
+			if len(scanned) == 0 {
+				// We do not add bindings without any valid sub-bindind
```
levb, line 81: 👍

server/apps/app.go
```diff
@@ -42,68 +44,16 @@ type App struct {
...
-		"oauth2_client_id":     a.OAuth2ClientID,
-		"oauth2_client_secret": a.OAuth2ClientSecret,
-		"oauth2_trusted_app":   a.OAuth2TrustedApp,
-		"bot_user_id":          a.BotUserID,
-		"bot_username":         a.BotUsername,
-		"bot_access_token":     a.BotAccessToken,
-		"granted_permissions":  a.GrantedPermissions.toStringArray(),
-		"granted_locations":    a.GrantedLocations.toStringArray(),
-	}
+	b, _ := json.Marshal(a)
```
levb, line 48: There's no reason for these functions then, is there? At least not ToMap - it can be generic in config.store

larkox: If we want to keep the ConfigMapper for future use or in case we change anything, we can keep them (I do not foresee any case we will need it, but I haven't thought much about it). If not, we can let the mapping be done directly by Store function.

larkox:
@levb @mickmister This seems to be working with the changes in redux and webapp linked.
- If the location is not among the granted permissions it is filtered out.
  - A debug message is logged, but since this happen on every getbinding (so on each channel navigation) may flood the logs. Wonder if we should log it or not.
- The app is stored in the config. Whenever a new app is added to the config, the OnConfigurationChange is triggered.

server/apps/impl/bindings.go
```diff
@@ -70,21 +66,26 @@ func (s *service) scanAppBindings(app *apps.App, bindings []*apps.Binding, locPr
 			}
 		}
 		if !allowed {
-			return nil, errors.Errorf("location %s is not granted to app %s", fql, app.Manifest.AppID)
+			// TODO Log this somehow to the app?
+			s.Mattermost.Log.Debug(fmt.Sprintf("location %s is not granted to app %s", fql, app.Manifest.AppID))
```
mickmister, line 70: I've been thinking about this, how can we communicate an error message to the app? But who will be reading the error, likely the MM admin? So having it in the MM logs may be the right place.

server/apps/impl/bindings.go
mickmister:
```suggestion
				// We do not add bindings without any valid sub-bindings
```

mickmister (approved):
LGTM :+1:

server/apps/impl/bindings.go
levb: 2/5 We should hook in the `Logger` soon enough and send the logs to the App's "admin" channel.

larkox: The discussion here is "who can fix things". Certain errors may make sense to log to the Mattermost instance (app admin, sysadmin, or whoever we want in the Mattermost instance). But for locations not granted, there is no actionable thing to do from the Mattermost side. The App provider is the one that should ask for more permissions or stop trying to register that location.

Also, this particular log is quite spammy whatever we do (as soon as one location is misconfigured, any time any user navigates to a channel, it will create a new log message).

PR #35, opened 2020-11-11
Interactivity payload docs

mickmister:
[Rendered docs](https://github.com/mattermost/mattermost-plugin-apps/blob/interactivity-docs/docs/interactivity.md)

#### Summary

This PR is not meant to be merged until we have solidified the payload structures between the proxy plugin and the App servers. Any suggested changes can be discussed in the comments. Once a given change is merged into master, this PR can be updated to reflect that.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-30440

### DIFF
--- a/docs/interactivity.md
+++ b/docs/interactivity.md
@@ -746,6 +746,7 @@ This page shows the payloads for browser request/responses, and App server reque
     "context": {
         "app_id": "hello",
         "location": "send",
+        "post_id": "jgbzgehjqirkb8mn38axjktufw",
         "team_id": "qe5ken7k9f8rdp5bqnfhhs5nzy",
         "channel_id": "zanqhwfdtjfi8yqyapd5qh6udh"
     },
--- a/docs/interactivity.md
+++ b/docs/interactivity.md
@@ -6,7 +6,7 @@ This page shows the payloads for browser request/responses, and App server reque
 
 <details><summary>Request from Browser</summary>
 
-`GET` http://localhost:8065/plugins/com.mattermost.apps/api/v1/bindings?user_id=f88nesmr7ifhicr8wwf94oxiwa&channel_id=b77x9mu8xindzxkgem8guadsra&scope=webapp
+`GET` http://localhost:8065/plugins/com.mattermost.apps/api/v1/bindings?user_id=mum5qskypidf3x3enkindgajrh&channel_id=zanqhwfdtjfi8yqyapd5qh6udh&scope=webapp
 </details>
 
 <details><summary>Response to Browser</summary>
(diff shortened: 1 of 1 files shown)


mickmister:
I have a modal implementation mostly ready for use. Will clean up and get it up in a PR

docs/interactivity.md
```diff
@@ -0,0 +1,483 @@
...
+            "type": "",
+            "props": {
+                "disable_group_highlight": true
+            },
+            "hashtags": "",
+            "pending_post_id": "",
+            "reply_count": 0
+        }
+    },
+    "expand": {
```
levb, line 467: remove `expand`

levb (changes requested):

docs/interactivity.md
```diff
@@ -0,0 +1,483 @@
...
```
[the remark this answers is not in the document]
mickmister: Everything in this doc PR is live data :)

docs/interactivity.md
```diff
@@ -0,0 +1,483 @@
...
+{"markdown": "Successfully sent survey"}
+```
+
+</details>
+
+<details><summary>Request to App server</summary>
+
+`POST` http://localhost:8065/plugins/com.mattermost.apps/hello/send
+
+Notice `type == ""`, or is missing. This means it is a form submission.
```
mickmister, line 429: @levb What are your thoughts on supporting both `type == ""` and `type == "submit"`. I like being able to see `submit` explicitly in the payload. I don't know what the tradeoffs are. We don't want to complicate things on the App side.

docs/interactivity.md
```diff
@@ -0,0 +1,483 @@
...
+        ]
+    }
+]
+```
+
+</details>
+
+<details><summary>Request to App server</summary>
+
+`GET` http://localhost:8065/plugins/com.mattermost.apps/hello/bindings?acting_user_id=f88nesmr7ifhicr8wwf94oxiwa&channel_id=b77x9mu8xindzxkgem8guadsra
```
mickmister, line 116: Currently the browser is providing URL params `user_id=f88nesmr7ifhicr8wwf94oxiwa&channel_id=b77x9mu8xindzxkgem8guadsra&scope=webapp`. The `user_id` is obviously unnecessary.

The App server is receiving params `acting_user_id=f88nesmr7ifhicr8wwf94oxiwa&channel_id=b77x9mu8xindzxkgem8guadsra`. Should we be sending more information? Possibly `team_id`, which could come straight from the browser, though there should be some vetting of these values on the proxy plugin side before sending to the App server.

The proxy plugin is expecting these values from the browser:

https://github.com/mattermost/mattermost-plugin-apps/blob/d2526bdbaef440095fbf5d7d0ee355a882b04e0f/server/http/restapi/bindings.go#L12-L18

larkox:
Maybe "early optimization" but are we considering the docs as we do for other repos like Zoom? (i.e. docs will become a gitbook page)

docs/interactivity.md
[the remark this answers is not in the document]
levb: Yep, it is a mistake, we should strip it on the way out.

docs/interactivity.md
levb: Agree, let's make "ok" and "submit" explicit, and default to them on ""

docs/interactivity.md
```diff
@@ -0,0 +1,457 @@
...
+</details>
+
+<details><summary>Request to App server</summary>
+
+`POST` http://localhost:8065/plugins/com.mattermost.apps/hello/send
+
+```json
+{
+    "url": "http://localhost:8065/plugins/com.mattermost.apps/hello/send",
+    "type": "form", // Notice type == "form". This means it is a request to fetch a form's definition.
```
mickmister, line 282: @larkox made a point [here](https://community-daily.mattermost.com/core/pl/9om7ewtszjrhijaj5itkrnjmfa) that this is forcing a convention onto the App developer. We should instead let the developer decide how they should structure their app, and not force them to use the same endpoint to fetch and submit a form.

docs/interactivity.md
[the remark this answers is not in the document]
larkox: There are a few small cases where the expanded app might be used.
Simplest one, install. On install we can get the OAuthClientID, Secret and BotID from there.
We could say that we could send all those with all calls, so the app does not have to store that information. 0/5 on that.

larkox (approved):
Awesome work Michael. Thanks for keeping this documented!

levb (approved):

PR #37, opened 2020-11-18
Add tests to Store package

larkox:
#### Summary
Add unit tests to Store package

#### Ticket Link
None

### DIFF
--- a/server/apps/store/subscriptions_test.go
+++ b/server/apps/store/subscriptions_test.go
@@ -1,0 +1,353 @@
+package store
+
+import (
+	"encoding/json"
+	"testing"
+
+	pluginapi "github.com/mattermost/mattermost-plugin-api"
+	"github.com/mattermost/mattermost-plugin-apps/server/apps"
+	"github.com/mattermost/mattermost-plugin-apps/server/configurator"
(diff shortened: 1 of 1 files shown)


levb (approved):

server/apps/store/subscriptions_test.go
```diff
@@ -0,0 +1,353 @@
...
+			"sub_channel_created_team-id",
+		},
+		string(apps.SubjectPostCreated): {
+			apps.SubjectPostCreated,
+			"team-id",
+			"channel-id",
+			"sub_post_created_channel-id",
+		},
+	} {
+		t.Run(name, func(t *testing.T) {
```
mickmister, line 348: I think you'll have to do this as well if you make the other change

```suggestion
		t.Run(string(name), func(t *testing.T) {
```

mickmister (approved):
Very nice to have tests added. Great job @larkox :+1:

No issues found but I left a comment for discussion on some test setup

mickmister:
I was wondering if we could use `gomock` instead of `plugintest.API`, but I noticed that we are using `pluginapi.Client`, which expects to receive the `plugin.API` struct.

Feel free to merge :+1:

PR #40, opened 2020-11-23
Updates for commands

mickmister:
#### Summary

This PR contains updates to support bindings for multiple apps. `AppID` is applied to all bindings and nested bindings. This is subject to change.

### DIFF
--- a/server/apps/impl/proxy.go
+++ b/server/apps/impl/proxy.go
@@ -5,9 +5,44 @@ package impl
 
 import (
 	"github.com/mattermost/mattermost-plugin-apps/server/apps"
+	"github.com/pkg/errors"
 )
 
+func (s *service) filterContext(c *apps.Call) error {
+	checkTeam := true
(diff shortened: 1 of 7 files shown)


larkox (approved):
LGTM. Great work!

server/apps/impl/proxy.go
```diff
@@ -5,9 +5,44 @@ package impl
 
 import (
 	"github.com/mattermost/mattermost-plugin-apps/server/apps"
+	"github.com/pkg/errors"
 )
 
+func (s *service) filterContext(c *apps.Call) error {
+	checkTeam := true
+	if c.Context.ChannelID != "" {
```
mickmister, line 13: We should also be checking for the root post id here, and using that to source the channel id if it is present.

levb (approved):
We should bump up implementing `Context.Mentioned` in priority... and consolidate all the `@` and `~` checks in `Expand`.

Otherwise LGTM.

PR #41, opened 2020-11-30
Add aws client caching support

iomodo:
#### Summary
This PR will add AWS client caching, also support of the Mattermost cloud. User can call 
`/apps experiment --app "appname" --release "http://github.com/....zip"` 
to install app in the Mattermost cloud. 

User should pass `keyID` and `secret` to install app in the 3rd party environment.

### DIFF
--- a/server/aws/client_cache.go
+++ b/server/aws/client_cache.go
@@ -1,0 +1,68 @@
+// Copyright (c) 2019-present Mattermost, Inc. All Rights Reserved.
+// See License for license information.
+
+package aws
+
+import (
+	sdk "github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/aws/credentials"
+
+	"github.com/pkg/errors"
+)
+
+// DefaultRegion describes default region in aws
+const DefaultRegion = "us-east-2"
+
+type clientCache struct {
+	store CredentialStoreService
+	cache map[string]*Client // appID - Client
+}
+
+func newClientCache(store CredentialStoreService) (*clientCache, error) {
+	cache := make(map[string]*Client)
+	list, err := store.GetAppList()
+	if err != nil {
+		return nil, errors.Wrap(err, "can't get credentials")
+	}
+	cc := &clientCache{
+		store: store,
+		cache: cache,
+	}
+	for _, appID := range list {
+		if _, err := cc.create(appID); err != nil {
+			return nil, errors.Wrapf(err, "can't create client for app %s", appID)
+		}
+	}
+	return cc, nil
+}
+
+func (c *clientCache) create(appID string) (*Client, error) {
+	keyID, secret, err := c.store.GetAWSCredential(appID)
+	if err != nil {
+		return nil, errors.Wrap(err, "can't get app from the store")
+	}
+	var config *sdk.Config
+	if keyID == "" || secret == "" {
+		config = &sdk.Config{
+			Region:      sdk.String(DefaultRegion),
+			Credentials: credentials.NewEnvCredentials(), // Read Mattermost cloud credentials from the environment variables
+		}
+	} else {
+		config = &sdk.Config{
+			Region:      sdk.String(DefaultRegion),
+			Credentials: credentials.NewStaticCredentials(keyID, secret, ""),
+		}
+	}
+	client := NewAWSClientWithConfig(config)
+	c.cache[appID] = client
+
+	return client, nil
+}
+
+func (c *clientCache) get(appID string) (*Client, error) {
+	client, ok := c.cache[appID]
+	if !ok {
+		return nil, errors.New("no client")
+	}
+	return client, nil
+}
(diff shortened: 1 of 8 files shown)


server/command/install.go
```diff
@@ -91,23 +87,14 @@ func (s *service) executeExperimentalInstall(params *params) (*model.CommandResp
...
-		return normalOut(params, nil, err)
+	if !s.apps.AWSProxy.IsAppInstalled(app) {
+		if err := s.apps.AWSProxy.InstallApp(app, releaseURL, id, secret); err != nil {
+			return normalOut(params, nil, err)
+		}
 	}
+
 	// An example of the function Invoke
-	// res, err := client.InvokeFunction("my_app_my_func", "blabla")
+	// res, err := s.apps.AWSProxy.InvokeFunction("my_app", "my_func", "blabla")
```
larkox, line 97: nit: Maybe change `blabla` for something like `my values`?

server/aws/credential_store.go
```diff
@@ -0,0 +1,138 @@
...
+	if len(value) == 0 {
+		ok, err := s.Mattermost.KV.Set(appsKey, &value)
+		if err != nil {
+			return errors.Wrapf(err, "can't set app list in KV store")
+		}
+		if !ok {
+			return errors.Errorf("can't set apps list in KV store")
+		}
+	}
+	return nil
```
larkox, line 52: Is this process needed? We are getting a value from the KVStore, which if it does not exist, will return a empty list of string, and we initialize it with a empty list of strings. Isn't it a bit redundant?

server/aws/credential_store.go
```diff
@@ -0,0 +1,138 @@
...
+	if !ok {
+		return errors.Errorf("can't set app %s credential from KV store", appID)
+	}
+
+	appsKey := getAppsKey()
+	var appsList []string
+	if err = s.Mattermost.KV.Get(appsKey, &appsList); err != nil {
+		return errors.Wrap(err, "can't get apps list from KV store")
+	}
+	appsList = append(appsList, appID)
```
larkox, line 83: Should we check if the app exist already in the list? Or is it handled somewhere else?

server/aws/credential_store.go
```diff
@@ -0,0 +1,138 @@
...
+		return errors.Wrapf(err, "can't set app %s credential from KV store", appID)
+	}
+	if !ok {
+		return errors.Errorf("can't set app %s credential from KV store", appID)
+	}
+
+	appsKey := getAppsKey()
+	var appsList []string
+	if err = s.Mattermost.KV.Get(appsKey, &appsList); err != nil {
+		return errors.Wrap(err, "can't get apps list from KV store")
```
larkox, line 81: Is there any big problem if the credentials are stored but the app is not added to the list? If there is, we should at least try here (and other early returns) to delete the credentials.

server/aws/credential_store.go
```diff
@@ -0,0 +1,138 @@
...
+	}
+
+	ok, err := s.Mattermost.KV.Set(appsKey, &appsList)
+	if err != nil {
+		return errors.Wrap(err, "can't set apps list in KV store")
+	}
+	if !ok {
+		return errors.Errorf("can't set apps list in KV store")
+	}
+	return nil
```
larkox, line 120: I wonder if we should try to be atomic with these updates on the list of apps. Set with options should allow atomic functionality.

server/aws/client_cache.go
```diff
@@ -0,0 +1,68 @@
...
+	if err != nil {
+		return nil, errors.Wrap(err, "can't get credentials")
+	}
+	cc := &clientCache{
+		store: store,
+		cache: cache,
+	}
+	for _, appID := range list {
+		if _, err := cc.create(appID); err != nil {
+			return nil, errors.Wrapf(err, "can't create client for app %s", appID)
```
larkox, line 33: Should this be an return or a log + continue?

server/aws/client_cache.go
```diff
@@ -0,0 +1,68 @@
...
+	list, err := store.GetAppList()
+	if err != nil {
+		return nil, errors.Wrap(err, "can't get credentials")
+	}
+	cc := &clientCache{
+		store: store,
+		cache: cache,
+	}
+	for _, appID := range list {
+		if _, err := cc.create(appID); err != nil {
```
larkox, line 32: Is it better to create all the clients on init, or it is better to create the clients on demand? (i.e. in the get function if the client does not exist)

server/aws/client_cache.go
```diff
@@ -0,0 +1,68 @@
...
+
+// DefaultRegion describes default region in aws
+const DefaultRegion = "us-east-2"
+
+type clientCache struct {
+	store CredentialStoreService
+	cache map[string]*Client // appID - Client
+}
+
+func newClientCache(store CredentialStoreService) (*clientCache, error) {
```
larkox, line 21: How does the cache gets updated / invalidated? For example, when an app gets added/deleted. Not sure if the apps will change the credentials at any moment.

larkox (changes requested):
Lots of questions, and a few nits here and there.

levb (changes requested):
Can we start with a much simpler implementation of having a single set of AWS credentials stored in the (MMinstance/plugin-apps) config?

2/5 Using KV store we'll run into the usual performance/can't cache issues.

PR #43, opened 2020-12-02
KV, Upstream

levb:
- adds `Upstream` interface so that we can connect to upstream Apps via HTTP, AWS Lambda `Invoke`, directly with Go built-in apps, and other protocols in the future.
- adds a basic KV store REST API.
- `GetBindings` is now a `Call` to the App, to simplify the protocol.
- Temporarily uses user session token for Apps, instead of removed OAuth2 support, https://mattermost.atlassian.net/browse/MM-31117 filed.
- `Context[.Expanded].Config` is gone, replaced with `Context.MattermostSiteURL` which is always set.

TODO: 
- Partially tested, embedded forms don't work
- ASW Lambda upstream is stubbed out
- Fixes/updates to `mattermost-redux` may be needed

Separate PRs to follow:
- https://mattermost.atlassian.net/browse/MM-31117: OAuth as a service for user tokens
- https://mattermost.atlassian.net/browse/MM-31118: support storing 3rd party user account info and mappings
- https://mattermost.atlassian.net/browse/MM-29358: AWS Lambda example app


iomodo (commented):
Several questions:
1. Why are we saving aws session in the config object?
2. Why do we have aws session keys in the config? For Mattermost Cloud they should be in the system variables, for third party aws cloud credentials should be passed on installation. When are these credentials in the config?

server/api/impl/aws/aws.go
```diff
@@ -0,0 +1,72 @@
...
+				l = l.WithField("params", string(paramBytes))
+			}
+			l.Debug(buffer.String())
+		}
+	})
+	return sess
+}
+
+func (s *Service) lambda() *lambda.Lambda {
+	return lambda.New(s.newSession(), aws.NewConfig().WithLogLevel(aws.LogDebugWithRequestErrors))
```
iomodo, line 67: why do we create different sessions each time we call lambda? or iam?

levb:
> 2. Why do we have aws session keys in the config? For Mattermost Cloud they should be in the system variables, for third party aws cloud credentials should be passed on installation. When are these credentials in the config?

A temp. placeholder. Filed https://mattermost.atlassian.net/browse/MM-31119 - we should define and implement it.

> 1. Why are we saving aws session in the config object?

The aws session is "just data", there is no OS/network resources associated with it, and it depends only on the stored config data. It was easiest to add it there (and follow the Refresh cycle) rather than create a separate mutex. This would change if we don't store the access token in the config.

server/api/impl/aws/aws.go
levb: again, it's just a couple structs... seems meaningless to cache them all, but we could, along with/instead of AWS session.

server/api/impl/proxy/auth.gogo
```diff
@@ -1,4 +1,4 @@
-package helloapp
+package http_hello
```
levb, line 1: This file will be gone once https://mattermost.atlassian.net/browse/MM-31117 is implemented, just keeping the code around

server/api/impl/proxy/bindings.go
```diff
@@ -31,41 +31,42 @@ func mergeBindings(bb1, bb2 []*apps.Binding) []*apps.Binding {
...
-		bb, err := s.Client.GetBindings(&appCC)
+		appCC.AppID = appID
+
+		up, err := p.upstreamForApp(app)
 		if err != nil {
-			return nil, errors.Wrapf(err, "failed to get bindings for %s", app.Manifest.AppID)
+			return nil, errors.Wrapf(err, "failed to make upstream for %s", appID)
+		}
+
+		// TODO PERF: Add caching
```
levb, line 48: Filed https://mattermost.atlassian.net/browse/MM-31123, https://mattermost.atlassian.net/browse/MM-31124

server/api/impl/proxy/expand.go
```diff
@@ -1,120 +1,147 @@
...
 	return e
 }
 
 // Expand collects the data that is requested in the expand argument, and is not
 // yet collected. It then returns a new Context, filtered down to what is
 // specified in expand.
-func (e *expander) Expand(expand *apps.Expand) (*apps.Context, error) {
+func (e *expander) Expand(expand *api.Expand) (*api.Context, error) {
 	clone := *e.Context
+
```
levb, line 38: These are always expanded, even if `Call.Expand` is empty

server/api/impl/proxy/manifest.go
```diff
@@ -0,0 +1,49 @@
+// Copyright (c) 2020-present Mattermost, Inc. All Rights Reserved.
```
levb, line 1: These are convenience methods.

As far as proxy is concerned, manifest is just text, and comes (usually) from the Marketplace, with the `InstallApp` API call.

server/api/impl/upstream/upawslambda/upstream.go
```diff
@@ -0,0 +1,31 @@
+// Copyright (c) 2020-present Mattermost, Inc. All Rights Reserved.
+// See License for license information.
+
+package upawslambda
```
levb, line 4: TODO: plug in the `aws` package here

server/api/impl/upstream/uphttp/upstream.go
```diff
@@ -0,0 +1,110 @@
...
+
+	cr := api.CallResponse{}
+	err = json.NewDecoder(resp.Body).Decode(&cr)
+	if err != nil {
+		return api.NewErrorCallResponse(err)
+	}
+	return &cr
+}
+
+func (u *Upstream) GetBindings(call *api.Call) ([]*api.Binding, error) {
```
levb, line 53: I left this in the interface because the go `json` peculiarities will not easily allow to decode specific types of `Data`, like []*Binding in this case.

The alternative was to expose the res.Body stream to the caller, and let the caller fo her own json.Decode, but I didn't like it.

server/command/debug.go
```diff
@@ -4,55 +4,95 @@
...
-		Context: &apps.Context{
-			AppID:        helloapp.AppID,
-			ActingUserID: params.commandArgs.UserId,
-			ChannelID:    params.commandArgs.ChannelId,
-			TeamID:       params.commandArgs.TeamId,
-			UserID:       params.commandArgs.UserId,
-		},
-	})
+func (s *service) executeDebugEmbeddedForm(params *params) (*model.CommandResponse, error) {
+	cc := api.NewCommandContext(params.commandArgs)
```
levb, line 35: @larkox I totally bastardized the Embedded functionality to make this PR work, will appreciate a quick discussion on how we can restore the functionality, it's pretty essential to `Hello` app :)

server/api/impl/upstream/uphttp/upstream.go
```diff
@@ -0,0 +1,110 @@
...
+	appSecret string
+}
+
+func NewUpstream(app *api.App) *Upstream {
+	return &Upstream{app.Manifest.HTTPRootURL, app.Secret}
+}
+
+func (u *Upstream) InvokeNotification(n *api.Notification) error {
+	// TODO
+	resp, err := u.post("", u.rootURL+"/notify/"+string(n.Subject), n)
```
levb, line 30: TODO: Need to change notification path to a relative URL, provided by the caller in a `Subscription`; maybe re-consider using the `Call` format for notifications in general, just a different `Call.Type`? Filed https://mattermost.atlassian.net/browse/MM-31125

server/api/impl/aws/install_app.go
```diff
@@ -145,9 +158,41 @@ func (c *Client) installApp(appName string, functions []functionInstallData) err
...
 			return errors.Wrapf(err, "can't install function for %s", appName)
 		}
 	}
 	return nil
 }
+
+// CreateFunction method creates lambda function
+func (s *Service) createFunction(zipFile io.Reader, function, handler, runtime, resource string) error {
+	if zipFile == nil || function == "" || handler == "" || resource == "" || runtime == "" {
+		return errors.Errorf("you must supply a zip file, function name, handler, ARN and runtime - %s %s %s %s %s", zipFile, function, handler, resource, runtime)
```
mickmister, line 171: Will the `io.Reader` potentially take up a very large log line here?

server/api/impl/upstream/uphttp/upstream.go
```diff
@@ -0,0 +1,110 @@
...
+	if err != nil {
+		return err
+	}
+	defer resp.Body.Close()
+	return nil
+}
+
+func (u *Upstream) InvokeCall(call *api.Call) *api.CallResponse {
+	resp, err := u.post(call.Context.ActingUserID, u.rootURL+call.URL, call)
```
mickmister, line 39: Maybe check if `call.URL` begins with a slash? If we want to do that, we can compute this full call URL in a helper function.

server/api/impl/proxy/proxy.go
```diff
@@ -0,0 +1,128 @@
...
+		return api.NewErrorCallResponse(err)
+	}
+	clone := *c
+	clone.Context = cc
+
+	return up.InvokeCall(&clone)
+}
+
+func (p *Proxy) Notify(cc *api.Context, subj api.Subject) error {
+	app, err := p.store.LoadApp(cc.AppID)
```
mickmister, line 55: Is `cc.AppID` ever defined for calls to this `Notify` function? It seems the `api.Context` is generated without an `AppID`, such as the call to [`api.NewUserContext(user)`](https://github.com/mattermost/mattermost-plugin-apps/blob/e887a213b6dab2fc904762b3e2a4d694ecafa1df/server/plugin.go#L106), which has no way of indicating an `AppID`.

mickmister (changes requested):
Mainly just some questions. I believe there's one functional issue found on my comment regarding calls to `proxy.Notify`. The issue may be pre-existing.

server/api/constants.go
```diff
@@ -0,0 +1,45 @@
...
+	InteractiveDialogPath = "/dialog"
+
+	// Top-level path(s) for HTTP example apps.
+	HelloHTTPPath = "/example/hello"
+
+	// Top-level path for the REST APIs exposed by the plugin itself.
+	APIPath = "/api/v1"
+
+	// Top-level path for the Apps namespaces, followed by /AppID/subpath.
+	AppsPath = "/apps"
```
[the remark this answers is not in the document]
levb: I toyed with the idea, but for now - let's stick with the `Call` API. The Call URLs should all now be "relative", but I don't think the user-agent currently composes them, so hopefully a non-change there.

Going forward, we may consider having /apps/id space transparently proxied for Calls and other HTTP requests.

server/api/impl/admin/install.go
```diff
@@ -28,70 +28,69 @@ func (s *service) InstallApp(cc *apps.Context, sessionToken apps.SessionToken, i
...
-			_ = s.Mattermost.Post.DM(app.BotUserID, actingUserID, &model.Post{
+			_ = adm.mm.Post.DM(app.BotUserID, actingUserID, &model.Post{
 				Message: fmt.Sprintf("Using existing OAuth2 App `%s`.", oauthApp.Id),
 			})
 
 			return oauthApp, nil
 		}
 	}
 
+	oauth2CallbackURL := adm.conf.GetConfig().PluginURL + api.AppsPath + "/" + string(manifest.AppID) + api.PathOAuth2Complete
```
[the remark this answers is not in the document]
levb: This is not even correct ATM :( will fix.

server/api/impl/upstream/uphttp/upstream.go
levb: 1/5 call "URLs" (paths) must start with a `/` semantically; if they don't - the resulting URL will be invalid.

0/5 consider renaming `Call.URL` to `Call.Path`

docs/interactivity.md
```diff
@@ -6,217 +6,182 @@ This page shows the payloads for browser request/responses, and App server reque
...
 [
     {
-        "app_id": "hello",
         "location": "/channel_header",
         "bindings": [
             {
-                "app_id": "hello",
+                "app_id": "builtin",
                 "location": "send",
-                "presentation": "modal",
```
mickmister, line 23: Why is the `presentation` property removed here? That is how a modal is opened with a binding, as used in https://github.com/mattermost/mattermost-plugin-apps/pull/45 . Can we have doc updates in their own PR so it's more obvious for discussion?

mickmister: Granted that I didn't intend the docs PR to be merged yet, as it does not reflect the code that's in master. Having a long running PR for docs is probably not the way to go, but that is what I originally intended.

So I propose we have a docs PR per significant change, or have the docs change within the code change PR as you have done here.

mickmister (approved):

iomodo (approved):

PR #46, opened 2021-01-11
Add apps synchronization

iomodo:
#### Summary
This PR syncs apps on plugin startup. 

Have several questions to discuss:
1. Should we move the AWS app provisioning in `admin.ProvisionApp`?
2. Should we support force installs, which is installing a new app automatically, by changing a mappings file in the S3?
3. Generating an AWS function name: `appID-version-function` if it is less than 64 bytes, otherwise `appID-version-hash`. Notice that appID length is limited to 32 bytes, app version length is limited to 10 bytes to accommodate v00.00.000 formatting.

P.S. Yet there is no implementation of the app upgrade/downgrade/delete functionality

#### Ticket Link

  Fixes https://mattermost.atlassian.net/browse/MM-31780


server/api/impl/admin/synchronize.go
```diff
@@ -0,0 +1,116 @@
...
+	}
+
+	apps, _, err := adm.ListApps()
+	if err != nil {
+		return errors.Wrap(err, "can't get apps list")
+	}
+
+	for _, app := range apps {
+		mapping, ok := mappings.Apps[string(app.Manifest.AppID)]
+		if !ok {
```
larkox, line 83: Maybe I don't understand the logic here, but if the app exist in the instance but not on AWS, shouldn't we provision the app instead of removing it from the instance?

larkox (approved):
LGTM, just one question.

server/api/impl/admin/synchronize.go
catalintomai: I believe this is because provisioning always precedes install and if for some reason an app provisioning fails it is prob "cleaner" to remove it from the instance. The provisioning functionality is only here temporarily, we'll follow/piggyback the cloud provisioning process in the future.

catalintomai (approved):

server/api/impl/admin/synchronize.go
iomodo: Correct, but it's not only for "cleanness". There might be a possibility that we want to remove an app from the marketplace and remove it on every installation. Or for example if customer did not pay for the app subscription. There might be other reasons too.

levb (changes requested):
@iomodo and I discussed a potential simplification offline

server/api/impl/admin/synchronize.go
```diff
@@ -0,0 +1,186 @@
...
+		if _, ok := listedAppsMap[appID]; !ok {
+			if err := adm.AddApp(manifest); err != nil {
+				return errors.Wrap(err, "can't add new app as listed")
+			}
+		}
+	}
+
+	listedAppsUpgraded := adm.store.ListApps()
+
+	// call onInstanceStartup. App migration happens here
```
catalintomai, line 106: Should we call it onInstanceStartup/OnStartup or OnActivate (to keep in sync with the plugin naming)?

server/api/impl/admin/synchronize.go
```diff
@@ -0,0 +1,186 @@
...
+			}
+		}
+	}
+
+	return nil
+}
+
+func (adm *Admin) DeleteApp(app *api.App) error {
+	// Call delete the function of the app
+	if err := adm.call(app, app.Manifest.Delete, nil); err != nil {
```
catalintomai, line 125: Should we use the On* naming - like OnDelete in this case?

server/api/impl/admin/synchronize.go
```diff
@@ -0,0 +1,186 @@
...
+		return errors.Wrap(err, "delete failed")
```
catalintomai, line 126: "delete failed" - add more detail (app Id, error from cloud app)?

server/api/impl/aws/install_app.go
```diff
@@ -6,6 +6,8 @@ package aws
 import (
```
catalintomai, line 6: should we rename the file to `provision_app.go` then?

server/plugin.go
```diff
@@ -86,6 +86,13 @@ func (p *Plugin) OnActivate() error {
 	if err != nil {
 		return errors.Wrap(err, "failed to initialize own command handling")
 	}
+
+	if err := p.api.Admin.SynchronizeApps(); err != nil {
+		mm.Log.Error("Can't synchronize", "err", err.Error())
+		// TODO should we fail or continue?
```
catalintomai, line 92: I think we should continue, but we need to figure out the monitoring piece that let's us know quickly which apps failed to synchronize (without sifting through the log of error and info entries).

server/api/impl/admin/admin.go
```diff
@@ -6,17 +6,20 @@ package admin
...
-	mm    *pluginapi.Client
-	conf  api.Configurator
-	store api.Store
-	proxy api.Proxy
+	mm         *pluginapi.Client
+	conf       api.Configurator
+	store      api.Store
+	proxy      api.Proxy
+	awsClient  *aws.Client
+	adminToken api.SessionToken // TODO populate admin token
```
catalintomai, line 18: Is this the token referred to in https://mattermost.atlassian.net/browse/MM-28542 - used for upgrade ops only? If yes, should we call it adminToken or use some other name?

server/api/impl/admin/synchronize.go
```diff
@@ -0,0 +1,186 @@
...
+
+	adm.mm.Log.Info("Deleted the app", "app_id", app.Manifest.AppID)
+
+	return nil
+}
+
+func (adm *Admin) AddApp(manifest *api.Manifest) error {
+	newApp := &api.App{}
+	newApp.Manifest = manifest
+	newApp.Status = api.AppStatusListed
```
catalintomai, line 154: 1/5 suggest using Registered instead of Listed (like in "app is registered for use")

server/api/impl/admin/synchronize.go
iomodo: I think name was intentionally changed/introduced by Lev not to mess with the plugin's `onAvtivate`. I'm 0/5 on naming. @levb thoughts?

server/api/impl/admin/admin.go
iomodo: Yes, that's the one. Agree, `adminToken` is kinda ambiguous name. In the similar sense `Admin` struct name is ambiguous too. This struct administers and manages the app installation, upgrade, etc. hence the name. It has nothing to do with the system admin and that's the misleading part of the name.
how about `administrationToken` or `proxyPluginToken` or `mainToken`. I'm 0/5

server/api/impl/admin/synchronize.go
catalintomai: Maybe then smth. like OnAppActivate to establish some kind of the parity/continuity? One goal is to reduce the amount of re-thinking for a plugin developer that wants to develop apps.

server/api/app.go
```diff
@@ -69,6 +131,7 @@ var DefaultBindingsCall = &Call{
 
 type App struct {
 	Manifest *Manifest `json:"manifest"`
```
levb, line 133: 4/5 We should "decouple" `Manifest` and `App`. Manifest is "listing data", static, same for all instances, immutable, comes from the marketplace via the env variable, and is only cached in memory. `App` is the local app's configuration, and is stored in the config. No need to store the manifest in the config, in fact we specifically want to avoid this (size limitations).

server/api/impl/admin/admin.go
levb: 3/5 Admin (access/session) token can not be a part of this struct - `Admin` is a "service, it lives for the entire duration of the proxy, across all apps and requests. The access token should be a request attribute.

server/api/app.go
```diff
@@ -18,9 +18,67 @@ func (at AppType) IsValid() bool {
...
+func (at AssetType) IsValid() bool {
+	return at == s3Asset ||
+		at == httpAsset
+}
+
+// AppStatus describes status of the app
+type AppStatus string
+
+const (
+	AppStatusRegistered AppStatus = "registered"
```
levb, line 54: 1/5 `s/registered/listed/` or even, `known`?

server/api/app.go
```diff
@@ -41,12 +99,16 @@ type Manifest struct {
...
+	OnInstall *Call `json:"install,omitempty"`
+	OnDelete  *Call `json:"delete,omitempty"`
+	OnStartup *Call `json:"on_startup,omitempty"`
 
 	// By default invoke "/bindings".
 	Bindings *Call `json:"bindings,omitempty"`
 
 	// Deployment manifest for hostable apps will include path->invoke mappings
+	Functions []Function
+	Assets    []Asset
```
levb, line 111: Like I said in the other comment, we don't really need each (static) asset to be listed individually, do we? They are all just files in a folder.

server/api/impl/admin/synchronize.go
levb: App "Takedown" is a very important use case. It should work reliably, and fast - without completely destroying the customer experience.

server/api/impl/admin/synchronize.go
```diff
@@ -0,0 +1,186 @@
...
+}
+
+// ToJSON serializes a list of app mappings to json.
+func (m *Mappings) ToJSON() []byte {
+	b, _ := json.Marshal(m)
+	return b
+}
+
+// SynchronizeApps synchronizes apps upgrading, downgrading and deleting apps
+func (adm *Admin) SynchronizeApps() error {
```
levb, line 48: 2/5 I still maintain that tracking the previous version, and providing upgrade/downgrade messages from the proxy plugin is premature.

If we do want to track the state and detect these state transitions, we need to deduplicate the App messages in HA, avoid sending 1 message/server.

server/api/impl/admin/synchronize.go
levb: Ditto my previous comment, need to deduplicate for HA (or have the message come from higher-level the Cloud scripts, that start the instance?)

server/api/impl/admin/synchronize.go
```diff
@@ -0,0 +1,186 @@
...
+			if err := adm.call(registeredApp, registeredApp.Manifest.OnStartup, values); err != nil {
+				adm.mm.Log.Error("Can't call onStartup func of the app", "app_id", registeredApp.Manifest.AppID, "err", err.Error())
+			}
+		}
+	}
+
+	return nil
+}
+
+func (adm *Admin) DeleteApp(app *api.App) error {
```
levb, line 123: 2/5 We need to delete the bot account also.

levb (changes requested):

server/api/impl/admin/admin.go
iomodo: If we agree that all changes to apps are made only in the plugin's `onActivate` method, we can have short lived admin token scoped down to the `onActivate` - Create admin token, use for the synchronization and delete.
Another approach could be to create token once, save it in the `Admin` struct(and update it every month or so for security reasons?).
Anyway this effort is the part of the different ticket.

server/api/app.go
```diff
@@ -18,9 +18,67 @@ func (at AppType) IsValid() bool {
...
+// Assets can be saved in S3 with appropriate permissions,
+// or they could be fetched as ordinary http resources.
+type AssetType string
+
+const (
+	s3Asset   AssetType = "s3_asset"
+	httpAsset AssetType = "http_asset"
+)
+
+func (at AssetType) IsValid() bool {
```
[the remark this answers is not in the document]
iomodo: Provisioning of the `/static` folder should be done on provisioning of the app. Provisioner will create the assets and put the data into the mappings file which are the manifests.

server/api/app.go
iomodo: Understood, should this be a the part of another ticket? Or should I include it in this PR?

server/api/app.go
iomodo: It was named `listed`, Catalin suggested to rename it to `registered`. I'm 0/5 on either one.

server/api/app.go
iomodo: We do. otherwise Proxy plugin will not be able to reach them. Another thing is that **app developer** should not do that. This will be done by the provisioner.

server/api/impl/admin/synchronize.go
iomodo: It's just the comment left from before. upgrade/downgrade messages are not sent. Only `onStartup` and `onDelete` - so there is no duplication.

server/api/impl/admin/synchronize.go
iomodo: This actually depends on our architecture. Should the `OnStartup` function be called for every node of an HA environment? For plugins `onActivate` is called for every node. How about `onDelete`?

server/api/impl/admin/synchronize.go
iomodo: Agree, but there is not endpoint for deleting the bots. I'll add a ticket for that

server/api/impl/admin/synchronize.go
levb: For plugins this too is an issue, and we are struggling with it (or just plain lucky that the synchronization issues don't bubble up, like in Jira 3.0 data migration)

3/5 We really want cluster-level lifecycle events for Apps, not instance-level.

server/api/app.go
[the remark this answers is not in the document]
levb: Got it, sorry for the misunderstanding. So, for the instance, each asset must (of course!) must be explicitly declared.

server/api/app.go
levb: 1/5 I'd say in this PR since in practical terms - can we even fit the mappings into the config? Isn't there a limit on the config value size?

server/api/app.go
levb: :) let's leave it as then, at least for the time being.

server/api/impl/admin/synchronize.go
```diff
@@ -0,0 +1,225 @@
...
+	registeredAppsMap := map[apps.AppID]*apps.App{}
+	updatedAppVersionsMap := map[apps.AppID]string{}
+	// Update apps
+	for _, registeredApp := range registeredApps {
+		registeredAppsMap[registeredApp.Manifest.AppID] = registeredApp
+		manifest, err := adm.store.LoadManifest(registeredApp.Manifest.AppID)
+		if err != nil {
+			return errors.Wrapf(err, "can't load manifest from store appID = %s", registeredApp.Manifest.AppID)
+		}
+		if err := adm.DeleteApp(registeredApp); err != nil {
```
catalintomai, line 78: Is the DeleteApp same as UninstallApp, and should we provide a path for the admin to directly (REST API/command) to uninstall the app?

server/api/app.go
catalintomai: I was 1/5 on it - listed seems to be the already established name. Either way it's fine from my pov.

apps/app.go
```diff
@@ -18,9 +18,52 @@ func (at AppType) IsValid() bool {
...
+	return at == s3Asset ||
+		at == httpAsset
+}
+
+// AppStatus describes status of the app
+type AppStatus string
+
+const (
+	AppStatusRegistered AppStatus = "registered"
+	AppStatusEnabled    AppStatus = "enabled"
```
catalintomai, line 41: When do we set this flag on?

server/api/impl/admin/synchronize.go
```diff
@@ -0,0 +1,225 @@
...
+const oldVersionKey = "update_from_version"
+
+// AppVersions describes versions for all the apps in all installations
+type AppVersions struct {
+	Apps      map[apps.AppID]string            `json:"apps"`
+	Overrides map[string]map[apps.AppID]string `json:"overrides"`
+}
+
+func getAppsForInstallation(installationID string) (map[apps.AppID]string, error) {
+	data, err := ioutil.ReadFile("assets/apps.json")
```
catalintomai, line 26: nit: move file name to a const

server/command/install.go
```diff
@@ -78,7 +78,7 @@ func (s *service) executeInstall(params *params) (*model.CommandResponse, error)
 	}, nil
 }
 
-func (s *service) executeExperimentalInstall(params *params) (*model.CommandResponse, error) {
+func (s *service) executeProvision(params *params) (*model.CommandResponse, error) {
```
catalintomai, line 81: Just to be clear - need to document who is executing this command and at what step in the "deploy new app" process.

server/api/store.go
```diff
@@ -12,4 +12,10 @@ type Store interface {
 	ListApps() []*apps.App
```
catalintomai, line 12: An improvement would be to use the DAL pattern used on the server and have an `AppsStore` wrapper and a `ManifestStore` wrapper and then rename StoreApp/StoreManifest to Save and `LoadApp/LoadManifest` to `Get` and 'ListApps` to `GetAll`

server/api/impl/store/apps.go
```diff
@@ -16,6 +16,7 @@ func (s *Store) ListApps() []*apps.App {
 	}
 	for _, v := range conf.Apps {
```
catalintomai, line 17: Why use `ListApps` for all apps and `LoadApp` for a single one? 1/5 to use `GetListedApps` and `GetListedApp` instead. Also, `StoreApp` is a bit redundant (since called on a store), 1/5 to use `SaveApp` instead (or as I was pointing down below, get rid off all App/Manifest suffixes and have more than one store wrapper).

server/api/impl/admin/synchronize.go
```diff
@@ -0,0 +1,225 @@
...
+	if err := adm.store.DeleteApp(app); err != nil {
+		return errors.Wrapf(err, "can't delete app - %s", app.Manifest.AppID)
+	}
+
+	adm.mm.Log.Info("Deleted the app", "app_id", app.Manifest.AppID)
+
+	return nil
+}
+
+func (adm *Admin) AddApp(manifest *apps.Manifest) error {
```
catalintomai, line 155: Maybe instead have `ListApp' and rename the previous `LoadApp` as `GetListedApp`

server/api/impl/admin/install.go
```diff
@@ -43,14 +43,17 @@ func (adm *Admin) InstallApp(cc *apps.Context, sessionToken apps.SessionToken, i
 		app.OAuth2ClientID = oAuthApp.Id
 		app.OAuth2ClientSecret = oAuthApp.ClientSecret
 		app.OAuth2TrustedApp = in.OAuth2TrustedApp
+		if app.Status == "" || app.Status == apps.AppStatusRegistered {
```
catalintomai, line 46: Line 22 above, should we check for `app` presence?

server/api/impl/admin/synchronize.go
```diff
@@ -0,0 +1,225 @@
...
+			// Note that we are not returning an error here.
+			adm.mm.Log.Error("can't get manifest for", "app", id, "version", version, "err", err)
+			continue
+		}
+		adm.store.StoreManifest(manifest)
+	}
+}
+
+// SynchronizeApps synchronizes apps with the mappings file stored in the env var.
+func (adm *Admin) SynchronizeApps() error {
```
catalintomai, line 58: TODO: Add some unit/integration tests that validate the register/install methods, store, the app statuses, the app config file, maybe mock manifest, AWS, the AWS provisioning steps, etc.

server/api/impl/admin/synchronize.go
```diff
@@ -0,0 +1,225 @@
...
+	if err := adm.store.StoreApp(newApp); err != nil {
+		return errors.Wrapf(err, "can't store app - %s", manifest.AppID)
+	}
+	adm.mm.Log.Info("App is registered", "app_id", manifest.AppID)
+	return nil
+}
+
+func (adm *Admin) callOnce(f func() error) error {
+	// Delete previous job
+	key := "PP_CallOnce_key"
```
levb, line 168: 1. 2/5 I do not find the prefix `PP` to be intention revealing.
2. 0/5 All key names should be const, and (somewhat) consolidated, declared in 1 (or a few) place(s)

server/api/impl/admin/synchronize.go
```diff
@@ -0,0 +1,225 @@
...
+	if err := adm.mm.KV.Delete(key); err != nil {
+		return errors.Wrap(err, "can't delete key")
+	}
+	// Ensure all instances run this
+	time.Sleep(10 * time.Second)
```
levb, line 173: `Sleep` is evil. How is 10s determined? Why is it not 1s or 100s?

server/api/impl/admin/synchronize.go
```diff
@@ -0,0 +1,225 @@
...
+	if err != nil {
+		return errors.Wrapf(err, "can't set key %s to %d", key, value)
+	}
+	if !ok {
+		return errors.Errorf("can't set key %s to %d", key, value)
+	}
+	return nil
+}
+
+func (adm *Admin) call(app *apps.App, call *apps.Call, values map[string]string) error {
```
levb, line 201: 1/5 `call` is not very intention-revealing. 0/5 Perhaps `expandedAdminCall`? Seems a handful to type/read. `adminCall`? Do all `Call`s made by Admin interface go through this function? If not, the name should reflect the intention.

server/api/impl/aws/provision_app.go
```diff
@@ -14,16 +16,13 @@ import (
...
 )
 
 const lambdaFunctionFileNameMaxSize = 64
-
-type function struct {
-	Name    string `json:"name"`
-	Handler string `json:"handler"`
-	Runtime string `json:"runtime"` // runtime can be detected automatically from reading the lambda function
-}
+const appIDLengthLimit = 32
```
levb, line 24: I'm continued to be bothered by these limits. Why they are important?

server/api/impl/aws/provision_app.go
```diff
@@ -201,6 +187,24 @@ func (c *Client) createFunction(zipFile io.Reader, function, handler, runtime, r
 }
 
 // getFunctionName generates function name for a specific app
-func getFunctionName(appName, function string) string {
-	return fmt.Sprintf("%s_%s", appName, function)
+// name can be 64 characters long.
```
levb, line 190: I know I have asked this before but why not use content-based names (functions 64-byte SHAs)

server/plugin.go
```diff
@@ -88,6 +94,11 @@ func (p *Plugin) OnActivate() error {
 	if err != nil {
 		return errors.Wrap(err, "failed to initialize own command handling")
 	}
+
+	if err := p.api.Admin.SynchronizeApps(); err != nil {
```
levb, line 98: 0/5 s/SynchronizeApps/LoadAppsList/?  Synchronize does not appear to be specifically intention-revealing.

server/plugin.go
```diff
@@ -68,12 +69,17 @@ func (p *Plugin) OnActivate() error {
...
+		return errors.Wrapf(err, "failed creating cluster mutex")
+	}
+
 	p.api = &api.Service{
 		Mattermost:   mm,
 		Configurator: conf,
 		Proxy:        proxy,
 		AppServices:  appservices.NewAppServices(mm, conf, store),
-		Admin:        admin.NewAdmin(mm, conf, store, proxy),
+		Admin:        admin.NewAdmin(mm, conf, store, proxy, awsClient, mutex),
```
levb, line 82: nit: 0/5 is there a strong reason to create cluster.Mutex in plugin.go rather than in the admin service itself?

server/command/install.go
levb: I too am a little confused by the long-term intention of this command; is it intended for developers?

server/api/store.go
levb: Whatever pattern we use here, I'd prefer that it is consistent across all stores in `plugin-apps`. `GetAll` seems more intention-revealing than `ListApps`.

server/api/impl/store/store.go
```diff
@@ -13,6 +14,8 @@ const prefixSubs = "sub_"
 type Store struct {
 	mm   *pluginapi.Client
 	conf api.Configurator
+	// Manifests of the registered apps for this installation
+	manifests map[apps.AppID]*apps.Manifest
```
levb, line 18: A little confused here. is `manifests` a cache? Or is `Store` an in-memory only "store"? If it's the latter, aka not persistent, then perhaps it can just be a map or a struct? 1/5 `Store` indicates persistent storage.

server/api/impl/store/apps.go
```diff
@@ -30,14 +31,19 @@ func (s *Store) LoadApp(appID apps.AppID) (*apps.App, error) {
...
 }
 
 func (s *Store) StoreApp(app *apps.App) error {
 	conf := s.conf.GetConfig()
 	if len(conf.Apps) == 0 {
 		conf.Apps = map[string]interface{}{}
 	}
+	// do not store manifest in the config
+	app.ID = app.Manifest.AppID
+	app.Manifest = nil
```
levb, line 46: 0/5 I'd advocate for complete separation of the `App` and `Manifest` structs. They represent different entities, and should not point at each other. The "client" should obtain them independently, as the use-case requires.

levb (changes requested):
I left some mostly cosmetic comments, still need to do another round of review on the substance.

apps/app.go
iomodo: It will be set by the system admin from the marketplace I think. We should provide the REST API to do it. There is a ticket for that.

server/command/install.go
iomodo: This command is for developers/testing. We will not use it in production.

apps/app.go
```diff
@@ -68,7 +115,9 @@ var DefaultBindingsCall = &Call{
 }
 
 type App struct {
+	ID       AppID     `json:"app_id"`
```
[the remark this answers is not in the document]
iomodo: `app.AppID` seems redundant to me

server/api/impl/admin/synchronize.go
```diff
@@ -0,0 +1,225 @@
...
+	"github.com/mattermost/mattermost-plugin-apps/apps"
+	"github.com/mattermost/mattermost-server/v5/model"
+	"github.com/pkg/errors"
+)
+
+const oldVersionKey = "update_from_version"
+
+// AppVersions describes versions for all the apps in all installations
+type AppVersions struct {
+	Apps      map[apps.AppID]string            `json:"apps"`
```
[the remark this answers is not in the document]
iomodo: Definitely an overkill for now :))

server/api/impl/admin/synchronize.go
iomodo: PP = proxy plugin. Why not?

server/api/impl/admin/synchronize.go
iomodo: Yes it is but I could not think of any other way implementing the distributed `callOnce`. If you have better suggestion I'm all ears.
I assumed 10 seconds is the **enough** time for all installations to finish the `KV.Delete` call.

server/api/impl/aws/provision_app.go
iomodo: Because it will be harder for us to provision apps, until we automatize the provisioning. Semantic names are easier - No need to update manifest with SHA-names. After the changes in provisioner we can pivot to content-based names.
Basically this is for v1.

server/plugin.go
iomodo: No strong reason, but mutex needs `p.API` for creation. I did not wanted to drag this field in the `Admin` as we already have the `pluginapi.client` there.

server/api/impl/store/store.go
iomodo: It's an in-memory store. This implementation is changed now. But I see store as an abstractions - no one should care weather it is an in-memory, persistent, file, db,...

server/api/impl/store/apps.go
iomodo: why? An app has a manifest - it's kinda natural. Can App exist without a manifest? I think not.
Separating them is possible, but it will change a lot of code and we will be doing lot's of lookups as well.

apps/app.go
catalintomai: cc @hanzei

iomodo: @catalintomai After syncing up with Ben, we came to the conclusion that installed app is automatically enabled(as configuring the app is the part of the installation process).

server/api/impl/admin/synchronize.go
```diff
@@ -0,0 +1,227 @@
...
+	registeredApps := adm.store.App().GetAll()
+
+	registeredAppsMap := map[apps.AppID]*apps.App{}
+	updatedAppVersionsMap := map[apps.AppID]string{}
+	// Update apps
+	for _, registeredApp := range registeredApps {
+		registeredAppsMap[registeredApp.Manifest.AppID] = registeredApp
+		manifest, err := adm.store.Manifest().Get(registeredApp.Manifest.AppID)
+		if err != nil {
+			return errors.Wrapf(err, "can't load manifest from store appID = %s", registeredApp.Manifest.AppID)
```
catalintomai, line 79: do we stop at the first failure?

server/command/install.go
catalintomai: I still miss smth. = ) - where do we call the AWS function provisioning in production?

server/api/impl/admin/synchronize.go
```diff
@@ -0,0 +1,227 @@
...
+			oldVersion := registeredApp.Manifest.Version
+			registeredApp.Manifest = manifest
+			if err := adm.store.App().Save(registeredApp); err != nil {
+				return errors.Wrapf(err, "can't store app - %s", registeredApp.Manifest.AppID)
+			}
+			updatedAppVersionsMap[registeredApp.Manifest.AppID] = oldVersion
+		}
+	}
+
+	// Add new apps as registered
```
catalintomai, line 95: since we cannot add them as smth. else (other than registered), maybe we can just say "register new apps"

apps/app.go
catalintomai: @iomodo, thanks, does that mean that we can remove that ticket you mentioned?

server/command/install.go
levb: @catalintomai we still need to implement a separate provisioning "script" that would create the lambdas/S3 and generate the manifest for an App.

apps/app.go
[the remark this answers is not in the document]
levb: A little redundancy IMO is ok to accomplish consistent, intention-revealing naming.

server/api/impl/admin/synchronize.go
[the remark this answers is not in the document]
levb: `Apps  map[apps.AppID]string` is less intention-revealing than `map[apps.AppID]apps.AppVersion`. FWIW, this one may benefit from being defined in the `apps` package, so `type VersionMap map[AppID]AppVersion` and then just 
`apps.VersionMap`. (I'm 0/5 on the actual name `VersionMap`)

server/api/impl/admin/synchronize.go
levb: I work on it, and I have to think about what it means every time. Why do we need this acronym? The keys are already in the namespace of the plugin.

server/api/impl/admin/synchronize.go
levb: I agree that we do not have a good mechanism for a `callOnce`. One of the reasons I am suggesting we do not even address these use-cases yet. 

> I assumed 10 seconds is the enough time for all installations to finish the KV.Delete call

Right, but then if there is no downside to making the timeout longer, why is it not even longer? And if there is a downside, then 10s seems like a very long time.

iomodo: You proposed to implement it, remember? This way there will be only one `OnStartup` call for the cluster.

This is used only once on startup of the plugin. 10 second delay for running app's `OnStartup` function might not be a big problem or is it?

apps/app.go
```diff
@@ -18,9 +18,52 @@ func (at AppType) IsValid() bool {
...
+		at == httpAsset
+}
+
+// AppStatus describes status of the app
+type AppStatus string
+
+const (
+	AppStatusRegistered AppStatus = "registered"
+	AppStatusEnabled    AppStatus = "enabled"
+	AppStatusDisabled   AppStatus = "disabled"
```
hanzei, line 44: This status seems to be used nowhere. Can we remove it for now?

hanzei:
@iomodo Heads up that https://github.com/mattermost/mattermost-plugin-apps/pull/57 is merged now.

server/http/restapi/marketplace.go
```diff
@@ -53,16 +53,16 @@ func (a *restapi) handleGetMarketplace(w http.ResponseWriter, req *http.Request,
...
-	for i := 0; i < len(apps); i++ {
+	result := make([]MarketplaceApp, len(localApps))
+	for i := 0; i < len(localApps); i++ {
 		result[i] = MarketplaceApp{
-			Manifest:  *apps[i].Manifest,
-			Installed: true,
+			Manifest:  *localApps[i].Manifest,
+			Installed: localApps[i].Status == apps.AppStatusInstalled,
 		}
 	}
```
hanzei, line 68: Do we still need `getRemoteApps` and `getLocalApps()` or can we just drop it given that the data is now in `assets/apps.json`?

iomodo: Actually we don't. I'll delete that as well as the merging part.

levb (approved):
I am good with merging this PR and iterating from there

PR #50, opened 2021-01-14
MM-29584: Make apps plugin hook invocation/registration conditional

catalintomai:
#### Summary
Make sure we do not invoke the MessageHasBeenPosted() handler functionality unless we explicitly set that functionality flag (allowWebhook) to true

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-29584

#### Related PRs
Server: https://github.com/mattermost/mattermost-server/pull/16769


server/plugin.go
```diff
@@ -130,6 +131,16 @@ func (p *Plugin) UserHasLeftTeam(pluginContext *plugin.Context, tm *model.TeamMe
 }
 
 func (p *Plugin) MessageHasBeenPosted(pluginContext *plugin.Context, post *model.Post) {
+	shouldProcessMessage, err := p.Helpers.ShouldProcessMessageNoBotCheck(post, p.api.Configurator.GetConfig().BotUserID)
+	if err != nil {
+		p.mm.Log.Error(fmt.Sprintf("Error while checking if the message should be processed %v", err))
```
hanzei, line 136: Should we use `github.com/pkg/errors.wrap` instead?

hanzei (approved):
LGTM, one nit pick

server/plugin.go
larkox: Another option (and I think is how log is supposed to be used) is something like `p.mm.Log.Error(message, "error", err.Error())`.

server/plugin.go
```diff
@@ -145,6 +145,16 @@ func (p *Plugin) UserHasLeftTeam(pluginContext *plugin.Context, tm *model.TeamMe
 }
 
 func (p *Plugin) MessageHasBeenPosted(pluginContext *plugin.Context, post *model.Post) {
+	shouldProcessMessage, err := p.Helpers.ShouldProcessMessage(post, plugin.BotID(p.api.Configurator.GetConfig().BotUserID))
+	if err != nil {
+		p.mm.Log.Error("Error while checking if the message should be processed %v", "err", err.Error())
```
larkox, line 150: The message here will be shown as `Error while checking if the message should be processed %v`, since no formatting is done to this string (the key value passed will just add another field to the log message, not do any kind of formatting). I would remove the %v.

server/manifest.go
```diff
@@ -19,8 +19,8 @@ const manifestStr = `
   "min_server_version": "5.26.0",
   "server": {
     "executables": {
-      "linux-amd64": "server/dist/plugin-linux-amd64",
       "darwin-amd64": "server/dist/plugin-darwin-amd64",
+      "linux-amd64": "server/dist/plugin-linux-amd64",
```
hanzei, line 23: Unrelated change?

server/plugin.go
hanzei: Good catch @larkox :+1:

hanzei (changes requested):

server/manifest.go
catalintomai: it is reordering, the result of some gofmt(?) op - should be innocuous(?). Do you think otherwise?

hanzei: Yes, it's definitely innocuous. Still, I would like to know why is this happening. `make apply` (which generates `server/manifest.go` generates the version with `linux-amd64` first.

catalintomai: @hanzei - just checked - on my box the `make apply` (I guess invoked as part of `make dist`) is updating the order in the file.

hanzei: Oh, well the mysterious behaviors of go. Could you please fix the CI failure and we check if CI is fine with our change?

catalintomai: The CI failure is because this PR depends of the server PR - which needs to go first. They should go away once the server PR is in.

hanzei: Why do you need to merge the server PR first? Can't you point to the server PR commit?

catalintomai: I'd rather that we have the server PR first and then this PR. Anyway this PR will not be merged without the CI tests passing - so I'm not quite sure how the CI tests passing/not passing influence the dev CR approval(?)

hanzei: > so I'm not quite sure how the CI tests passing/not passing influence the dev CR approval

You will need to update `go.mod` to point to the newly added commit in https://github.com/mattermost/mattermost-server/pull/16769. That shouldn't require a re-review, right?

catalintomai: Does it? Unless I miss smth, wouldn't the CI system ensure that the `go.mod` is the correct one? 
Validating that `go.mod` is updated - CI would not pass without it. 
Validating that the `go.mod` update is the correct one - the commit id is rather opaque to the user and CI would not pass without the right commit being pointed to?

hanzei: My approval is not tied to the `go.mod` changes but rather to CI passing in general. If CI doesn't pass, additional changes are required and hence I won't approve right now. Does that make sense to you? Happy to discus in a meeting.

larkox (approved):

server/plugin.go
```diff
@@ -145,6 +145,16 @@ func (p *Plugin) UserHasLeftTeam(pluginContext *plugin.Context, tm *model.TeamMe
 }
 
 func (p *Plugin) MessageHasBeenPosted(pluginContext *plugin.Context, post *model.Post) {
+	shouldProcessMessage, err := p.Helpers.ShouldProcessMessage(post, plugin.BotID(p.api.Configurator.GetConfig().BotUserID))
```
larkox, line 148: Since we are using the proxy bot here, it will consider the shouldProcessMessage as if the proxy bot. Looking at the shouldprocessmessage behaviour, this would filter out messages sent directly to the app bot. This logic probably should be handled in the Notify function, or even further down.

larkox (changes requested):
Just realized that this may be filtering too many messages.

server/manifest.go
catalintomai: I guess I see the review process as smth independent of the CI passing. CI can fail for multiple reasons (test failure, linting/check-style failure, redux/server dependency, etc.) I have the trust in that the developer will fix those issues and and I'd not wait for him to fix them to approve their PR ( that would be like rubber-stamping the CI result - it is almost as if one reviewer would wait for another reviewer(Ci in this case)  approval before giving their own approval). Should one delay(or even tie their approval) to a lint failure issue? If the CI somehow detects some major issue(s) that impact the already reviewed code in  significant manner, then I trust that the developer (for his and everyone's sake) will submit it for re-review (that happens in my experience in very few of the cases).

server/plugin.go
catalintomai: @larkox , can you detail the scenario? Are there posts created by the apps bot that we want the apps bot to notify the app on? The `shouldProcessMessage` explicitly precludes that bot-to-bot messaging, we can relax it by adding another option, but I'd like to understand our scenario first?

larkox: After taking a look at the code, without more options, it will be fine. Nevertheless, I still believe it is in base wrong. Here a few examples of what will happen if we were to add a few more options.

If we were to add `AllowBots`:
- App A will think it should process it's own posts, which it shouldn't.
- App A will not think it should process Proxy plugin bot messages.

If we were to add `OnlyBotDMs`:
- App A will ignore messages on its own DM.
- App A will process messages on proxy app bot DM.

larkox (approved):

server/plugin.go
larkox: I am still worried about future changes in this function, since may introduce future confusion, but for the time being, works as expected.

levb (approved):

hanzei (commented):
Code LGTM

catalintomai:
> Code LGTM

@hanzei, can we establish a standard for the "requested changes"-> "merging is blocked" step (at least for our team)? I propose that if a PR has two devs approvals and the remaining comments are not "blocking", to avoid adding the "requested changes" blocking tag (irrespective of CI stamp of approval - one cannot checkin anyway without CI passing and we have trust that we don't do anything nefarious like commenting out tests).

hanzei:
@catalintomai I'm still subconscious about the changes in `server/manifest.go`. What ever commands you did run locally, they should have not changed that file. Given that I can't reproduce that change locally there might be something fishing with the file generation. This is why I insisted in fixing the CI issue. CI will act as a single source of truth here and help me understand where the issue is.

hanzei (approved):
I just figured out that the `go.mod` update caused the changed in `server/manifest.go`. LGTM :+1:

levb:
@catalintomai @DHaussermann ping?

DHaussermann (approved):
Tested and passed 
- Not much in terms of functional testing 
- Regression tested conventional bot posts 
- Regression tested Post made by an MM apps bot such as Service Now

No issues found.

LGTM!

PR #52, opened 2021-01-19
MM-29853: Expose plugin-apps REST APIs

catalintomai:
#### Summary
Expose plugin-apps REST APIs

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-29583


modelapps/mattermost_client_pp.go
```diff
@@ -0,0 +1,123 @@
...
+
+func (c *ClientPP) DoApiDelete(url string) (*http.Response, *model.AppError) {
+	return c.DoApiRequest(http.MethodDelete, c.ApiUrl+url, "", "")
+}
+
+func (c *ClientPP) DoApiRequest(method, url, data, etag string) (*http.Response, *model.AppError) {
+	return c.doApiRequestReader(method, url, strings.NewReader(data), etag)
+}
+
+func (c *ClientPP) doApiRequestReader(method, url string, data io.Reader, etag string) (*http.Response, *model.AppError) {
```
catalintomai, line 81: Would like to re-use the mm-server client4.go methods instead, however, they'll need to be factored out of client4.go first.

modelapps/mattermost_client_pp.go
```diff
@@ -0,0 +1,123 @@
...
+
+	if len(etag) > 0 {
+		rq.Header.Set(HEADER_ETAG_CLIENT, etag)
+	}
+
+	if len(c.AuthToken) > 0 {
+		rq.Header.Set(HEADER_AUTH, c.AuthType+" "+c.AuthToken)
+	}
+
+	if c.HttpHeader != nil && len(c.HttpHeader) > 0 {
```
larkox, line 95: If I am not mistaken, len should return 0 on nil maps, so the first check should not be needed.

modelapps/mattermost_client_pp.go
```diff
@@ -0,0 +1,123 @@
...
+	HttpHeader map[string]string // Headers to be copied over for each request
+
+	// TrueString is the string value sent to the server for true boolean query parameters.
+	trueString string
+
+	// FalseString is the string value sent to the server for false boolean query parameters.
+	falseString string
+}
+
+func NewAPIClientPP(url string) *ClientPP {
```
larkox, line 37: What does Client PP stands for?

larkox (approved):
LGTM, just one nit and one question.

modelapps/mattermost_client_pp.go
catalintomai: PP-ProxyPlugin (I think Shota introduced it first) - it is the golang client driver for the PP REST API.

levb (changes requested):
1/5 s/modelapps/apps/ and 0/5 move all Client stuff into a separate package (/apps/mmclient?)

this way, `apps.Form` or `apps.Call` is intention-revealing and compact; ditto mmclient.Client. The `api` folder can be adjusted since there's little left in there; separate PR.

server/api/impl/appservices/subscriptions.go
```diff
@@ -3,12 +3,22 @@
...
 
-import "github.com/mattermost/mattermost-plugin-apps/server/api"
+import (
+	"github.com/mattermost/mattermost-plugin-apps/apps"
+)
 
-func (a *AppServices) Subscribe(sub *api.Subscription) error {
-	return a.store.StoreSub(sub)
+func (a *AppServices) Subscribe(sub *apps.Subscription) error {
+	err := a.store.StoreSub(sub)
```
levb, line 11: nit: why?

levb (approved):
nice! thanks for some of the cleanup!

server/api/impl/appservices/subscriptions.go
catalintomai: leftover from debugging, will remove.

PR #53, opened 2021-01-20
[MM-31782] static asset proxy

iomodo:
#### Summary
This PR adds static asset proxy

#### Ticket Link
  Fixes https://mattermost.atlassian.net/browse/MM-31782


server/api/proxy.go
```diff
@@ -7,6 +7,7 @@ type Proxy interface {
 	GetBindings(*Context) ([]*Binding, error)
 	Call(SessionToken, *Call) *CallResponse
 	Notify(cc *Context, subj Subject) error
+	Asset(AppID, string) ([]byte, error)
```
levb, line 10: 1/5 s/Asset/GetStatic/ or GetAsset

server/http/restapi/asset.go
```diff
@@ -0,0 +1,45 @@
...
+		return
+	}
+
+	sessionID := req.Header.Get("MM_SESSION_ID")
+	if sessionID == "" {
+		httputils.WriteUnauthorizedError(w, errors.New("no user session"))
+		return
+	}
+
+	_, err := a.api.Mattermost.Session.Get(sessionID)
```
levb, line 24: Why are we doing the session check? I thought it comes into the plugin already validated?

levb (changes requested):

server/http/restapi/asset.go
```diff
@@ -0,0 +1,45 @@
...
+
+import (
+	"net/http"
+
+	"github.com/mattermost/mattermost-plugin-apps/server/api"
+	"github.com/mattermost/mattermost-plugin-apps/server/utils/httputils"
+	"github.com/pkg/errors"
+)
+
+func (a *restapi) assetGet(w http.ResponseWriter, req *http.Request) {
```
catalintomai, line 11: 1/5 s/assetGet/handleGetAsset to be in sync with the other non-KV router mappings.

server/http/restapi/asset.go
```diff
@@ -0,0 +1,45 @@
...
+	query := req.URL.Query()
+
+	assetName := query.Get("name")
+	appID := query.Get("app_id")
+
+	// TODO verify that request is from the correct app
+
+	data, err := a.api.Proxy.Asset(api.AppID(appID), assetName)
+	if err != nil {
+		httputils.WriteBadRequestError(w, err)
```
catalintomai, line 38: On success do we return the OK(200) in the header?

server/api/impl/proxy/asset.go
```diff
@@ -0,0 +1,48 @@
...
+	"github.com/pkg/errors"
+)
+
+func (p *Proxy) Asset(appID api.AppID, assetName string) ([]byte, error) {
+	app, err := p.store.LoadApp(appID)
+	if err != nil {
+		return nil, errors.Wrapf(err, "can't load app - %s", appID)
+	}
+	for _, asset := range app.Manifest.Assets {
+		if asset.Name == assetName {
```
catalintomai, line 20: nit: flattening the "arrow code" (https://blog.codinghorror.com/flattening-arrow-code/) and maybe replacing the ifs with a switch

server/api/impl/proxy/asset.go
```diff
@@ -0,0 +1,48 @@
...
+			if asset.Type == api.S3Asset {
+				data, err := p.awsClient.S3FileDownload(asset.Bucket, asset.Key)
+				if err != nil {
+					return nil, errors.Wrapf(err, "can't download s3 file for appID - %s, assetName - %s from %s/%s", appID, assetName, asset.Bucket, asset.Key)
```
catalintomai, line 24: nit: instead of "s3 file" could use `api.S3Asset` type.

server/http/restapi/asset.go
iomodo: You are correct

server/http/restapi/asset.go
iomodo: Yes, `Write` method does this automatically, also determines `Content-Type`, adds `Content-Length` if needed, etc.

server/api/constants.go
```diff
@@ -37,8 +37,8 @@ const (
 	KVPath          = "/kv"
 	SubscribePath   = "/subscribe"
 	UnsubscribePath = "/unsubscribe"
-
-	BindingsPath = "/bindings"
+	AssetPath       = "/asset"
```
levb, line 40: 1/5 cc @hanzei should this be `/static`, as in the plugins?

server/api/impl/proxy/asset.go
```diff
@@ -0,0 +1,52 @@
...
+		errorMessage := fmt.Sprintf("can't download %s for appID - %s, assetName - %s from", asset.Type, appID, assetName)
+		switch asset.Type {
+		case apps.S3Asset:
+			data, err := p.awsClient.S3FileDownload(asset.Bucket, asset.Key)
+			if err != nil {
+				return nil, errors.Wrapf(err, "%s %s/%s", errorMessage, asset.Bucket, asset.Key)
+			}
+			return data, nil
+		case apps.HTTPAsset:
+			/* #nosec G107 */
```
levb, line 34: with HTTP we should pipe from resp.Body into the incoming' response, not return the bytes. Ideally ditto for S3. A separate (performance) ticket?

Also, size limit?

server/http/restapi/asset.go
levb: We should pass the HTTP code from the upstream, not report a BAD REQUEST.

levb (dismissed):
I'm ok with merging as is, but we need 2 tickets - one (performance) to pipe the bytes from the upstream to downstream without a `ReadAll`, and one for passing back the HTTP status code.

Gateway errors should be returned as 502-504, as appropriate

server/api/constants.go
hanzei: Good question. I like `static` because it indicates the nature of the files. So there would be  `static assets` served under `/static` and `dynamic assets` generated and served by the app via the lambda.

catalintomai (approved):

server/http/restapi/asset.go
```diff
@@ -0,0 +1,27 @@
...
+
+	"github.com/mattermost/mattermost-plugin-apps/apps"
+	"github.com/mattermost/mattermost-plugin-apps/server/utils/httputils"
+)
+
+func (a *restapi) handleGetStaticAsset(w http.ResponseWriter, req *http.Request, actingUserID string) {
+	query := req.URL.Query()
+
+	assetName := query.Get("name")
+	appID := query.Get("app_id")
```
hanzei, line 14: 1/5 Using query strings to indicate which asset is requested seems a bit verbose to type. Maybe we want to use the URL path to indicate that i.e. `/static/$APP_ID/$NAME`

server/http/restapi/asset.go
hanzei: @iomodo Do do plan to add that?

server/api/impl/aws/provision_app.go
```diff
@@ -187,31 +220,21 @@ func (c *Client) createFunction(zipFile io.Reader, function, handler, runtime, r
...
+	for _, asset := range assets {
+		key := getAssetFileKey(manifest.AppID, manifest.Version, asset.name)
+		if err := c.S3FileUpload(key, asset.file); err != nil {
+			return nil, errors.Wrapf(err, "can't provision asset - %s with key - %s", asset.name, key)
+		}
+		manifest.Assets = append(manifest.Assets, apps.Asset{
+			Name:   asset.name,
+			Type:   apps.S3Asset,
+			Bucket: c.appsS3Bucket,
+			Key:    key,
```
hanzei, line 236: I'm wondering if we even need need all of that metadata for an Asset. 

1. Type: Is it safe to assume that an HTTP App only has HTTP assets and an AWS App only has S3 Assets? I don't see a case were we would mix that.
2. Key: Why do we need to store they key if we can generate it from the other data?
3. Why no we need to store the bucket, if that is fixed?

server/api/impl/proxy/asset.go
```diff
@@ -0,0 +1,52 @@
...
+			resp, err := http.Get(asset.URL)
```
hanzei, line 35: Can we simplify here and assume assets are below a well known path i.e. `$ROOT_URK/static/$NAME`?

server/http/restapi/asset.go
```diff
@@ -0,0 +1,27 @@
...
+
+	// TODO verify that request is from the correct app
```
hanzei, line 16: Why do we need to check for that? If an app dev uses static assets to as an icon in a modal, there won't be an app connected to that request, correct?

hanzei (changes requested):
I think we can simplify the overall design here

server/api/impl/proxy/asset.go
```diff
@@ -0,0 +1,48 @@
...
+	"io/ioutil"
+	"net/http"
+
+	"github.com/pkg/errors"
+
+	"github.com/mattermost/mattermost-plugin-apps/apps"
+	"github.com/mattermost/mattermost-plugin-apps/server/api/impl/aws"
+)
+
+func (p *Proxy) GetAsset(appID apps.AppID, assetName string) (*http.Response, error) {
```
hanzei, line 18: It's a bit odd forge a `http.Response` here, given that we only want to communicate a SatusCode and a Body. Should return `io.ReadCloser, int, error` instead?

server/http/restapi/asset.go
hanzei: Gentle bump on that question

server/http/restapi/asset.go
```diff
@@ -0,0 +1,56 @@
...
+
+	resp, err := a.api.Proxy.GetAsset(apps.AppID(appID), assetName)
+	if err != nil {
+		httputils.WriteBadRequestError(w, err)
+		return
+	}
+
+	copyHeaders(w.Header(), resp.Header)
```
hanzei, line 38: Do we really need that? For AWS apps `resp.Header` is `nil` anyway. By dropping that line AWS and HTTP act the same which is what we want here, right?

hanzei (approved):
Great work :+1: A few nit picks, but nothing blocking

server/http/restapi/asset.go
iomodo: We just using user tokes for the authorization so "malicious" app can get a static asset of a different app. It's not a problem for now but might become later.

server/api/impl/proxy/asset.go
iomodo: Sure, but I did not want to loose the headers from the http apps.

server/http/restapi/asset.go
iomodo: Actually I'm 0/5. I still left headers there, just in case. AWS will not have any headers but HTTP apps might. I don't know if we would ever need them though.

server/http/restapi/asset.go
hanzei: Let me ask the question in another way: Do we consider the content of static assets to be confidential information?

server/api/impl/proxy/asset.go
hanzei: Then my questions boils down to if we need the headers or not: https://github.com/mattermost/mattermost-plugin-apps/pull/53#discussion_r583874562

server/http/restapi/asset.go
hanzei: I'm 2/5 that we can add the header later if we need them. (Maybe as a whitelist?) For now, we should we strict as dropping them at some point is harder.

server/http/restapi/asset.go
iomodo: I think so. There might be some data files in the static folder, not just an icon.

server/http/restapi/asset.go
iomodo: Agree.

apps/mmclient/mattermost_client_pp.go
```diff
@@ -45,6 +45,7 @@ func NewAPIClientPP(url string) *ClientPP {
 
 func (c *ClientPP) KVSet(id string, prefix string, in map[string]interface{}) (map[string]interface{}, *model.Response) {
 	query := fmt.Sprintf("%v/kv/%v?prefix=%v", APIPathPP, id, prefix)
+	//nolint
```
hanzei, line 48: Do we really need that?

go.mod
```diff
@@ -6,6 +6,7 @@ require (
 	github.com/aws/aws-sdk-go v1.35.33
 	github.com/dgrijalva/jwt-go v3.2.0+incompatible
 	github.com/golang/mock v1.4.3
+	github.com/google/go-cmp v0.5.1 // indirect
```
hanzei, line 9: That seems like an unrelated change

hanzei (approved):
The code looks much cleaner now :+1:

server/http/restapi/asset.go
levb: I am confused... `/static` is just public images, nothing "private" about it AFAIC.

PR #55, opened 2021-01-22
add multiselect field option

jfrerich:
#### Summary
The api needs to accept the multiselect boolean value in the Field Struct

#### Ticket Link
n/a


server/api/field.go
```diff
@@ -36,6 +36,7 @@ type Field struct {
 
 	// Select props
 	SelectRefresh       bool           `json:"refresh,omitempty"`
+	SelectMultiselect   bool           `json:"multiselect,omitempty"`
```
levb, line 39: @jfrerich 1/5 `SelectMultiselect` is consistent, but redundant a name. Perhaps, just `Multiselect`, or even `Multi`?

@mickmister this (multiple selection support) seems like a bigger deal than just adding the flag for it, or are other pieces in place? Autocomplete worries me in particular.

levb (dismissed):

server/api/field.go
jfrerich: I think the `Select` prefix is valuable because it is grouping Field values that are explicitly related to a select form.  I suppose a `SelectIsMulti` my be more intention revealing, implying a boolean option for a select field

jfrerich: For reference, I have not tested this for autocompletion, but have multiselect fields working in the zendesk app in this PR https://github.com/mattermost/mattermost-app-zendesk/pull/21

mickmister:
@jfrerich I'm thinking we hold off on merging this until it is supported in the client. What do you think?

server/api/field.go
mickmister: > but have multiselect fields working in the zendesk app in this PR

Are you saying that it is already supported in the webapp?

jfrerich: Yes. There is an oddity where the multiselect opens with an empty value already in the select box, but it is working.

jfrerich: image

jfrerich: Here is the multiselect with multiple selected fields

image

mickmister: > Yes. There is an oddity where the multiselect opens with an empty value already in the select box, but it is working.

Can you make a ticket for this if it is not resolved?

mickmister: 1/5 `SelectIsMulti` sounds good to me

mickmister (approved):

apps/field.go
```diff
@@ -61,6 +61,8 @@ type Field struct {
 	// TODO: ModalLabel should default to Label, Name
 	ModalLabel string `json:"modal_label"`
 
+	SelectIsMulti bool `json:"multiselect,omitempty"`
```
larkox, line 64: Since everything is now heavily documented, should we add here some documentation too?

larkox (approved):

hanzei:
@jfrerich Is this PR needed for V1 of either of the two apps?

DHaussermann (approved):
Tested and passed.
- Verified with Zendesk App 
- Control is working as expected
- Data sent is accurate 
- Follow-up ticket needed for RN client support

LGTM!

jfrerich:
@levb, this is ready to merge. Dylan and I tested again today! For some reason, I do not have merge permissions.

hanzei:
@jfrerich Heads up that docs need to be submitted to https://developers.mattermost.com/integrate/apps/

hanzei:
@levb Are you fine with merging this PR?=

jfrerich:
@hanzei we discussed in the meeting this morning that we will merge this PR, but will not yet advertise it in the docs for the following reasons.
* command form is not yet handled
* mobile has to be handled by the app

For zendesk, which does use this function, I don't have multiselect via a slash command and will handle the mobile case in the app.
cc: @mickmister

jfrerich:
Here is the PR for React Native implementation.  I've added a note to add the developer docs when that Ticket is worked on.

https://mattermost.atlassian.net/browse/MM-34194

hanzei:
@jfrerich Sounds good to me :+1:

PR #56, opened 2021-01-29
Rework embedded forms

larkox:
#### Summary
Add needed types, and example to hello app.

#### Ticket Link
Redux: Soon
Webapp: https://github.com/mattermost/mattermost-webapp/pull/7423


server/api/embedded_form.go
```diff
@@ -0,0 +1,8 @@
+package api
+
+type EmbeddedForm struct {
```
levb, line 3: 1. 1/5 maybe we should stop using the term `Form` here as it may be confusing with `api.Form`, and replace it with something like `type InPost` or `Embedded`
2. 0/5 maybe we should use the `AppID` in the binding - this way in theory we can attach bindings from different apps to a single Post. It's an esoteric possibility, but I'd rather not preclude it in he data model.

levb (approved):
Some data model feedback, otherwise LGTM.

server/api/embedded_form.go
larkox: 1. I tried to start a discussion about naming in community: https://community.mattermost.com/core/pl/zgn3medxujfcj84igbxxzcfxge . When we get to a decission I will make the changes.
2. Right now we have no way to impede posting a form that will perform an action on a different app. I feel that is a security vector we should control. That is why I wanted to put that on a top level. Nevertheless, if we want to go rogue in that direction, I don't mind using the id on the binding.

catalintomai, line 3: For (my) clarity - an EmbeddedForm (or another name) is still a Binding (with a implicit location associated)?  Do we consider Binding as the root/base type and EmbeddedForm as a specialization of it?

larkox: For the time being, it is not a binding nor a form. It could be a binding, but then we would have to extend the properties a binding to allow all needed by a embed form. We could go back to my original idea (serveral months old) (`type Binding struct {AppID: string; Location: string; Call: AppCall; Props: interface{}}; Bindings: AppBindings[]`) and then define some utility types like `type PostEmbedProps struct {Title: string; Text: string;}` or `type ChannelHeaderProps struct {Label: string; IconURL: string}`.

larkox:
After coming back and forth in my head, I ended up leaving the whole "attachment" as a binding. We already use this structure "for everything", so if we need to redo the structure to support new stuff "only for embedded bindings" we can add them to this struct (or consider dividing this struct later).

CC @levb

server/examples/go/hello/submit_survey.go
```diff
@@ -0,0 +1,36 @@
...
+	}
+	selected := location[len(location)-1]
+	if selected == "button" {
+		bot := mmclient.AsBot(c.Context)
+		p := &model.Post{
+			Id:      c.Context.PostID,
+			Message: "The survey will not be sent",
+		}
+		_, err := bot.UpdatePost(c.Context.PostID, p)
+		fmt.Println(err)
```
catalintomai, line 29: do we need these?

larkox: Nope, I added those when I was debugging and forgot to remove them :facepalm:

catalintomai (approved):

levb (approved):

PR #57, opened 2021-02-04
[MM-32476] Apps in the Marketplace

hanzei:
#### Summary
This PR adds an `/marketplace` endpoint to the plugin. It's used by the webapp to fetch the list of installed and available apps. The latter is a hard coded list for now as discussed in the Apps meeting.

I've added Zendesk as a POW. It only works if the App runs locally, given that it's not yet in AWS.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-32476
https://mattermost.atlassian.net/browse/MM-32535
https://mattermost.atlassian.net/browse/MM-32539

### DIFF
--- a/server/http/restapi/marketplace.go
+++ b/server/http/restapi/marketplace.go
@@ -1,0 +1,145 @@
+package restapi
+
+import (
+	"net/http"
+	"sort"
+	"strings"
+
+	"github.com/mattermost/mattermost-plugin-apps/apps"
+	"github.com/mattermost/mattermost-plugin-apps/server/utils/httputils"
+
+	"github.com/mattermost/mattermost-server/v5/model"
+	"github.com/pkg/errors"
+)
+
+type MarketplaceApp struct {
+	Installed bool                     `json:"installed"`
+	Labels    []model.MarketplaceLabel `json:"labels,omitempty"`
+	Manifest  apps.Manifest            `json:"manifest"`
+}
+
+func (a *restapi) handleGetMarketplace(w http.ResponseWriter, req *http.Request, actingUserID string) {
+	filter := req.URL.Query().Get("filter")
+
+	localApps, err := a.getLocalApps()
+	if err != nil {
+		httputils.WriteInternalServerError(w, err)
+		return
+	}
+
+	remoteApps, err := a.getRemoteApps()
+	if err != nil {
+		httputils.WriteInternalServerError(w, err)
+		return
+	}
+
+	apps := mergeApps(remoteApps, localApps)
+
+	// Filter plugins.
+	var result []MarketplaceApp
+	for _, a := range apps {
+		if appMatchesFilter(a.Manifest, filter) {
+			result = append(result, a)
+		}
+	}
+
+	// Sort result alphabetically.
+	sort.SliceStable(result, func(i, j int) bool {
+		return strings.ToLower(result[i].Manifest.DisplayName) <
+			strings.ToLower(result[j].Manifest.DisplayName)
+	})
+
+	httputils.WriteJSON(w, result)
+}
+
+func (a *restapi) getLocalApps() ([]MarketplaceApp, error) {
+	apps, _, err := a.api.Admin.ListApps()
+	if err != nil {
+		return nil, errors.Wrap(err, "Failed to list local apps")
+	}
+
+	result := make([]MarketplaceApp, len(apps))
+	for i := 0; i < len(apps); i++ {
+		result[i] = MarketplaceApp{
+			Manifest:  *apps[i].Manifest,
+			Installed: true,
+		}
+	}
+
+	return result, nil
+}
+
+func (a *restapi) getRemoteApps() ([]MarketplaceApp, error) {
+	m := apps.Manifest{
+		AppID:       "zendesk",
+		Type:        apps.AppTypeHTTP,
+		DisplayName: "Zendesk",
+		Description: "A Zendesk App.",
+		HomepageURL: "https://github.com/mattermost/mattermost-app-zendesk",
+		HTTPRootURL: "http://localhost:4000/mattermost/manifest.json",
+		RequestedPermissions: apps.Permissions{
+			apps.PermissionActAsUser,
+			apps.PermissionActAsBot,
+		},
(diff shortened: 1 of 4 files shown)


hanzei:
@catalintomai You might also want to take a look

server/http/restapi/marketplace.go
```diff
@@ -0,0 +1,145 @@
...
+		result[i] = MarketplaceApp{
+			Manifest:  *apps[i].Manifest,
+			Installed: true,
+		}
+	}
+
+	return result, nil
+}
+
+func (a *restapi) getRemoteApps() ([]MarketplaceApp, error) {
```
catalintomai, line 72: Can we add a comment that talks about the difference(s) between local and remote apps?

server/http/restapi/marketplace.go
```diff
@@ -0,0 +1,145 @@
...
+
+	"github.com/mattermost/mattermost-plugin-apps/apps"
+	"github.com/mattermost/mattermost-plugin-apps/server/utils/httputils"
+
+	"github.com/mattermost/mattermost-server/v5/model"
+	"github.com/pkg/errors"
+)
+
+type MarketplaceApp struct {
+	Installed bool                     `json:"installed"`
```
iomodo, line 16: What `Installed` means?
We have 3 states of the apps for now:
1. `registered`/`listed`
2. `enabled`
3. `disabled`

server/http/restapi/marketplace.go
```diff
@@ -0,0 +1,145 @@
...
+	sort.SliceStable(result, func(i, j int) bool {
+		return strings.ToLower(result[i].Manifest.DisplayName) <
+			strings.ToLower(result[j].Manifest.DisplayName)
+	})
+
+	httputils.WriteJSON(w, result)
+}
+
+func (a *restapi) getLocalApps() ([]MarketplaceApp, error) {
+	apps, _, err := a.api.Admin.ListApps()
```
iomodo, line 56: Just to clarify, apps stored here are all apps `registered` for this installation. Some of them might be already installed by the instance admin, others might not. Installed apps might be `enabled` or `disabled`.

server/http/restapi/marketplace.go
```diff
@@ -0,0 +1,145 @@
...
+	"github.com/pkg/errors"
+)
+
+type MarketplaceApp struct {
+	Installed bool                     `json:"installed"`
+	Labels    []model.MarketplaceLabel `json:"labels,omitempty"`
+	Manifest  apps.Manifest            `json:"manifest"`
+}
+
+func (a *restapi) handleGetMarketplace(w http.ResponseWriter, req *http.Request, actingUserID string) {
```
larkox, line 21: Just to make sure. We are checking the user is authorized, but then we don't check anything else. Therefore, in theory, any user could fetch this information. I don't think there is any "sensitive information" returned here, and I understand eventually we want to allow all users to access the marketplace, but I wanted to be 100% sure.

server/http/restapi/marketplace_test.go
```diff
@@ -0,0 +1,90 @@
...
+	"github.com/stretchr/testify/assert"
+)
+
+func TestMergeApps(t *testing.T) {
+	for name, test := range map[string]struct {
+		a        []MarketplaceApp
+		b        []MarketplaceApp
+		expected []MarketplaceApp
+	}{
+		"same slices": {
```
larkox, line 16: Just for completion, maybe add a test with both slices empty?

larkox (approved):
LGTM

server/http/restapi/marketplace.go
hanzei: In these terms it means `enabled`.

Are these three terms used somewhere in code or docs? I would like to use consistent naming.

server/http/restapi/marketplace.go
hanzei: Should `master` already behave that way? Right now Apps are only stored when they get installed or provisioned.

I'm hesitant to base the PR on changes that are not yet in `master`.

server/http/restapi/marketplace.go
hanzei: Good question. The behaviour here mimics the one from the Marketplace endpoint on the MM server. I don't see any reason not to open the endpoint for all users.

server/http/restapi/marketplace_test.go
hanzei: :+1: added in a429ba6

server/http/restapi/marketplace.go
iomodo: No. [PR](https://github.com/mattermost/mattermost-plugin-apps/pull/46) is not merged yet.

server/http/restapi/marketplace.go
catalintomai: In this context, can we live with the definition registered+enabled == installed? Can one app be enabled but not listed/registered?

iomodo (approved):
Discussed with Ben offline, Approving as it is. Will make necessary changes myself when App synchronization PR is merged

server/http/restapi/marketplace.go
hanzei: A comment will make much more sense once https://github.com/mattermost/mattermost-plugin-apps/pull/46 is merged.

server/http/restapi/marketplace.go
iomodo: @catalintomai Yes, that is ture.
Should I change app state names as `Registered`, `Installed`, `Uninstalled` to be more inline with the marketplace?

catalintomai: I'd say having common naming is a definite must. Maybe sending a message to the MM Apps channel with the proposed state naming (and meaning of each) would close this discussion for all.Like: 
registered/listed - means that the app has a manifest present in s3 bucket and is present in the apps.json bundle file, etc.
enabled - ...
installed - ...
uninstalled:...

catalintomai: @iomodo, let's use the naming of the marketplace then.

catalintomai (approved):

levb (approved):

PR #58, opened 2021-02-04
Fix panic that happens on server restart

larkox:
#### Summary
The error came from the embedded struct (StoredConfig) not being init on server restart.

This was due to the server not calling "OnConfigurationChange" after activate. On Enable it behaves like that, (first activate, then onconfigurationchange), but not on server restart.

#### Ticket Link
None


levb (approved):

catalintomai (approved):

hanzei:
Fixes https://community-daily.mattermost.com/core/pl/xyomyzk34pg48gra54zbeh9ixc

PR #62, opened 2021-02-11
Provision static assets

iomodo:
#### Summary
This PR will provision static assets in S3.

### DIFF
--- a/server/api/impl/aws/aws_keys.go
+++ b/server/api/impl/aws/aws_keys.go
@@ -36,10 +36,10 @@ func getFunctionName(appID apps.AppID, version apps.AppVersion, function string)
 	return name, nil
 }
 
-// getManifestFileName generates key for a specific manifest in S3,
+// getManifestFileKey generates key for a specific manifest in S3,
 // key can be 1024 characters long.
 func getManifestFileKey(appID apps.AppID, version apps.AppVersion) string {
-	return fmt.Sprintf("manifests/%s_%s_manifest", appID, version)
+	return fmt.Sprintf("manifests/%s_%s", appID, version)
 }
 
 // getAssetFileKey generates key for a specific asset in S3,
--- a/server/api/impl/aws/aws_keys.go
+++ b/server/api/impl/aws/aws_keys.go
@@ -1,0 +1,48 @@
+// Copyright (c) 2019-present Mattermost, Inc. All Rights Reserved.
+// See License for license information.
+
+package aws
+
+import (
+	"crypto/sha256"
+	"encoding/hex"
+	"fmt"
+
+	"github.com/mattermost/mattermost-plugin-apps/apps"
+	"github.com/pkg/errors"
+)
+
+// getFunctionName generates function name for a specific app
+// name can be 64 characters long.
+func getFunctionName(appID apps.AppID, version apps.AppVersion, function string) (string, error) {
+	if len(appID) > appIDLengthLimit {
+		return "", errors.Errorf("appID %s too long, should be %d bytes", appID, appIDLengthLimit)
+	}
+	if len(version) > len(versionFormat) {
+		return "", errors.Errorf("version %s too long, should be in %s format", version, versionFormat)
+	}
+	name := fmt.Sprintf("%s-%s-%s", appID, version, function)
+	if len(name) <= lambdaFunctionFileNameMaxSize {
+		return name, nil
+	}
+	functionNameLength := lambdaFunctionFileNameMaxSize - len(appID) - len(version) - 2
+	hash := sha256.Sum256([]byte(name))
+	hashString := hex.EncodeToString(hash[:])
+	if len(hashString) > functionNameLength {
+		hashString = hashString[:functionNameLength]
+	}
+	name = fmt.Sprintf("%s-%s-%s", appID, version, hashString)
+	return name, nil
+}
+
+// getManifestFileName generates key for a specific manifest in S3,
+// key can be 1024 characters long.
+func getManifestFileKey(appID apps.AppID, version apps.AppVersion) string {
+	return fmt.Sprintf("manifests/%s_%s_manifest", appID, version)
+}
+
+// getAssetFileKey generates key for a specific asset in S3,
+// key can be 1024 characters long.
+func getAssetFileKey(appID apps.AppID, version apps.AppVersion, name string) string {
+	return fmt.Sprintf("%s%s_%s_app/%s", staticAssetsFolder, appID, version, name)
+}
--- a/server/api/impl/aws/aws_keys.go
+++ b/server/api/impl/aws/aws_keys.go
@@ -39,7 +39,7 @@ func getFunctionName(appID apps.AppID, version apps.AppVersion, function string)
 // getManifestFileKey generates key for a specific manifest in S3,
 // key can be 1024 characters long.
 func getManifestFileKey(appID apps.AppID, version apps.AppVersion) string {
-	return fmt.Sprintf("manifests/%s_%s", appID, version)
+	return fmt.Sprintf("manifests/%s_%s.json", appID, version)
 }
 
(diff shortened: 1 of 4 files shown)


server/api/impl/aws/provision_app.go
```diff
@@ -103,7 +122,20 @@ func (c *Client) ProvisionApp(releaseURL string) error {
 			}
 		}
 	}
-	return c.provisionApp(mani.AppID, mani.Version, resFunctions)
+
+	newManifest, err := c.provisionAssets(&mani, assets)
+	if err != nil {
+		return errors.Wrapf(err, "can't provision assets of the app - %s", mani.AppID)
```
catalintomai, line 128: might be helpful to add an error wrapper across the PP (similar to Lev's approach in https://github.com/mattermost/mattermost-webapp/pull/7426/commits/15d9c853c6aea0b7097aa78a10a1eeb170fc7d4a
 - see `asError/errorMessage()`
 to consolidate the capture of `appId` (and potentially other metadata)

server/api/impl/aws/aws_keys.go
```diff
@@ -0,0 +1,49 @@
...
+	hash := sha256.Sum256([]byte(name))
+	hashString := hex.EncodeToString(hash[:])
+	if len(hashString) > functionNameLength {
+		hashString = hashString[:functionNameLength]
+	}
+	name = fmt.Sprintf("%s-%s-%s", appID, version, hashString)
+	return name, nil
+}
+
+// getManifestFileName generates key for a specific manifest in S3,
```
catalintomai, line 39: getManifestFileKey

server/api/impl/aws/aws_keys.go
```diff
@@ -0,0 +1,49 @@
...
+	"encoding/hex"
+	"fmt"
+
+	"github.com/pkg/errors"
+
+	"github.com/mattermost/mattermost-plugin-apps/apps"
+)
+
+// getFunctionName generates function name for a specific app
+// name can be 64 characters long.
```
catalintomai, line 17: do we need to ensure (after all the steps below to generate the final name) that it is or up to 64 chars long?

server/api/impl/aws/provision_app.go
hanzei: 1/5 that the caller should add metadata like this that is tied to a function call.

server/api/impl/aws/aws_keys.go
```diff
@@ -0,0 +1,49 @@
...
+		hashString = hashString[:functionNameLength]
+	}
+	name = fmt.Sprintf("%s-%s-%s", appID, version, hashString)
+	return name, nil
+}
+
+// getManifestFileName generates key for a specific manifest in S3,
+// key can be 1024 characters long.
+func getManifestFileKey(appID apps.AppID, version apps.AppVersion) string {
+	return fmt.Sprintf("manifests/%s_%s_manifest", appID, version)
```
hanzei, line 42: 0/5 having `manifest` multiple times seems redundant. Should we go with
```suggestion
	return fmt.Sprintf("manifests/%s_%s", appID, version)
```
?

hanzei (changes requested):
Is there a ticket we can link here?

hanzei:
Talked with @iomodo offline: This PR should go into `master`. He will rebase it.

iomodo:
@hanzei My mistake no need to rebase here. This PR goes to `MM-31782-static-asset-proxy` and it goes to `master`.

server/api/impl/aws/provision_app.go
iomodo: If we had code with an `app` context I'd agree. But it seems like there is not much. Just an install, provision part. And they all get `app` as an argument. So `appID` is not a state.

server/api/impl/aws/aws_keys.go
iomodo: It is ensured in the code. Final name will be less than 64 chars.

server/api/impl/aws/provision_app.go
```diff
@@ -46,20 +53,21 @@ type functionInstallData struct {
 //      |-- __pycache__
 //      |-- certifi/
 func (c *Client) ProvisionApp(releaseURL string) error {
-	zipFile, err := downloadFile(releaseURL)
-	if err != nil {
-		return errors.Wrapf(err, "can't install app from url %s", releaseURL)
+	zipFile, zipErr := downloadFile(releaseURL)
```
hanzei, line 56: 2/5 on using the word `bundle` for the zip file
```suggestion
	bundle, bundleErr := downloadFile(releaseURL)
```

server/api/impl/aws/provision_app.go
```diff
@@ -46,20 +53,21 @@ type functionInstallData struct {
...
+		return errors.Wrapf(zipErr, "can't install app from url %s", releaseURL)
 	}
 	bundleFunctions := []functionInstallData{}
 	var mani apps.Manifest
+	assets := []assetData{}
 
 	// Read all the files from zip archive
 	for _, file := range zipReader.File {
-		if strings.HasSuffix(file.Name, "manifest.json") {
+		if strings.HasSuffix(file.Name, "manifest.json") { // nolint:gocritic
```
hanzei, line 70: Should we use a `switch` here?

server/api/impl/aws/provision_app.go
```diff
@@ -187,25 +219,21 @@ func (c *Client) createFunction(zipFile io.Reader, function, handler, runtime, r
...
+		key := getAssetFileKey(manifest.AppID, manifest.Version, asset.name)
+		if err := c.S3FileUpload(key, asset.file); err != nil {
+			return nil, errors.Wrapf(err, "can't provision asset - %s with key - %s", asset.name, key)
+		}
+		manifest.Assets = append(manifest.Assets, apps.Asset{
+			Name:   asset.name,
+			Type:   apps.S3Asset,
+			Bucket: c.appsS3Bucket,
+			Key:    key,
+		})
```
hanzei, line 236: It would be simpler if the manifest would be immutable and we won't store that information. Can we generate the information on the fly? We only have one asset type and the bucket should be the same for all apps.

hanzei (changes requested):

hanzei:
@iomodo The PR is hard to review for me as `MM-31782-static-asset-proxy` will conflict with `master` once these changes are in. 1/5 that it's simpler to review if the PR goes to `master`.

server/api/impl/aws/provision_app.go
catalintomai: My point was more towards consolidating the collection of metadata for error/logging purposes. Right now, in different places we collect diff. pieces of metadata to provide the best context for an error,  like `appId`, `file name`, `key`, or none. In some cases we can/makes sense to collect one piece of metadata, in others, multiple. I just want us to make sure that we collect the best context for each and do not miss smth. that could help the debugging process later.

server/api/impl/aws/aws_keys.go
catalintomai: prob. would be clearer if we add a final check as opposed to trying to determine that the individual pieces are of the right size, so that when concatenated they amount to less than 64 chars.

server/api/impl/aws/aws_keys.go
```diff
@@ -0,0 +1,49 @@
...
+		hashString = hashString[:functionNameLength]
+	}
+	name = fmt.Sprintf("%s-%s-%s", appID, version, hashString)
+	return name, nil
+}
+
+// getManifestFileKey generates key for a specific manifest in S3,
+// key can be 1024 characters long.
+func getManifestFileKey(appID apps.AppID, version apps.AppVersion) string {
+	return fmt.Sprintf("manifests/%s_%s", appID, version)
```
hanzei, line 42: Should we append `.json` as it makes the file more readable?

iomodo:
Heads up I'll merge this into the feature branch. You can review PR to master.

PR #63, opened 2021-02-11
MM-31122: Make sure all Admin APIs enforce sysadmin permissions

catalintomai:
#### Summary
Make sure all Admin APIs enforce sysadmin permissions

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-31122


levb (approved):

levb:
Is there more to be done?

catalintomai:
> Is there more to be done?

Some check were added in a previous PR. Some of the `Admin` methods don;t think they apply (`UninstallApp`). If you think smth., is missing, please lmk.

larkox:
What kind of permission we want to give the debug commands (if we will keep them on production)? Specially the `debug-install*` should require the same ones as install.

hanzei (commented):
Should we check for the permissions in the command layer instead of checking that further down the application? That would mimic how permission checks are done in the core product.

catalintomai:
> Should we check for the permissions in the command layer instead of checking that further down the application? That would mimic how permission checks are done in the core product.

Looked at how we are doing currently in (some of) the other plugins, like incident-collaboration and Jira, and they look to follow the same placement.

catalintomai:
> What kind of permission we want to give the debug commands (if we will keep them on production)? Specially the `debug-install*` should require the same ones as install.

The check were added in a previous checkin - and they are the same as for the non-debug commands.

hanzei:
> An argument in favor of the current approach could be that some lower-level methods could be invoked from entry points other than commands

Could you please elaborate? Right now we have two entry points: the slash command and the Rest API. Are there other ways to invoke lower level code?

What I like about checking the entry points is that we fail and return early. That reduces the scope of code that gets executed without permission.

catalintomai:
> > An argument in favor of the current approach could be that some lower-level methods could be invoked from entry points other than commands
> 
> Could you please elaborate? Right now we have two entry points: the slash command and the Rest API. Are there other ways to invoke lower level code?
> 
> What I like about checking the entry points is that we fail and return early. That reduces the scope of code that gets executed without permission.

As I mentioned, the existing pattern for the existing plugins (incident-collaboration and Jira, maybe others)  is to do the check at this level (for their API calls) - seems to be the prevalent one for the server codebase as well. For the apps plugin we introduced a new pattern with the `checkAdmin` (for restapi - incomplete check) and then later with `checkSystemAdmin` (for commands). I think there are pros/cons to each - gating at entry points has the potential to introduce these "each entry point doing his thing" - not necessarily a bad thing; OTOH - one could say that what we want to protect is the actual logic, which should be invoked/shared by the the entry-points.

levb:
Closed in favor of #107

PR #65, opened 2021-02-15
[MM-32862] Sanitized app version and id for function name

hanzei:
#### Summary
AWS only allows letters, number, underscores and hyphens as function names.

Given that the app id and version are used to generate the function name, we need to limit the characters used in them to only the ones stated above. A version is likely contains a dots, hence I think it's fine to sanitized it before generating the function name.

I've also added an `IsValid` method to `AppID` and `AppVersion` to state that. They should get called in the provision process, but the seemed out of scope for this PR.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-32862

### DIFF
--- a/apps/app.go
+++ b/apps/app.go
@@ -1,17 +1,69 @@
 package apps
 
-import "encoding/json"
+import (
+	"encoding/json"
+	"unicode"
 
+	"github.com/pkg/errors"
+)
+
+// AppID is a globally unique identifier that represents a Mattermost App.
+// Allowed characters are letters, numbers, underscores and hyphens.
 type AppID string
-type AppType string
+
+func (id AppID) IsValid() error {
+	for _, c := range id {
+		if unicode.IsLetter(c) {
+			continue
+		}
+
+		if unicode.IsNumber(c) {
+			continue
+		}
+
+		if c == '-' || c == '_' {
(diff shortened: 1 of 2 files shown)


iomodo (approved):

levb (changes requested):
I am so confused... the names of the functions are something that *we* create when we create them, so why do we need methods that validate or correct them? We just need to generate valid names when we create...

Also, I am still 1/5 that content-addressed names are better than the semantic names anyway. The functions' SHA are unique and that's all that matters.

hanzei:
> I am so confused... the names of the functions are something that we create when we create them, so why do we need methods that validate or correct them? We just need to generate valid names when we create...

We create the function names of out of the app's name and it's version. These two input might contain characters that are invalid for a function name. Hence, we need to sanitize. There is no need to validate the function name after we created it. Does that make the change more clear, @levb?

> Also, I am still 1/5 that content-addressed names are better than the semantic names anyway. The functions' SHA are unique and that's all that matters.

While I agree that content-addressed names are better, I don't have a good idea what we want to use there. I assume you are talking about the SHA of the function zip file, correct? In that case it's not reproducible, which might cause issues.

levb (approved):

PR #66, opened 2021-02-15
Fix Install flow

iomodo:
#### Summary
This PR will fix install flow of the apps


server/command/install.go
```diff
@@ -40,7 +40,36 @@ func (s *service) executeInstall(params *params) (*model.CommandResponse, error)
...
+	}
+
+	if !s.api.Mattermost.User.HasPermissionTo(params.commandArgs.UserId, model.PERMISSION_MANAGE_SYSTEM) {
+		return errorOut(params, errors.New("forbidden"))
+	}
+
+	app, err := s.api.Admin.GetApp(apps.AppID(appID))
+	if err != nil {
+		return errorOut(params, errors.Wrap(err, "App not found"))
+	}
```
hanzei, line 68: Should we check the `AppType` just to be sure?

server/api/impl/store/apps.go
```diff
@@ -57,7 +57,7 @@ func (s AppStore) Save(app *apps.App) error {
 	app.AppID = app.Manifest.AppID
 	app.Manifest = nil
 
-	conf.Apps[app.Manifest.AppID] = app.ConfigMap()
+	conf.Apps[app.AppID] = app.ConfigMap()
```
hanzei, line 60: Not strictly related to this PR: Why do we store the `AppID` both in `App` and in `App.Manfest`. Are there Apps without a `Manifest`?

hanzei (commented):
Two questions from my side

server/api/impl/store/apps.go
iomodo: Good question. Apps are stored in the config. Manifests are stored in memory. We need appID to "connect" them. Manifests are not stored in config because of the size.
What that means:
Synchronization of the `apps.json` is run in every done of the HA cluster, so manifests as well as apps are synced.
But whenever we install a non-`apps.json` app from a slash command, manifest will be saved only on one node. And in the HA environment it will not work. I suppose that path is for testing and developer story only anyway and should not be a big problem.
cc @levb

hanzei: First of all, thanks for the explanation :+1: 

Can we rely for non-`apps.json` apps i.e. HTTP apps (correct?) to fetch that data from the app ifself via HTTP and then cache in memory?

levb (approved):
fixes our immediate problems

server/command/install.go
```diff
@@ -40,7 +40,36 @@ func (s *service) executeInstall(params *params) (*model.CommandResponse, error)
 	if err != nil {
 		return errorOut(params, err)
 	}
+	return s.installApp(manifest, appSecret, force, params)
+}
+
+func (s *service) executeInstallAWSApp(params *params) (*model.CommandResponse, error) {
```
catalintomai, line 46: Is there anything AWS-specific invoked here?

server/command/install.go
```diff
@@ -40,7 +40,36 @@ func (s *service) executeInstall(params *params) (*model.CommandResponse, error)
...
+	appID := ""
+	appSecret := ""
+	force := false
+
+	fs := pflag.NewFlagSet("", pflag.ContinueOnError)
+	fs.StringVar(&appID, "app-id", "", "ID of the app")
+	fs.StringVar(&appSecret, "app-secret", "", "App secret")
+	fs.BoolVar(&force, "force", false, "Force re-provisioning of the app")
```
catalintomai, line 54: rephrase "re-provisioning" ?

server/command/main.go
```diff
@@ -25,7 +25,8 @@ func (s *service) handleMain(in *params) (*model.CommandResponse, error) {
 		"debug-install-builtin": s.executeDebugInstallBuiltinHello,
 		"debug-install-http":    s.executeDebugInstallHTTPHello,
 		"debug-install-aws":     s.executeDebugInstallAWSHello,
-		"provision":             s.executeProvision,
+		"provision-aws":         s.executeProvisionAWSApp,
+		"install-aws":           s.executeInstallAWSApp,
```
catalintomai, line 29: Whenever we add a command/other entry point, it would be helpful to document (README, some other place) info like:
- what does it do, who is supposed to run it, context needed, constraints, usage scenarios, etc.

server/api/impl/store/apps.go
```diff
@@ -71,7 +71,7 @@ func (s AppStore) Save(app *apps.App) error {
 
 func (s AppStore) Delete(app *apps.App) error {
 	conf := s.conf.GetConfig()
-	delete(conf.Apps, app.Manifest.AppID)
+	delete(conf.Apps, app.AppID)
```
larkox, line 74: Should we delete also the manifest? Or is it handled somewhere else?

server/command/install.go
```diff
@@ -61,7 +90,7 @@ func (s *service) executeInstall(params *params) (*model.CommandResponse, error)
 	// Finish the installation when the Dialog is submitted, see
 	// <plugin>/http/dialog/install.go
 	err = s.api.Mattermost.Frontend.OpenInteractiveDialog(
-		dialog.NewInstallAppDialog(manifest, appSecret, conf.PluginURL, params.commandArgs))
+		dialog.NewInstallAppDialog(app.Manifest, app.Secret, conf.PluginURL, params.commandArgs))
```
larkox, line 93: app.Manifest or manifest? Is app.Manifest defined in this context?

larkox: Same consideration on line 105.

larkox (approved):
In general looks OK, but there are some minor questions.

server/api/impl/store/apps.go
iomodo: We can, but that's not implemented yet. And we have to do that on every HA node.

server/command/install.go
iomodo: Correct there is nothing else aws specific. We install with `app-id` or with the manifest URL.

server/command/install.go
iomodo: Yes, `ProvisionApp` returns the app with all the context.

server/command/install.go
catalintomai: Then prob. we should change the name of the method from `executeInstall**AWS**App`

README.md
```diff
@@ -45,6 +45,11 @@ Change your directory back to `mattermost-plugin-apps` and run the end to end te
 make test-e2e
 ```
 
+## Usage
+Use `apps provision-aws url` to provision an app in AWS. Pass the url of the archived app.
```
hanzei, line 49: Let's remove that for now as https://github.com/mattermost/mattermost-plugin-apps/pull/67 will introduce a seperate tool

server/command/install.go
hanzei: Should we also rename `executeInstall` to `executeInstallHTTPApp` then and also change the slash command name?

hanzei (changes requested):

DHaussermann:
@iomodo just a heads up I tried to use this build to install the Service Now app and I see the follow errors when enabling the plugin and when trying to install the app.  
```
ERRO[2021-02-19T09:37:50.429322-05:00] This should not have happened. No manifest available for  app_id= caller="mlog/sugar.go:25" plugin_id=com.mattermost.apps
```
```
ERRO[2021-02-19T09:37:50.4413528-05:00] Plugin failed to ServeHTTP, RPC call failed   caller="mlog/log.go:232" error="unexpected EOF" plugin_id=com.mattermost.apps
```

```
ERRO[2021-02-19T09:38:15.969943-05:00] Plugin for /apps is not working. Please contact your system administrator  caller="mlog/log.go:232" err_details="err= Plugin has recently crashed: com.mattermost.apps" err_where=ExecutePluginCommand http_code=500 ip_addr=174.94.35.166 method=POST path=/api/v4/commands/execute request_id=ff5sowbo73gkdpu6dzdczo7zna user_id=gfdrnf97j3f99gie1ywyq8rdow
```

Maybe this is a separate issue but I just wanted to point this out. Maybe I should try again when this is merged to master?

server/api/impl/admin/list.go
```diff
@@ -25,3 +25,7 @@ func (adm *Admin) ListApps() ([]*apps.App, md.MD, error) {
 
 	return apps, out, nil
 }
+
+func (adm *Admin) GetApp(appID apps.AppID) (*apps.App, error) {
+	return adm.store.App().Get(appID)
```
catalintomai, line 30: To add an (admin) permissions check?

server/command/main.go
```diff
@@ -25,10 +25,11 @@ func (s *service) handleMain(in *params) (*model.CommandResponse, error) {
...
 		"info":                  s.executeInfo,
 		"list":                  s.executeList,
-		"install":               s.executeInstall,
+		"install":               s.executeInstallHTTPApp,
```
hanzei, line 32: 3/5 that an the standard case should be install aws app. Should we even combine these two commands into a single `/install` command that either takes an `appID` of a listed AWS app `/install zendesk` or an URL `/install -url http://localhost:4000/manifest.json`?

server/api/impl/admin/provision.go
```diff
@@ -45,6 +45,7 @@ func (adm *Admin) ProvisionApp(cc *apps.Context, sessionToken apps.SessionToken,
 	if err != nil {
 		return nil, "", err
 	}
+	adm.store.Manifest().Save(manifest)
```
hanzei, line 48: It not obvious for the caller that the manifest also has be be saved in the store. `App().Save` should handle that.

server/command/install.go
```diff
@@ -40,7 +40,39 @@ func (s *service) executeInstall(params *params) (*model.CommandResponse, error)
 	if err != nil {
 		return errorOut(params, err)
 	}
```
hanzei, line 42: Should we check the App type here too?

hanzei (changes requested):
I'm wondering if we should change the slash command pattern. See my comment below

server/api/impl/admin/provision.go
iomodo: You are correct, but right now appStore and manifestStore are handled differently and this change will break [this PR](https://github.com/mattermost/mattermost-plugin-apps/pull/78) I'll refactor after both PRs are merged

hanzei: Could you please file a ticket for that so don't forget that?

hanzei (approved):
LGTM :+1:

PR #67, opened 2021-02-17
[MM-32702] Introduce appsctl

hanzei:
#### Summary
This PR introduces `appsctl` as a tool to manage and provision apps.

The important command is `appsctl provision app` which takes an apps bundle as input and provisions it to AWS. For development purposes there is an `--update` flag that updates the function. In production, we won't be using that, but instead release a new version.

I've added some documentation to the readme.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-32702

### DIFF
--- a/assets/apps.json
+++ b/assets/apps.json
@@ -1,7 +1,6 @@
 {
     "apps" : {
-        "zendesk" : "0.0.1"
-    },
-    "overrides" : {
+        "zendesk" : "0.0.1",
+        "com.mattermost.servicenow" : "0.1.0"
     }
 }
--- a/assets/apps.json
+++ b/assets/apps.json
@@ -1,11 +1,7 @@
 {
     "apps" : {
-        "jira" : "v3.0.0",
-        "todo" : "v2.2.3"
+        "zendesk" : "0.0.1"
     },
     "overrides" : {
-        "my_installation_id" : {
-            "jira" : "v.3.0.1"
-        }
     }
 }
--- a/server/api/impl/aws/provision_app.go
+++ b/server/api/impl/aws/provision_app.go
@@ -256,9 +256,13 @@ func (c *Client) createFunction(zipFile io.Reader, function, handler, runtime, r
 		if _, ok := err.(*lambda.ResourceConflictException); !ok {
 			return errors.Wrapf(err, "Can't create function. Response: %v\n", result)
 		}
+
+		c.logger.Info(fmt.Sprintf("funcion %s already exists. Not updating it.", function))
+
+		return nil
 	}
 
-	c.logger.Info(fmt.Sprintf("function named %s was created with response: %v", function, result))
+	c.logger.Info(fmt.Sprintf("function %s was created with response: %v", function, result))
 
 	return nil
 }
--- a/server/api/impl/aws/provision_app.go
+++ b/server/api/impl/aws/provision_app.go
@@ -13,6 +13,7 @@ import (
 	"io"
 	"io/ioutil"
 	"net/http"
+	"os"
 	"strings"
 
 	"github.com/aws/aws-sdk-go/service/lambda"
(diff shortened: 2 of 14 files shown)


assets/apps.json
```diff
@@ -1,11 +1,7 @@
 {
     "apps" : {
-        "jira" : "v3.0.0",
-        "todo" : "v2.2.3"
+        "zendesk" : "0.0.1"
```
hanzei, line 3: I've update the file to only contain "valid" data as zendesk is the only app that runs in AWS right now.

server/api/impl/aws/provision_app.go
```diff
@@ -45,21 +46,42 @@ type functionInstallData struct {
...
+
+	b, err := ioutil.ReadAll(f)
+	if err != nil {
+		return errors.Wrap(err, "can't read file")
+	}
+
+	return c.ProvisionApp(b, shouldUpdate)
+}
+
+func (c *Client) ProvisionApp(b []byte, shouldUpdate bool) error {
```
hanzei, line 72: I'm aware that it's bad practice to pass a byte slice, but it seemed fine to me given that the code will need a refactoring anyway.

server/api/impl/aws/provision_app.go
```diff
@@ -85,8 +107,16 @@ func (c *Client) ProvisionApp(releaseURL string) error {
...
 			})
+			c.logger.Debug("Found function bundle", "file", file.Name)
+		default:
+			c.logger.Info("Unknown file found in app bundle", "file", file.Name)
 		}
 	}
+
+	if mani == nil {
+		return errors.New("no manifest found")
+	}
```
hanzei, line 118: Helps with debugging malformed bundles

server/api/impl/aws/s3.go
```diff
@@ -27,12 +27,41 @@ func (c *Client) S3FileDownload(bucket, item string) ([]byte, error) {
...
+		return errors.Wrap(err, "failed to create bucket")
+	}
+
+	return nil
+}
+
+// CheckIfBucketExists return true if a bucket with the given name exists.
+// Otherwise it returns false.
+func (c *Client) CheckIfBucketExists(name string) (bool, error) {
+	buckets, err := c.Service().s3.ListBuckets(&s3.ListBucketsInput{})
```
hanzei, line 45: There doesn't seem to exist a `getBucket` function. `ListBuckets` is also what is used in the docs https://docs.aws.amazon.com/sdk-for-ruby/v3/developer-guide/s3-example-does-bucket-exist.html.

server/api/impl/aws/service.go
```diff
@@ -89,8 +92,9 @@ func createAWSConfig(awsAccessKeyID, awsSecretAccessKey string) *aws.Config {
 // NewService creates a new instance of Service.
 func NewService(sess *session.Session) *Service {
 	return &Service{
-		lambda:       lambda.New(sess, aws.NewConfig().WithLogLevel(aws.LogDebugWithRequestErrors)),
```
hanzei, line 92: The debug logs are _way_ to noisy. `NewAWSSessionWithLogger` already ensure HTTP requests are logged on `Debug`. I don't think we should rely on the AWS logger and hence I removed this.

server/api/impl/aws/provision_app.go
```diff
@@ -146,10 +176,55 @@ func (c *Client) provisionApp(appID apps.AppID, appVersion apps.AppVersion, func
...
+			if err := c.createFunction(function.zipFile, name, function.handler, function.runtime, policyName); err != nil {
+				return errors.Wrapf(err, "can't install function for %s", appID)
+			}
 		}
 	}
+
+	return nil
+}
+
+func (c *Client) createOrUpdateFunction(zipFile io.Reader, function, handler, runtime, resource string) error {
```
levb, line 194: Why would we ever update a function? Dev-mode?

levb (approved):
This PR should also remove provision.go and iam.go from `server/.../aws`, no?

server/api/impl/aws/provision_app.go
hanzei: Yes, updating a function should only happen in dev-mode. Never in production.

hanzei:
> This PR should also remove provision.go and iam.go from `server/.../aws`, no?

Could you please elaborate a bit on that proposal? Both files are required to provision an app. If we move them out of there, where would they do? Into `appsctl`? Would we remove the slash command then?

levb:
> Would we remove the slash command then

That's my thinking. The provisioning `/-command` was meant to be for dev mode only, I don't see why we wouldn't just recommend using `appctl` instead.

The `/apps install` should not be affected, as per our other discussions its only input is the manifest itself, whether coming from the "global list", or provided (as a URL? as a file?) by the developer.

hanzei:
@levb I'm fine with removing the provision slash command. Let me sync with @iomodo on that.

Yes, `/apps install should accept either an id or a manifest via an URL.

iomodo (approved):
LGTM!
We should probably add the provision from URL option as well from the command. I'd assume our production bundles will be in the github or in S3, correct?

hanzei:
@iomodo Good question. I leaved the provision from URL code as it is because we might need to later. But I think it's to early to add an URL option because we don't know yet how the release process looks exactly.

PR #70, opened 2021-02-17
[MM-32910] Fix app store

hanzei:
#### Summary
`gob` can't deal with a `map[apps.AppID]interface{}` as it's an unknown type to it. We either need to use a `[]inferface{}` or `map[string]interface{}`. I went with the latter as there are some code areas  that make use of the `map` and I wanted to keep the diff simple. Open for other suggestions.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-32910

### DIFF
--- a/server/api/config.go
+++ b/server/api/config.go
@@ -2,14 +2,12 @@ package api
 
 import (
 	"github.com/mattermost/mattermost-server/v5/model"
-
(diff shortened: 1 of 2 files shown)


levb (approved):

levb:
Oops, sorry hit merge by accident, @iomodo if you disagree, please revert or submit separate fix(es)

PR #72, opened 2021-02-19
[MM-33080] Set status after installation

hanzei:
#### Summary
I've checked the code base and this seems like the only place that was missing the status.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-33080

### DIFF
--- a/server/api/impl/admin/provision.go
+++ b/server/api/impl/admin/provision.go
@@ -36,6 +36,7 @@ func (adm *Admin) ProvisionApp(cc *apps.Context, sessionToken apps.SessionToken,
 
 	app := &apps.App{
 		Manifest:       manifest,
+		Status:         apps.AppStatusInstalled,
 		BotUserID:      bot.UserId,
 		BotUsername:    bot.Username,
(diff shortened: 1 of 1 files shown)


server/api/impl/admin/provision.go
```diff
@@ -36,6 +36,7 @@ func (adm *Admin) ProvisionApp(cc *apps.Context, sessionToken apps.SessionToken,
 
 	app := &apps.App{
 		Manifest:       manifest,
+		Status:         apps.AppStatusInstalled,
```
catalintomai, line 39: Why do we set the `AppStatusInstalled` on the `ProvisionApp` call (as opposed to the `InstallApp` call)?

hanzei: Good point. Maybe I've misunderstood what `ProvisionApp` does. Let me double check and come back to you.

levb:
https://github.com/mattermost/mattermost-plugin-apps/pull/85 makes this PR redundant

hanzei:
Yep, let's close this one.

PR #74, opened 2021-02-19
[MM-33083] Use HTTP request like payload to invoke lambda

hanzei:
#### Summary
Because the Apps expect HTTP-like requests we have to adopt the payload that is used to invoke the lambda function. The format is documented [here](https://docs.aws.amazon.com/apigateway/latest/developerguide/http-api-develop-integrations-lambda.html). I didn't fill out all fields given that we only need a handful.

There is some double JSON Unmarshaling in the code that I couldn't prevent. Maybe we need to optimize that later.

Given that there is no per path mapping to function right now, all calls go to the first defined function. Hence, the PR kind of drops support for more then one function.

I've tested the PR with the Zendesk app and can confirm that it works.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-33083

### DIFF
--- a/server/api/impl/upstream/upawslambda/upstream.go
+++ b/server/api/impl/upstream/upawslambda/upstream.go
@@ -22,6 +22,8 @@ type Upstream struct {
 	aws *aws.Client
 }
 
+// invocationPayload is a scoped down version of
+// https://pkg.go.dev/github.com/aws/aws-lambda-go@v1.13.3/events#APIGatewayProxyRequest
 type invocationPayload struct {
 	Path       string            `json:"path"`
 	HTTPMethod string            `json:"httpMethod"`
@@ -29,6 +31,8 @@ type invocationPayload struct {
 	Body       interface{}       `json:"body"`
 }
 
+// invocationResponse is a scoped down version of
+// https://pkg.go.dev/github.com/aws/aws-lambda-go@v1.13.3/events#APIGatewayProxyResponse
 type invocationResponse struct {
 	StatusCode int    `json:"statusCode"`
 	Body       string `json:"body"`
--- a/server/api/impl/upstream/upawslambda/upstream.go
+++ b/server/api/impl/upstream/upawslambda/upstream.go
@@ -4,11 +4,14 @@
 package upawslambda
 
 import (
-	"bytes"
+	"encoding/json"
 	"io"
 	"io/ioutil"
+	"net/http"
+	"strings"
 
 	"github.com/aws/aws-sdk-go/service/lambda"
+	"github.com/pkg/errors"
 
 	"github.com/mattermost/mattermost-plugin-apps/apps"
 	"github.com/mattermost/mattermost-plugin-apps/server/api/impl/aws"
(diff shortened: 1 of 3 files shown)


levb (approved):

levb:
Thank you, this should make the app developer's job easier.

server/api/impl/upstream/upawslambda/upstream.go
```diff
@@ -27,14 +42,52 @@ func NewUpstream(app *apps.App, aws *aws.Client) *Upstream {
 }
 
 func (u *Upstream) OneWay(call *apps.Call) error {
-	_, err := u.aws.InvokeLambda(u.app.Manifest.AppID, u.app.Manifest.Version, call.URL, lambda.InvocationTypeEvent, call)
+	payload, err := CallToInvocationPayload(call)
+	if err != nil {
+		return errors.Wrap(err, "failed to covert call into invocation payload")
+	}
+
+	_, err = u.aws.InvokeLambda(u.app.Manifest.AppID, u.app.Manifest.Version, u.app.Manifest.Functions[0].Name, lambda.InvocationTypeEvent, payload)
```
iomodo, line 50: why do we invoke `u.app.Manifest.Functions[0].Name`?

server/api/impl/upstream/upawslambda/upstream.go
```diff
@@ -27,14 +42,52 @@ func NewUpstream(app *apps.App, aws *aws.Client) *Upstream {
...
+
+	if resp.StatusCode != http.StatusOK {
+		return nil, errors.Errorf("lambda invocation failed with status code %v and body %v", resp.StatusCode, resp.Body)
+	}
+
+	return ioutil.NopCloser(strings.NewReader(resp.Body)), nil
+}
+
+func CallToInvocationPayload(call *apps.Call) ([]byte, error) {
+	request := invocationPayload{
```
iomodo, line 80: What is the reason of having the HTTP like payload? Matching with the HTTP apps?

server/api/impl/upstream/upawslambda/upstream.go
hanzei: What mechanism do we have to decide which function to invoke?

iomodo: It's in the `call`

hanzei: Maybe I'm misunderstanding. What field in `Call` can be used to to decide which function to invoke? A Call might only have a `Path` which we can only use for routing if a function also have something like a `Path`.

server/api/impl/upstream/upawslambda/upstream.go
hanzei: Apps expect an HTTP like payload given that they are essentially HTTP servers. That approach makes the transition from an HTTP app to an AWS app seamless. Given that we don't place an AWS Gateway in front of the lambda we have to generate the payload ourself.

levb: I 👍  @hanzei's view here. This approach would also make a transition to a self-hosted (lambda) easier on the developer, should they choose to do so.

iomodo (commented):
My only concern with this PR is `u.app.Manifest.Functions[0].Name`
To map `path` to the functions name, we could assume that all path formats are `func_name/somepath...` @hanzei thoughts?

hanzei:
@iomodo I don't think that assumption is valid. E.g. `/install` is the default path which would ever be valid then. We've decided offline that we want multi function apps, but I'm hesitant to add support for that in this PR. @levb is working on that anyway.

iomodo:
This PR will work only for a single function. I'm OK merging as is. We can track the multi-function effort in the separate PR

iomodo (approved):

PR #75, opened 2021-02-19
[MM-30046] Add /uninstall command

hanzei:
#### Summary
This PR adds an `/uninstall` slash command. I've briefly tested it locally, but given some bugs in the code I wasn't able to fully test.

Still, the slash command is a valuable tool for debugging.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-30046

### DIFF
--- a/server/api/impl/admin/synchronize.go
+++ b/server/api/impl/admin/synchronize.go
@@ -87,7 +87,7 @@ func (adm *Admin) LoadAppsList() error {
 			adm.mm.Log.Error("can't load manifest from store", "app_id", registeredApp.Manifest.AppID)
 			continue
 		}
-		if err := adm.UninstallApp(registeredApp); err != nil {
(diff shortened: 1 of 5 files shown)


server/api/impl/admin/synchronize.go
```diff
@@ -137,7 +137,12 @@ func (adm *Admin) LoadAppsList() error {
 	return nil
 }
 
-func (adm *Admin) UninstallApp(app *apps.App) error {
+func (adm *Admin) UninstallApp(appID apps.AppID) error {
+	app, err := adm.store.App().Get(appID)
```
catalintomai, line 141: To add (admin) permission check?

hanzei: @catalintomai I've added a permission check check in 94e7ab16c6927ca8124b5de876d7a0a62552cc34. Please let me know what you think about that implementation.

levb (approved):

catalintomai (approved):

PR #78, opened 2021-02-23
[MM-33177] Simplify app sync

iomodo:
#### Summary
This PR simplifies app sync

#### Ticket Link
  Fixes https://mattermost.atlassian.net/browse/MM-33177

### DIFF
--- a/assets/apps.json
+++ b/assets/apps.json
@@ -1,11 +1,6 @@
 {
-    "apps" : {
-        "jira" : "v3.0.0",
-        "todo" : "v2.2.3"
-    },
-    "overrides" : {
-        "my_installation_id" : {
-            "jira" : "v.3.0.1"
-        }
+    "apps": {
+        "jira": "v3.0.0",
+        "todo": "v2.2.3"
     }
 }
--- a/server/api/impl/admin/synchronize.go
+++ b/server/api/impl/admin/synchronize.go
@@ -25,12 +25,12 @@ type AppVersions struct {
 	Apps apps.AppVersionMap `json:"apps"`
 }
 
-type appNewVersion struct {
+type appOldVersion struct {
 	oldApp     *apps.App
 	newVersion apps.AppVersion
 }
 
-type appOldVersion struct {
+type appNewVersion struct {
 	newApp     *apps.App
 	oldVersion apps.AppVersion
 }
@@ -50,7 +50,7 @@ func (adm *Admin) LoadAppsList() error {
 	// Here are apps that are provisioned and registered on this installation.
 	registeredApps := adm.store.App().GetAll()
 
-	adm.populateManifests(appsForRegistration)
+	adm.getAndStoreManifests(appsForRegistration)
 
 	appsToRegister, appsToUpgrade, appsToRemove := mergeApps(appsForRegistration, registeredApps)
 	adm.removeApps(appsToRemove)
@@ -60,15 +60,15 @@ func (adm *Admin) LoadAppsList() error {
 	registeredAppsUpgraded := adm.store.App().GetAll()
 
 	// call onInstanceStartup. App migration happens here
-	for _, registeredApp := range registeredAppsUpgraded {
-		if registeredApp.Status == apps.AppStatusInstalled {
-			upgradedApp, ok := upgradedApps[registeredApp.AppID]
+	for _, registeredAppUpgraded := range registeredAppsUpgraded {
+		if registeredAppUpgraded.Status == apps.AppStatusInstalled {
+			upgradedApp, ok := upgradedApps[registeredAppUpgraded.AppID]
 			values := map[string]string{}
 			if ok {
 				// app was upgraded send the message
 				values[oldVersionKey] = string(upgradedApp.oldVersion)
 			}
-			adm.callOnStartupOnceWithValues(registeredApp, values)
+			adm.callOnStartupOnceWithValues(registeredAppUpgraded, values)
 		}
 	}
 
@@ -126,7 +126,7 @@ func getAppsForRegistration(path string) (apps.AppVersionMap, error) {
 	return apps.Apps, nil
 }
 
-func (adm *Admin) populateManifests(appVersions apps.AppVersionMap) {
+func (adm *Admin) getAndStoreManifests(appVersions apps.AppVersionMap) {
 	adm.store.Manifest().Cleanup()
 
 	for id, version := range appVersions {
@@ -144,12 +144,12 @@ func (adm *Admin) populateManifests(appVersions apps.AppVersionMap) {
 	}
 }
 
-// mergeApp merges too files and detects apps that should upgrade, register or remove
+// mergeApp merges two pile of apps and detects apps that should upgrade, register or remove
 // appsForRegistration are apps from the apps.json file
 // registeredApps are apps in the config already registered.
-func mergeApps(appsForRegistration apps.AppVersionMap, registeredApps []*apps.App) (apps.AppVersionMap, map[apps.AppID]appNewVersion, map[apps.AppID]*apps.App) {
+func mergeApps(appsForRegistration apps.AppVersionMap, registeredApps []*apps.App) (apps.AppVersionMap, map[apps.AppID]appOldVersion, map[apps.AppID]*apps.App) {
 	appsToRegister := apps.AppVersionMap{}
-	appsToUpgrade := map[apps.AppID]appNewVersion{}
+	appsToUpgrade := map[apps.AppID]appOldVersion{}
 	appsToRemove := map[apps.AppID]*apps.App{}
 
 	allRegisteredAppsMap := map[apps.AppID]*apps.App{}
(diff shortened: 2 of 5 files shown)


server/api/impl/admin/synchronize.go
```diff
@@ -69,73 +41,34 @@ func (adm *Admin) LoadAppsList() error {
...
-	}
+	appsToRegister, appsToUpgrade, appsToRemove := mergeApps(appsForRegistration, registeredApps)
+	adm.removeApps(appsToRemove)
+	upgradedApps := adm.upgradeApps(appsToUpgrade)
+	adm.registerApps(appsToRegister)
 
 	registeredAppsUpgraded := adm.store.App().GetAll()
 
 	// call onInstanceStartup. App migration happens here
 	for _, registeredApp := range registeredAppsUpgraded {
```
catalintomai, line 63: why not `registeredAppUpgraded`?

server/api/impl/admin/synchronize.go
```diff
@@ -180,7 +113,121 @@ func (adm *Admin) UninstallApp(appID apps.AppID) error {
...
+	}
+	var apps *AppVersions
+	if err := json.Unmarshal(data, &apps); err != nil || apps == nil {
+		return nil, errors.Wrapf(err, "can't unmarshal %s file", appsJSONFile)
+	}
+
+	return apps.Apps, nil
+}
+
+func (adm *Admin) populateManifests(appVersions apps.AppVersionMap) {
```
catalintomai, line 129: s/`populateManifests`/`getAndStoreManifests`? (maybe populate is not the right action here).

server/api/impl/admin/synchronize.go
```diff
@@ -180,7 +113,121 @@ func (adm *Admin) UninstallApp(appID apps.AppID) error {
...
+		case apps.AppStatusRegistered:
+			// delete app from proxy plugin, not removing the data
+			if err := adm.store.App().Delete(appToRemove); err != nil {
+				adm.mm.Log.Error("can't delete app from store", "app_id", appToRemove.AppID, "err", err.Error())
+			}
+		}
+	}
+}
+
+func (adm *Admin) upgradeApps(appsToUpgrade map[apps.AppID]appNewVersion) map[apps.AppID]appOldVersion {
```
catalintomai, line 193: somewhat counter-intuitive that the `upgradeApps` method takes `appNewVersion` objs as input and returns `appOldVersion` objs. I would've expected the opposite.

server/api/impl/admin/synchronize.go
```diff
@@ -180,7 +113,121 @@ func (adm *Admin) UninstallApp(appID apps.AppID) error {
...
+		upgradedApps[upgradedApp.AppID] = appOldVersion{
+			newApp:     upgradedApp,
+			oldVersion: oldVersion,
+		}
+	}
+	return upgradedApps
+}
+
+func (adm *Admin) registerApps(appsToRegister apps.AppVersionMap) {
+	for appID, manifest := range adm.store.Manifest().GetAll() {
```
catalintomai, line 221: maybe add a comment about why we are iterating over manifests instead of `populateManifests`.

server/api/impl/admin/synchronize.go
```diff
@@ -180,7 +113,121 @@ func (adm *Admin) UninstallApp(appID apps.AppID) error {
...
+			adm.mm.Log.Error("failed to get manifest", "app", id, "version", version, "err", err.Error())
+
+			continue
+		}
+
+		adm.store.Manifest().Save(manifest)
+	}
+}
+
+// mergeApp merges too files and detects apps that should upgrade, register or remove
```
catalintomai, line 147: `too` typo.

server/api/impl/admin/synchronize.go
iomodo: Even though this is fewer lines of code, iterating over `appsForRegistration` will be more readable. I'll change that.

server/api/impl/admin/synchronize.go
```diff
@@ -180,7 +113,127 @@ func (adm *Admin) UninstallApp(appID apps.AppID) error {
 	return nil
 }
 
-func (adm *Admin) RegisterApp(manifest *apps.Manifest) error {
+func getAppsForRegistration(path string) (apps.AppVersionMap, error) {
```
hanzei, line 116: What does the `ForRegistration` part of the name mean?

server/api/impl/admin/synchronize.go
```diff
@@ -180,7 +113,127 @@ func (adm *Admin) UninstallApp(appID apps.AppID) error {
...
+
+func (adm *Admin) removeApps(appsToRemove map[apps.AppID]*apps.App) {
+	for _, appToRemove := range appsToRemove {
+		switch appToRemove.Status {
+		case apps.AppStatusInstalled:
+			if err := adm.UninstallApp(appToRemove.AppID); err != nil {
+				adm.mm.Log.Error("can't uninstall app", "app_id", appToRemove.AppID, "err", err.Error())
+			}
+		case apps.AppStatusRegistered:
+			// delete app from proxy plugin, not removing the data
```
hanzei, line 188: The comment seems misleading to me. Given that the apps is only listed, there is no data anyway, right?

server/api/impl/admin/synchronize.go
```diff
@@ -180,7 +113,127 @@ func (adm *Admin) UninstallApp(appID apps.AppID) error {
...
+		}
+		upgradedApps[upgradedApp.AppID] = appNewVersion{
+			newApp:     upgradedApp,
+			oldVersion: oldVersion,
+		}
+	}
+	return upgradedApps
+}
+
+func (adm *Admin) registerApps(appsToRegister apps.AppVersionMap) {
```
hanzei, line 223: Should we use the term `listed App` instead? Register indicates to me that the App did register itself, which is not the case. The term `listed` is also used in the Marketplace for the same kind of Apps as here.

hanzei (changes requested):
The code is in general still hard to read. I don't have any concrete suggestions on how to simplify it. It seems to me as too many things are "done at the same time". Moving the code into a single function might make it more readable as there is less hopping around.

It might help if the code deals with registration first, then deletes the unlisted apps and then does the upgrade.

hanzei:
How does this PR interact with https://github.com/mattermost/mattermost-plugin-apps/pull/85/? Are they both valid?

server/api/impl/admin/synchronize.go
```diff
@@ -69,73 +41,34 @@ func (adm *Admin) LoadAppsList() error {
...
-			// update app in proxy plugin
-			oldVersion := registeredApp.Manifest.Version
-			registeredApp.Manifest = manifest
-			if err := adm.store.App().Save(registeredApp); err != nil {
-				adm.mm.Log.Error("can't save an app", "app_id", registeredApp.Manifest.AppID)
-			}
-			updatedAppVersionsMap[registeredApp.Manifest.AppID] = oldVersion
-		}
-	}
+	adm.getAndStoreManifests(appsForRegistration)
```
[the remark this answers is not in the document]
iomodo: To me decomposition makes code more readable. I don't like bloated functions

server/api/impl/admin/synchronize.go
```diff
@@ -69,73 +41,34 @@ func (adm *Admin) LoadAppsList() error {
...
-
-	appsForInstallation, err := getAppsForInstallation(appsPath, installationID)
+	appsForRegistration, err := getAppsForRegistration(appsPath)
 	if err != nil {
 		return errors.Wrap(err, "can't get apps for installation")
 	}
 
-	adm.populateManifests(appsForInstallation)
-
+	// Here are apps that are provisioned and registered on this installation.
```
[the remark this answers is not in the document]
iomodo: Nope `provisioned` is the right term to use here. `provisioned and registered` apps might not be installed yet.

server/api/impl/admin/synchronize.go
iomodo: App that are "listed" in the marketplace are called "registered" there status is also "registered". We also have "installed" and "uninstalled" apps

server/api/impl/admin/synchronize.go
iomodo: correct

server/api/impl/admin/synchronize.go
iomodo: I'm 0/5 on naming. IT was called `listed` in my previous PR but we changed it to `registered`.

server/api/impl/admin/synchronize.go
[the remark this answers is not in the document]
hanzei: Pushing that discussion to https://github.com/mattermost/mattermost-plugin-apps/pull/85

server/api/impl/admin/synchronize.go
hanzei: Moving to https://github.com/mattermost/mattermost-plugin-apps/pull/85

hanzei (approved):
This PR is a step in the right direction. Further work will be done in https://github.com/mattermost/mattermost-plugin-apps/pull/85

assets/apps.json
```diff
@@ -1,6 +1,6 @@
 {
-    "apps" : {
-        "zendesk" : "0.0.1",
-        "com.mattermost.servicenow" : "0.1.0"
+    "apps": {
```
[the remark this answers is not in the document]
levb: I'm proposing a replacement to this soon in #85 so maybe leave it as is for now?

levb (approved):

PR #82, opened 2021-02-26
Add user agent to call context

larkox:
#### Summary
Add user agent to call context to allow discriminating between contexts (specially for bindings, but also may be interesting for other calls).

#### Ticket Link
None

#### Related PRs
Redux: https://github.com/mattermost/mattermost-redux/pull/1379
Mobile: https://github.com/mattermost/mattermost-mobile/pull/5193

### DIFF
--- a/server/api/constants.go
+++ b/server/api/constants.go
@@ -45,5 +45,5 @@ const (
 	PropTeamID    = "team_id"
 	PropChannelID = "channel_id"
 	PropPostID    = "post_id"
-	PropUserAgent = "scope"
+	PropUserAgent = "user_agent_type"
 )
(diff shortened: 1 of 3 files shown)


server/api/constants.go
```diff
@@ -45,4 +45,5 @@ const (
 	PropTeamID    = "team_id"
 	PropChannelID = "channel_id"
 	PropPostID    = "post_id"
+	PropUserAgent = "scope"
```
levb, line 48: ???

levb (changes requested):

server/api/constants.go
larkox: So far we have been calling that scope. We can do the change in the webapp/mobile repo and call it something different (like user_agent). Any preference on the naming?

levb: I don't think `scope` is intention-revealing. `user-agent` may have confusing expectations as to the format of the string. 0/5 `user-agent-type`?

levb (approved):

PR #83, opened 2021-02-26
Add refresh websocket event on app install

larkox:
#### Summary
When the app gets installed, send a websocket event to the user that installed the app.
There is no need to send the websocket event to all the users, since their experience will not be affected if they have to navigate to another channel to see the effects of the install (since they are not installing the app, they don't know when the app gets installed).

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-33129

#### Related PRs
Mobile: https://github.com/mattermost/mattermost-mobile/pull/5194
Webapp: https://github.com/mattermost/mattermost-webapp/pull/7599

### DIFF
--- a/server/api/constants.go
+++ b/server/api/constants.go
@@ -40,7 +40,7 @@ const (
 
 	BindingsPath = "/bindings"
 
-	WesocketEventRefreshBindings = "refresh_bindings"
+	WebsocketEventRefreshBindings = "refresh_bindings"
 )
 
 const (
--- a/server/api/constants.go
+++ b/server/api/constants.go
@@ -40,7 +40,7 @@ const (
 
 	BindingsPath = "/bindings"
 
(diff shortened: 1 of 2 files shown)


levb (approved):
We should prob. make a Proxy/REST API for this (separate ticket/PR) otherwise, making it look easy, LGTM.

hanzei (approved):

server/api/constants.go
```diff
@@ -39,6 +39,8 @@ const (
 	UnsubscribePath = "/unsubscribe"
 
 	BindingsPath = "/bindings"
+
+	WesocketEventRefreshBindings = "refresh_bindings"
```
catalintomai, line 43: s/WesocketEventRefreshBindings/We**b**socketEventRefreshBindings ?

PR #84, opened 2021-02-28
(partial?) fix for e2e tests

levb:
Fixed passing the bundle into the e2e tests.

@hanzei would be great if you could add running `make test-e2e` to ci? What happens if we just include it in `make all`?

PR #85, opened 2021-03-01
App/Manifest store and some refactoring

levb:
- Added apps, manifest store based on config and KV
`/apps debug-add-manifest --url=<URL>` to install a manifest as a local "override"
`/apps install --app-id=appid` to install an app from all available manifests
`/apps list to view available and installed apps`

Other:
- restructured `/server/api.impl/aws` (service) and `/awsclient` packages (NOT TESTED)
- consolidated KV prefixes and known keys in constants.go
- got rid of the `api.Service` monstrocity
- added path matching to AWS lambda upstream


.gitignore
```diff
@@ -8,3 +8,8 @@ dist/
 
 # Jetbrains
 .idea/
+
```
levb, line 11: includes https://github.com/mattermost/mattermost-plugin-apps/pull/84 for debugging

Makefile
```diff
@@ -99,7 +99,6 @@ mock:
 ifneq ($(HAS_SERVER),)
 	go install github.com/golang/mock/mockgen
 	mockgen -destination server/api/mock_api/mock_appservices.go github.com/mattermost/mattermost-plugin-apps/server/api AppServices
-	mockgen -destination server/api/mock_api/mock_proxy.go github.com/mattermost/mattermost-plugin-apps/server/api Proxy
```
levb, line 102: seemed unused

apps/app.go
```diff
@@ -98,9 +104,13 @@ const (
 // Function describes app's function mapping
 // For now Function can be either AWS Lambda or HTTP function
 type Function struct {
+	Path    string `json:"path"`
```
levb, line 107: Path is the call path, will be used for routing

apps/app.go
```diff
@@ -113,19 +123,26 @@ type Asset struct {
...
 
+	DisplayName string `json:"display_name,omitempty"`
+	Description string `json:"description,omitempty"`
 	HomepageURL string `json:"homepage_url,omitempty"`
 
-	// HTTPRootURL applicable For AppTypeHTTP.
-	//
-	// TODO: check if it is used in the // user-agent, consider naming
-	// consistently.
+	OnDisable   *Call `json:"on_disable,omitempty"`
```
levb, line 138: On Enable/Disable should come as soon as we add `apps disable/enable` command :)

assets/app-test-zendesk.json
```diff
@@ -0,0 +1,25 @@
+{
```
levb, line 1: not fully functional, an illustration of a manifest bundled in a local file, should probably be updated to something that works

server/api/admin.go
```diff
@@ -6,10 +6,12 @@ import (
...
+	AddLocalManifest(*apps.Context, apps.SessionToken, *apps.Manifest) (md.MD, error)
 	GetApp(appID apps.AppID) (*apps.App, error)
+	GetManifest(appID apps.AppID) (*apps.Manifest, error)
 	InstallApp(*apps.Context, apps.SessionToken, *apps.InInstallApp) (*apps.App, md.MD, error)
+	ListInstalledApps() map[apps.AppID]*apps.App
+	ListMarketplaceApps(filter string) map[apps.AppID]*apps.MarketplaceApp
 	UninstallApp(appID apps.AppID) error
-	ProvisionApp(*apps.Context, apps.SessionToken, *apps.InProvisionApp) (*apps.App, md.MD, error)
-	LoadAppsList() error
+	UpdateInstalledApps() error
```
levb, line 16: `UpdateInstalledApps` is the new `LoadAppList`, others are self-explanatory

server/api/constants.go
```diff
@@ -46,3 +49,13 @@ const (
 	PropChannelID = "channel_id"
 	PropPostID    = "post_id"
 )
+
+// KV namespace
+const (
+	KeyCallOnce     = "CallOnce"
```
larkox, line 55: Not sure what will be finally the keys for apps KV values, but does it make sense to add a prefix to these? Something like "mmapps_"? That way it is easier to notify and check whether an app key is valid (if we form them as appid_key we just have to make sure the app_id != mmapps)

awsclient/client.go
```diff
@@ -0,0 +1,95 @@
...
+	"github.com/aws/aws-sdk-go/service/lambda"
+	"github.com/aws/aws-sdk-go/service/lambda/lambdaiface"
+	"github.com/aws/aws-sdk-go/service/s3"
+	"github.com/aws/aws-sdk-go/service/s3/s3iface"
+	"github.com/aws/aws-sdk-go/service/s3/s3manager"
+	"github.com/aws/aws-sdk-go/service/s3/s3manager/s3manageriface"
+)
+
+// DefaultRegion describes default region in aws
+const DefaultRegion = "us-east-2"
```
catalintomai, line 27: Does this need to change pre-release/should it be passed in?

cmd/appsctl/provision.go
```diff
@@ -54,9 +56,9 @@ var provisionBucketCmd = &cobra.Command{
 			return err
 		}
 
-		name := awsClient.AppsS3Bucket
+		name := awsclient.BucketWithDefaults("")
```
catalintomai, line 59: s/`BucketWithDefaults`/`S3BucketWithDefaults` ? (to keep in line with the other `S3BucketExists`, `CreateS3Bucket`.

cmd/appsctl/provision_app.go
```diff
@@ -0,0 +1,146 @@
...
+)
+
+type functionInstallData struct {
+	zipFile io.Reader
+	name    string
+	handler string
+	runtime string
+}
+
+// ProvisionApp gets a release URL parses the release and creates an App in AWS
```
catalintomai, line 28: TODO: actualize comments.

cmd/appsctl/provision_app.go
```diff
@@ -0,0 +1,146 @@
...
+		return errors.Wrap(err, "can't get zip reader")
+	}
+	bundleFunctions := []functionInstallData{}
+	var mani *apps.Manifest
+
+	// Read all the files from zip archive
+	for _, file := range zipReader.File {
+		switch {
+		case strings.HasSuffix(file.Name, "manifest.json"):
+
```
catalintomai, line 67: nit: space

apps/app.go
```diff
@@ -167,10 +171,12 @@ var DefaultBindingsCall = &Call{
...
 
+// App describes an App installed on a Mattermost instance. App should be
+// abbreviated as `app`.
 type App struct {
-	AppID    AppID     `json:"app_id"`
-	Manifest *Manifest `json:"manifest"`
-	Status   AppStatus `json:"app_status"`
+	Manifest
+
+	Disabled bool `json:"disabled,omitempty"`
```
catalintomai, line 179: nit: use `Enabled` instead to be in line with `MarketplaceApp`, etc.?

apps/app.go
```diff
@@ -192,16 +198,9 @@ type App struct {
...
-	_ = json.Unmarshal(b, &out)
-	return out
-}
-
-func AppFromConfigMap(in interface{}) *App {
-	b, _ := json.Marshal(in)
-	var out App
-	_ = json.Unmarshal(b, &out)
-	return &out
+type MarketplaceApp struct {
```
catalintomai, line 201: TODO: add comment describing MarketplaceApp  (e.g. what is the difference from `Manifest`).

cmd/appsctl/provision_app.go
```diff
@@ -0,0 +1,146 @@
...
+	if err != nil {
+		return errors.Wrap(err, "can't get zip reader")
+	}
+	bundleFunctions := []functionInstallData{}
+	var mani *apps.Manifest
+
+	// Read all the files from zip archive
+	for _, file := range zipReader.File {
+		switch {
+		case strings.HasSuffix(file.Name, "manifest.json"):
```
catalintomai, line 66: move `manifest.json` to a const?

awsclient/client.go
levb: This setting was there before, I didn't change it. I'm 1/5 against having any defaults for AWS settings for the time being, the more explicit the better.

apps/app.go
levb: This was deliberate: meaningful nil defaults, by default (a newly created) App should be enabled, unless explicitly disabled. In UX "enabled" makes more sense.

apps/app.go
levb: @hanzei introduced this data struct, do you mind writing up a comment about how it is used?

cmd/appsctl/provision_app.go
levb: This is the only place where it's used in a functional way, and it's in `appsctl`, tempted to leave it alone for now.

server/api/constants.go
```diff
@@ -7,11 +7,14 @@ package api
 const (
 	Repository     = "mattermost-plugin-apps"
 	CommandTrigger = "apps"
+	ManifestsFile  = "manifests.json"
```
catalintomai, line 10: Is it replacing the `apps.json`? (in which case we need to update the references)

server/api/impl/admin/list.go
```diff
@@ -4,28 +4,64 @@
...
 	"github.com/mattermost/mattermost-plugin-apps/apps"
-	"github.com/mattermost/mattermost-plugin-apps/server/utils/md"
 )
 
-func (adm *Admin) ListApps() ([]*apps.App, md.MD, error) {
-	apps := adm.store.App().GetAll()
-	if len(apps) == 0 {
-		return nil, "no apps installed", nil
-	}
+func (adm *Admin) GetApp(appID apps.AppID) (*apps.App, error) {
```
catalintomai, line 12: s/`GetApp`/`GetInstalledApp` (to keep in line with `ListInstalledApps`) ?

server/api/impl/admin/list.go
```diff
@@ -4,28 +4,64 @@
...
+
+func (adm *Admin) GetManifest(appID apps.AppID) (*apps.Manifest, error) {
+	return adm.store.Manifest().Get(appID)
+}
+
+func (adm *Admin) ListInstalledApps() map[apps.AppID]*apps.App {
+	return adm.store.App().AsMap()
+}
+
+func (adm *Admin) ListMarketplaceApps(filter string) map[apps.AppID]*apps.MarketplaceApp {
```
catalintomai, line 24: Is this the proper name since manifests "They generally come from the marketplace, and can also be installed as overrides by developers."?

catalintomai, line 24: Is this the proper name since manifests "They generally come from the marketplace, and can also be installed as overrides by developers."?

levb: Right, `ListAvailableApps` is probably better.

server/api/impl/admin/synchronize.go
```diff
@@ -4,196 +4,64 @@
...
-
-func (adm *Admin) populateManifests(appVersions apps.AppVersionMap) {
-	adm.store.Manifest().Cleanup()
-
-	for id, version := range appVersions {
-		manifest, err := adm.awsClient.GetManifest(id, version)
-		if err != nil {
-			// Note that we are not returning an error here. Instead we drop the app from the list.
-			delete(appVersions, id)
+// LoadAppsList synchronizes apps with the apps.json file.
```
catalintomai, line 17: TODO: update comments

server/api/impl/store/manifest.go
```diff
@@ -0,0 +1,304 @@
...
+	}
+	defer f.Close()
+
+	return s.initGlobal(awscli, f, assetPath)
+}
+
+func (s *manifestStore) initGlobal(awscli awsclient.Client, manifestsFile io.Reader, assetPath string) error {
+	global := map[apps.AppID]*apps.Manifest{}
+
+	// Read in the marketplace-listed manifests from S3, as per versions
```
catalintomai, line 68: The manifests are not S3 only anymore.

server/api/impl/store/manifest.go
```diff
@@ -0,0 +1,304 @@
...
+	data, err := json.Marshal(m)
+	if err != nil {
+		return err
+	}
+	sha := fmt.Sprintf("%x", sha1.Sum(data)) // nolint:gosec
+	if sha == prevSHA {
+		return nil
+	}
+
+	_, err = s.mm.KV.Set(api.PrefixLocalManifest+sha, m)
```
catalintomai, line 208: why do we need to `Set` and then `Delete` the manifest?

server/api/impl/store/manifest.go
```diff
@@ -0,0 +1,304 @@
...
+	s.mutex.Unlock()
+
+	updated := map[string]string{}
+	for k, v := range conf.LocalManifests {
+		updated[k] = v
+	}
+	updated[string(m.AppID)] = sha
+	sc := *conf.StoredConfig
+	sc.LocalManifests = updated
+	err = s.conf.StoreConfig(&sc)
```
catalintomai, line 234: why do we need to store the manifest in both the config and the KV store?

apps/app.go
```diff
@@ -98,9 +99,13 @@ const (
 // Function describes app's function mapping
 // For now Function can be either AWS Lambda or HTTP function
 type Function struct {
+	Path    string `json:"path"`
 	Name    string `json:"name"`
 	Handler string `json:"handler"`
 	Runtime string `json:"runtime"`
+
+	// filled out by appsctl
+	LambdaName string `json:"lambda_name,omitempty"`
```
catalintomai, line 108: is the assignment part of this PR?

server/api/config.go
```diff
@@ -6,22 +6,31 @@ import (
...
+	// and the App struct is stored in KV under app_<sha1(App)>. Implemenrtation
+	// in `store.App`.
+	InstalledApps map[string]string `json:"installed_apps,omitempty"`
 
-type ConfigMapper interface {
-	ConfigMap() (result map[string]interface{})
-}
+	// LocalManifests is a list of locally-stored manifests. Local is in
+	// contrast to the "global" list of manifests which in the initial version
+	// is loaded from S3.
```
catalintomai, line 21: I guess there is the corner case where the global list contains a pointer to a file:// manifest which can be a local one - local server/aws deployment would be another contrived example ?)

server/plugin.go
```diff
@@ -61,58 +66,84 @@ func (p *Plugin) OnActivate() error {
...
+		return errors.Wrap(err, "failed to initialize the manifest store")
+	}
+	err = mstore.InitGlobal(p.aws.Client())
+	if err != nil {
+		return errors.Wrap(err, "failed to initialize the global manifest list from marketplace")
+	}
+	// app store
+	appstore := p.store.App()
+	appstore.InitBuiltin(
+		builtin_hello.App(),
```
catalintomai, line 99: why is that we do not include the `aws_hello` as a builtin app? (I guess the underlying q is: should all apps with builtin manifests  be automatically installed?)

apps/app.go
iomodo: This will make manifest file mutable right? I'm 1/5 that if possible we should stick with the immutable manifest.

apps/app.go
iomodo: Is `enable/disable` p1 feature? I thought we where only having `install/uninstall`.

server/api/config.go
```diff
@@ -6,22 +6,31 @@ import (
...
-	Apps               map[string]interface{}
-	AWSAccessKeyID     string
-	AWSSecretAccessKey string
-}
+	// InstalledApps is a list of all apps installed on the Mattermost instance.
+	//
+	// For each installed app, an entry of string(AppID) -> sha1(App) is added,
+	// and the App struct is stored in KV under app_<sha1(App)>. Implemenrtation
+	// in `store.App`.
+	InstalledApps map[string]string `json:"installed_apps,omitempty"`
```
iomodo, line 17: If we are storing apps in the KV store why do we need this struct?

server/api/impl/admin/install.go
```diff
@@ -6,27 +6,33 @@ package admin
...
 
-	app, err := adm.store.App().Get(cc.AppID)
+	m, err := adm.store.Manifest().Get(cc.AppID)
 	if err != nil {
 		return nil, "", err
 	}
 
-	app.GrantedPermissions = in.GrantedPermissions
-	app.GrantedLocations = in.GrantedLocations
+	app, _ := adm.store.App().Get(cc.AppID)
```
iomodo, line 28: I'd argue that app should come with the manifest attached already. Or we could add `getWithManifest`/`getWithoutManifest` methods for more clarity

server/api/impl/admin/list.go
iomodo: Are all of these apps `installed` though?

server/api/impl/admin/list.go
```diff
@@ -4,28 +4,64 @@
...
-	}
+func (adm *Admin) GetApp(appID apps.AppID) (*apps.App, error) {
+	return adm.store.App().Get(appID)
+}
+
+func (adm *Admin) GetManifest(appID apps.AppID) (*apps.Manifest, error) {
+	return adm.store.Manifest().Get(appID)
+}
+
+func (adm *Admin) ListInstalledApps() map[apps.AppID]*apps.App {
```
iomodo, line 20: Should this be `installed` or `registered`?

apps/app.go
hanzei: What is the difference between `Name` and `LambdaName`?

server/api/impl/store/manifest.go
```diff
@@ -0,0 +1,304 @@
...
+	if s.builtin == nil {
+		s.builtin = map[apps.AppID]*apps.Manifest{}
+	}
+	for _, m := range manifests {
+		s.builtin[m.AppID] = m
+	}
+	s.mutex.Unlock()
+}
+
+func (s *manifestStore) InitGlobal(awscli awsclient.Client) error {
```
iomodo, line 50: I don't like the coupling here. Why should a store have a knowledge of the aws?

server/api/impl/admin/list.go
levb: Yes, "Apps" are locally installed and active (unless disabled, but that's not implemented yet), and "Manifests" are "known" - i.e. listed in the marketplace.json, or added locally with `/apps debug-add-manifest`.

server/api/impl/store/manifest.go
levb: We store the map of `ID`:`sha` in the config (to keep it small), and store the payload in the KV as `man_$(SHA)`.

levb: I'll add a comment at the top of the app.go and manifest.go files

apps/app.go
levb: 1. `Name` - my understanding is that's the name of an entrypoint that AWS will use to invoke a function. `LambdaName` is the **invokable** name, the instance only needs to know `Path`->`LambdaName` and does not need anything else.
2. There really are 2 different things, both described as "the manifest":
  - a. a declaration of metadata and assets by the developer
  - b. a complete, "actionable" manifest generated by `appsctl`. The latter can be different from the former. 
Why is immutability of the manifest important? I'd much rather see the functions themselves immutable, i.e. content-addressed.

server/api/config.go
levb: The case of having a `file://` listed in `manifests.json` is hypothetical, and is intended mostly for internal debugging. 

The locally-installed manifests are those explicitly added - for now dev mode, eventually to add unlisted (custom/3rd party) Apps. One _could_ just edit the manifests file, but this is an override mechanism.

server/plugin.go
levb: A `builtin` app is one that is compiled in, and implements `Upstream` directly. No AWS (nor an HTTP) app can be "builtin". In the manifests store, "built-in" is a misnomer since all that is builtin is a manifest, so in this case AWS Hello appears listed and is available to install. 2/5 we could get rid of "builtin manifests" and just add them to `manifest file` if needed, there is no present use-case except for the AWS Hello, and that one can just expose its manifest as `file:`.

server/api/config.go
levb: To reload the KV values in memory upon changes. A change triggers a config change, triggers OnCOnfigurationChange on all servers, re-loading the cache

server/api/impl/admin/install.go
levb: Not sure what this means. An `App` embeds a `Manifest`.

server/api/impl/store/manifest.go
levb: 'Cause the list needs to be initialized. I can move the "InitGlobal" into a function in the `Admin` package, and then have an `InitGlobal` in the store take a map, but really why? This is a 1-time function, and it's passed a client to do its job. manifestStore does not hold on to AWS.

server/api/impl/store/manifest.go
```diff
@@ -0,0 +1,304 @@
...
+func (s *manifestStore) getFromS3(awscli awsclient.Client, bucket string, appID apps.AppID, version apps.AppVersion) ([]byte, error) {
+	name := fmt.Sprintf("manifest_%s_%s", appID, version)
+	data, err := awscli.GetS3(bucket, name)
+	if err != nil {
+		return nil, errors.Wrapf(err, "failed to download manifest %s", name)
+	}
+	return data, nil
+}
+
+func validateManifest(m *apps.Manifest) error {
```
hanzei, line 286: Should we stick to the naming pattern used in the MM server and use  `IsValid`?

hanzei, line 286: That functionality does seem store specific. Should we move it next to the `Manifest` struct to be used in e.g the install command?

server/command/install.go
```diff
@@ -4,91 +4,53 @@
...
 )
 
 func (s *service) executeInstall(params *params) (*model.CommandResponse, error) {
 	appID := ""
-	manifestURL := ""
 	appSecret := ""
-	force := false
 	fs := pflag.NewFlagSet("", pflag.ContinueOnError)
 	fs.StringVar(&appID, "app-id", "", "ID of the app")
-	fs.StringVar(&manifestURL, "url", "", "manifest URL")
```
hanzei, line 26: Should we keep the `--url` parameter to allow the installation with only one command, not two?

server/command/list.go
```diff
@@ -4,13 +4,68 @@
...
 	"github.com/mattermost/mattermost-server/v5/model"
+
+	"github.com/mattermost/mattermost-plugin-apps/server/utils/md"
 )
 
 func (s *service) executeList(params *params) (*model.CommandResponse, error) {
-	_, txt, err := s.api.Admin.ListApps()
-	if err != nil {
-		return errorOut(params, err)
+	marketplaceApps := s.admin.ListMarketplaceApps("")
```
hanzei, line 15: Should that be 
```suggestion
	listedApps := s.admin.ListMarketplaceApps("")
```

server/command/list.go
```diff
@@ -4,13 +4,68 @@
...
+	installedApps := s.admin.ListInstalledApps()
+
+	txt := md.MD("| Name | Status | Version | Account | Locations | Permissions |\n")
+	txt += md.MD("| :-- |:-- | :-- | :-- | :-- | :-- |\n")
+
+	for _, app := range installedApps {
+		mapp := marketplaceApps[app.AppID]
+		if mapp == nil {
```
hanzei, line 23: Nit pick:
```suggestion
		mapp, ok := marketplaceApps[app.AppID]
		if !ok || mapp == nil {
```

server/http/dialog/install.go
```diff
@@ -14,36 +14,38 @@ import (
...
+		md.Markdownf("\nApplication %s requires to add the following to the Mattermost user interface:", m.DisplayName)) + "\n"
+	for _, l := range m.RequestedLocations {
 		intro += md.Markdownf("- %s\n", l.Markdown())
 	}
 	intro += "\n---\n"
 
-	elements := []model.DialogElement{
-		{
+	elements := []model.DialogElement{}
+	if m.Type == apps.AppTypeHTTP {
```
hanzei, line 37: TODO: Talk about offline with @levb if we need that

server/utils/httputils/utils.go
```diff
@@ -81,17 +67,55 @@ func WriteJSONStatus(w http.ResponseWriter, statusCode int, v interface{}) {
...
+func ReadAndClose(in io.ReadCloser) ([]byte, error) {
+	defer in.Close()
+	return LimitReadAll(in, InLimit)
+}
+
+func LimitReadAll(in io.Reader, limit int64) ([]byte, error) {
+	if in == nil {
+		return []byte{}, nil
+	}
+	return ioutil.ReadAll(&io.LimitedReader{R: in, N: limit})
```
hanzei, line 96: nit pick: `ioutil` is deprecated 
```suggestion
	return io.ReadAll(&io.LimitedReader{R: in, N: limit})
```

server/utils/httputils/utils.go
```diff
@@ -81,17 +67,55 @@ func WriteJSONStatus(w http.ResponseWriter, statusCode int, v interface{}) {
...
+}
+
+func GetFromURL(url string) ([]byte, error) {
+	// nolint:gosec
+	resp, err := http.Get(url)
+	if err != nil {
+		return nil, err
+	}
+	defer resp.Body.Close()
+	return ioutil.ReadAll(resp.Body)
```
hanzei, line 120:
```suggestion
	return io.ReadAll(resp.Body)
```

hanzei (changes requested):
There are a lot of good changes in here. I quite like the plugin store. :+1:

server/command/debug.go
```diff
@@ -4,131 +4,57 @@
...
-			Manifest: manifest,
-			Force:    true,
-		},
-	)
-	s.api.Mattermost.Log.Error(fmt.Sprintf("app = %v", app))
-
+	data, err := httputils.GetFromURL(manifestURL)
 	if err != nil {
-		return errorOut(params, err)
+		return nil, err
```
hanzei, line 42: Given that `nil` is returned, there is no output to the user. Should we changed that?	i.e. when running `/apps debug-add-manifest --url`

hanzei:
I've tested the code with two AWS app and good news: The manifests synchronization works. Still, two things stood out to me:
1. The plugin takes like 10 seconds to start. This does _not_ come from the time it takes to load the manifests. While I don't mind the duration itself, it weird that it takes so long.
2.The data format for `manifests.json` looks weird. While full URLs are used for HTTP and relative paths for local files, manifests stored in S3 use only the version.
```json
 {
    "test-zendesk": "file:/app-test-zendesk.json",
    "com.mattermost.servicenow": "s3:0.1.0",
	"some-other-app": "http://localhost:8080/manifest",
    "zendesk": "s3:0.0.1"
}
```
Either we should use the full S3-URI e.g. `s3://mattermost-apps-bucket/manifests/com.mattermost.servicenow_0.1.0.json` or use a separate sub field which is a `map[id]version`.

apps/app.go
hanzei: Maybe I'm missing something, but `Name`  is what we use right now to invoke the lambda function (https://github.com/mattermost/mattermost-plugin-apps/blob/6764b4cdaf4628e7f0e87236e58d5c74ed134989/server/api/impl/upstream/upawslambda/upstream.go#L64)

server/utils/httputils/utils.go
levb: `ReadAll` is not in the `io` package, or am I behind on my go versions?

server/http/dialog/install.go
levb: @hanzei did we discuss it to your satisfaction?

server/command/list.go
levb: It is unnecessary to use an ok here since `marketplaceApps` is guaranteed to be non-nil.

apps/app.go
levb: I had the flag in my branch, already used in a few places (like `/apps list`) - figured no harm leaving it in the code even if not implemented yet, seems a rather obvious next step?

server/api/impl/store/manifest.go
```diff
@@ -0,0 +1,304 @@
...
+	conf := s.conf.GetConfig()
+	for appID, loc := range manifestLocations {
+		parts := strings.SplitN(string(loc), ":", 2)
+		switch {
+		case len(parts) == 1:
+			data, err = s.getFromS3(awscli, conf.AWSManifestBucket, appID, apps.AppVersion(parts[0]))
+		case len(parts) == 2 && parts[0] == "s3":
+			data, err = s.getFromS3(awscli, conf.AWSManifestBucket, appID, apps.AppVersion(parts[1]))
+		case len(parts) == 2 && parts[0] == "file":
+			data, err = ioutil.ReadFile(filepath.Join(assetPath, parts[1]))
```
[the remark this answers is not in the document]
levb: Prob. not. It just seemed easy and potentially useful to have it in a local file, even if for development/debugging.

server/api/impl/store/apps.go
```diff
@@ -4,94 +4,185 @@
...
 	"github.com/mattermost/mattermost-plugin-apps/apps"
 	"github.com/mattermost/mattermost-plugin-apps/server/api"
 	"github.com/mattermost/mattermost-plugin-apps/server/utils"
 )
 
-type AppStore struct {
+type appStore struct {
 	*Store
+
+	mutex sync.RWMutex
```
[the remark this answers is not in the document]
levb: `installed`, changed in OnConfigurationChanged

server/api/impl/store/apps.go
```diff
@@ -4,94 +4,185 @@
...
+		out[appID] = app
 	}
-	app := apps.AppFromConfigMap(v)
-	app = s.populateAppWithManifest(app)
-	return app, nil
+	return out
 }
 
-func (s AppStore) Save(app *apps.App) error {
+func (s *appStore) Save(app *apps.App) error {
```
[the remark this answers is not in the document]
levb: Yes it can, on multiple levels, and no, I don't think it's a big issue - apps and manifests are not very volatile. I'd be more concerned about 2 rapid app installs resulting in only one actually installed, but again, probably not realistic.

server/api/impl/store/apps.go
```diff
@@ -4,94 +4,185 @@
...
-	cApp := &apps.App{}
-	*cApp = *app
-	// do not store manifest in the config
-	cApp.AppID = app.Manifest.AppID
-	cApp.Manifest = nil
+	data, err := json.Marshal(app)
+	if err != nil {
+		return err
+	}
+	sha := fmt.Sprintf("%x", sha1.Sum(data)) // nolint:gosec
```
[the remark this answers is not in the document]
levb: See https://play.golang.org/p/aRWpdf4XSSP

fmt is the simplest that I know. It is hacky.

server/api/impl/store/manifest.go
```diff
@@ -0,0 +1,304 @@
...
+	manifestLocations := apps.AppVersionMap{}
+	err := json.NewDecoder(manifestsFile).Decode(&manifestLocations)
+	if err != nil {
+		return err
+	}
+
+	var data []byte
+	conf := s.conf.GetConfig()
+	for appID, loc := range manifestLocations {
+		parts := strings.SplitN(string(loc), ":", 2)
```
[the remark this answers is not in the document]
levb: My hesitation there was 2-fold, (a) file:// would not be a "true" file URL, but one relative to `/assets`; (b) `s3://appversion` is not a valid S3 URL. It'd be my preference if it were a `map[AppID]valid_url_string`

server/api/impl/store/manifest.go
```diff
@@ -0,0 +1,304 @@
...
+		switch {
+		case len(parts) == 1:
+			data, err = s.getFromS3(awscli, conf.AWSManifestBucket, appID, apps.AppVersion(parts[0]))
```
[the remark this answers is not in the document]
levb: I suggest we consider another PR, if at all. This is an "internal" format, and likely to go away soon altogether.

server/api/store.go
```diff
@@ -12,22 +15,27 @@ type Store interface {
 }
 
 type AppStore interface {
+	Configurable
+
+	AsMap() map[apps.AppID]*apps.App
```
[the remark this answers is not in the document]
levb: For the store 1/5 made more sense to me to return the "raw" data struct. `admin.GetInstalledApps` and `admin.GetListedApps` - perhaps would be better off returning sorted slices. Separate PR perhaps if we decide so?

server/api/impl/admin/list.go
levb: These (Apps) are installed. There is no longer a "registered", only "Listed" aka Marketplace.

server/api/impl/admin/list.go
levb: I don't see any significant difference in what these would do and the data structure, so 1/5 keep them as one `GetListedApps() map[AppID]*ListedApp`

apps/app.go
levb: Removed `LambdaName`

server/api/impl/store/apps.go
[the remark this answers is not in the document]
hanzei: Hm, I'm wondering why my suggestion didn't work. Anyway, I'm fine with `fmt`.

server/api/impl/store/apps.go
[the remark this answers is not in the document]
hanzei: 0/5 Maybe add a comment on that?

server/api/impl/store/apps.go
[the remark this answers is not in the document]
hanzei: Should a comment make that clear?

server/http/dialog/install.go
hanzei: Yes, resolved

server/utils/httputils/utils.go
hanzei: Yes, go 1.16 added it https://golang.org/pkg/io/#ReadAll

server/api/impl/admin/synchronize.go
```diff
@@ -4,196 +4,64 @@
...
+	installed := adm.ListInstalledApps()
+	marketplace := adm.ListMarketplaceApps("")
 
-			adm.mm.Log.Error("failed to get manifest", "app", id, "version", version, "err", err.Error())
+	changed := map[apps.AppID]*apps.App{}
+	for _, app := range installed {
+		mapp := marketplace[app.AppID]
 
+		// exclude unlisted apps, or those that need no action.
+		if mapp == nil || app.Version == mapp.Manifest.Version {
```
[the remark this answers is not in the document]
levb: Not needed, map is non-nil

server/utils/httputils/utils.go
levb: 👍  already got it, thx!

server/api/impl/admin/synchronize.go
```diff
@@ -4,196 +4,64 @@
...
+func (adm *Admin) UpdateInstalledApps() error {
+	installed := adm.ListInstalledApps()
+	marketplace := adm.ListMarketplaceApps("")
 
-			adm.mm.Log.Error("failed to get manifest", "app", id, "version", version, "err", err.Error())
+	changed := map[apps.AppID]*apps.App{}
+	for _, app := range installed {
+		mapp := marketplace[app.AppID]
 
+		// exclude unlisted apps, or those that need no action.
```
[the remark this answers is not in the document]
levb: I am making an argument that we only provide onVersionChanged callback to the App; and do not presently offer automatic install/uninstall functionality.

server/api/impl/admin/synchronize.go
```diff
@@ -4,196 +4,64 @@
...
 
-	// register new apps
-	for appID, manifest := range adm.store.Manifest().GetAll() {
-		if _, ok := registeredAppsMap[appID]; ok {
-			continue
-		}
-		if err := adm.RegisterApp(manifest); err != nil {
-			adm.mm.Log.Error("can't register an app", "app_id", appID)
+		app.Manifest = *mapp.Manifest
+		err := adm.store.App().Save(app)
```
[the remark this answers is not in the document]
levb: 0/5 I'd prefer that we'd had `admin.UpdateApp`, similar to InstallApp, and UninstallApp; all of them can have callOnce functionality; separate PR?

apps/app.go
```diff
@@ -167,10 +171,12 @@ var DefaultBindingsCall = &Call{
 	Path: DefaultBindingsCallPath,
 }
 
+// App describes an App installed on a Mattermost instance. App should be
+// abbreviated as `app`.
 type App struct {
```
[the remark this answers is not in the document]
levb: A separate PR? I agree that we need a `install --url <xyz> --type http --root-url <abc>` that would override/augment the otherwise "awslambda" manifest.

awsclient/generate_names.go
```diff
@@ -1,27 +1,33 @@
...
 
-// getFunctionName generates function name for a specific app
-// name can be 64 characters long.
-func getFunctionName(appID apps.AppID, version apps.AppVersion, function string) (string, error) {
-	if len(appID) > appIDLengthLimit {
-		return "", errors.Errorf("appID %s too long, should be %d bytes", appID, appIDLengthLimit)
+const LambdaFunctionFileNameMaxSize = 64
+const AppIDLengthLimit = 32
+const VersionFormat = "v00.00.000"
+const StaticAssetsFolder = "static/"
```
catalintomai, line 21: I believe we have them in `provision_app.go` as well (and moved to another file in https://github.com/mattermost/mattermost-plugin-apps/pull/93/files)

server/api/impl/store/apps.go
```diff
@@ -4,94 +4,185 @@
...
-	if len(conf.Apps) == 0 {
-		return out
+func (s *appStore) Configure(conf api.Config) error {
+	newInstalled := map[apps.AppID]*apps.App{}
+
+	for id, key := range conf.InstalledApps {
+		var app *apps.App
+		err := s.mm.KV.Get(api.PrefixInstalledApp+key, &app)
+		if err != nil {
+			s.mm.Log.Error(
```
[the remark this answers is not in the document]
catalintomai: isn't it skipped (not needing to add the `break` in golang)?

apps/app.go
```diff
@@ -110,20 +64,50 @@ type Manifest struct {
...
+	OAuth2ClientSecret string `json:"oauth2_client_secret,omitempty"`
+	OAuth2TrustedApp   bool   `json:"oauth2_trusted_app,omitempty"`
+
+	BotUserID      string `json:"bot_user_id,omitempty"`
+	BotUsername    string `json:"bot_username,omitempty"`
+	BotAccessToken string `json:"bot_access_token,omitempty"`
+
+	// Grants should be scopable in the future, per team, channel, post with
+	// regexp.
+	GrantedPermissions Permissions `json:"granted_permissions,omitempty"`
```
catalintomai, line 96: Would be useful to document the initial set (AsBot, etc.) like present+future.

assets/manifests.json
```diff
@@ -0,0 +1,5 @@
+{
+    "awsHello": "file:example-aws-hello.json",
+    "zendesk": "0.0.1",
```
catalintomai, line 3: why is it not "com.mattermost.zendesk"? (or do we support both)

awsclient/lambda.go
```diff
@@ -0,0 +1,99 @@
...
+		}
+	}
+	c.logger.Info(fmt.Sprintf("function named %s was created with result - %v", function, result))
+
+	return nil
+}
+
+func (c *client) CreateOrUpdateLambda(zipFile io.Reader, function, handler, runtime, resource string) error {
+	if zipFile == nil || function == "" {
+		return errors.New("you must supply a zip file, function name, handler, ARN and runtime")
```
catalintomai, line 71: remove "handler", etc.?

awsclient/s3.go
```diff
@@ -0,0 +1,75 @@
...
+		Bucket: aws.String(bucket),
+		Key:    aws.String(item),
+	})
+	if err != nil {
+		return nil, errors.Wrap(err, "failed to download file")
+	}
+	return buffer.Bytes(), nil
+}
+
+// S3FileUpload uploads file to a specific S3 bucket
```
catalintomai, line 37: s/S3FileUpload/UploadS3

awsclient/s3.go
```diff
@@ -0,0 +1,75 @@
...
+		Bucket: aws.String(bucket),
+		Key:    aws.String(key),
+		Body:   body,
+	}); err != nil {
+		return errors.Wrap(err, "failed to upload file")
+	}
+	return nil
+}
+
+func (c *client) CreateS3Bucket(bucket string) error {
```
catalintomai, line 49: nit: remove `Bucket` (here and below) or add it to the methods above.

cmd/appsctl/provision_app.go
catalintomai: this one still needs to be updated.

server/api/impl/store/manifest.go
```diff
@@ -0,0 +1,266 @@
...
+	f, err := os.Open(filepath.Join(assetPath, api.ManifestsFile))
+	if err != nil {
+		return errors.Wrap(err, "failed to load global list of available apps")
+	}
+	defer f.Close()
+
+	return s.initGlobal(awscli, bucket, f, assetPath)
+}
+
+// initGlobal reads in the list of known (i.e. marketplace listed) app
```
catalintomai, line 56: would be good to add a comment as to why (unlike apps) we do not store global manifests in the KV store.

server/api/impl/store/apps.go
```diff
@@ -4,97 +4,190 @@
...
+
+	"github.com/pkg/errors"
+
 	"github.com/mattermost/mattermost-plugin-apps/apps"
 	"github.com/mattermost/mattermost-plugin-apps/server/api"
 	"github.com/mattermost/mattermost-plugin-apps/server/utils"
 )
 
-type AppStore struct {
+// appStore combines installed and builtin Apps.  The installed Apps are stored
```
catalintomai, line 19: as an observation, it is maybe worth noticing that for apps - we store the builtin ones in memory and the non-builtin/external/global(?) in KV store while for manifests we store the global ones in memory and the local ones in the KV store.

catalintomai (approved):
couple small obs. LGTM!

hanzei:
@levb The e2e tests now fail with a more reasonable error message:
```
    restapitestlib.go:104: 
        	Error Trace:	restapitestlib.go:104
        	            				restapi_test.go:18
        	Error:      	Received unexpected error:
        	            	failed to initialize the global manifest list from marketplace: failed to load global manifest for zendesk: failed to download manifest manifest_zendesk_0.0.1: failed to download file: EmptyStaticCreds: static credentials are empty
        	Test:       	TestPPAPI
--- FAIL: TestPPAPI (28.21s)
FAIL
```

awsclient/generate_names.go
levb: Yes, I moved provision_app into `appsctl`, but the name generation functions are needed in the `server` also. Technically, these constants do not need to be public, but it's nice for godoc purposes IMO.

assets/manifests.json
levb: ATM, the top-level command is the same as appid. If we decouple them, the app_id is just a (unique) string. I'm 0/5 on the semantics and conventions.

awsclient/lambda.go
levb: not sure, @iomodo?

awsclient/s3.go
levb: It creates the bucket, thus the distinction (others operate on entities in the bucket)

awsclient/lambda.go
iomodo: correct `handler, ARN and runtime` are not needed in the error

cmd/appsctl/provision_app.go
```diff
@@ -0,0 +1,216 @@
...
+				continue
+			}
+		}
+	}
+
+	err = provisionAssets(awscli, m, assets)
+	if err != nil {
+		return errors.Wrapf(err, "can't provision assets of the app - %s", m.AppID)
+	}
+	err = provisionFunctions(awscli, m, resFunctions, shouldUpdate)
```
iomodo, line 139: this too

cmd/appsctl/provision_app.go
```diff
@@ -0,0 +1,216 @@
...
+	if err != nil {
+		return errors.Wrapf(err, "can't marshal manifest for app - %s", m.AppID)
+	}
+	buffer := bytes.NewBuffer(data)
+
+	bucket := awsclient.GenerateS3BucketNameWithDefaults("")
+	key := awsclient.GenerateManifestS3Name(m.AppID, m.Version)
+	if err := awscli.UploadS3(bucket, key, buffer); err != nil {
+		return errors.Wrapf(err, "can't upload manifest file for the app - %s", m.AppID)
+	}
```
iomodo, line 153: would be nice to have a method like `provisionManifest` or at least a comment. It would read nicely: provisionAssets, provisionFunctions, provisionManifest.

server/api/impl/upstream/upawslambda/upstream.go
```diff
@@ -100,3 +118,20 @@ func callToInvocationPayload(call *apps.Call) ([]byte, error) {
 
 	return payload, nil
 }
+
+func match(callPath string, app *apps.App) (string, error) {
+	matchedName := ""
+	matchedPath := ""
+	for _, f := range app.Functions {
+		if strings.HasPrefix(callPath, f.Path) {
+			if len(f.Path) > len(matchedPath) {
```
iomodo, line 127: `matchedPath` assignment is missing

server/plugin.go
```diff
@@ -62,73 +67,95 @@ func (p *Plugin) OnActivate() error {
...
 	}
 
-	p.api = &api.Service{
-		Mattermost:   mm,
-		Configurator: conf,
-		Proxy:        proxy,
-		AppServices:  appservices.NewAppServices(mm, conf, store),
-		Admin:        admin.NewAdmin(mm, conf, store, proxy, awsClient, mutex),
-		AWS:          awsClient,
+	assetBucket := awsclient.GenerateS3BucketNameWithDefaults("")
```
iomodo, line 115: manifestBucket and assetBucket is the same.

server/api/impl/proxy/proxy.go
```diff
@@ -19,18 +19,26 @@ import (
...
 
-	mm        *pluginapi.Client
-	conf      api.Configurator
-	store     api.Store
-	awsClient *aws.Client
+	mm            *pluginapi.Client
+	conf          api.Configurator
+	store         api.Store
+	aws           aws.Service
+	s3AssetBucket string
```
iomodo, line 28: Why do we store bucket name here and not in the aws?

server/utils/httputils/utils.go
iomodo: still needs change

iomodo (changes requested):

server/api/impl/proxy/proxy.go
levb: 1/5 By keeping configuration issues out of the client, we can define how we configure each use case's bucket where it needs to be? The Client is now as simple as it can be, just a convenience wrapper on top of the AWS API + auth, retries, etc.

server/plugin.go
levb: Yes, by design since they were currently the same, I didn't want to disturb it, just added potential de-coupling.

server/api/impl/upstream/upawslambda/upstream.go
levb: Oops, had a favorable order in the test. Fixed.

server/api/impl/store/apps.go
```diff
@@ -4,97 +4,190 @@
...
+	updatedInstalled := map[apps.AppID]*apps.App{}
+	for k, v := range installed {
+		if k != app.AppID {
+			updatedInstalled[k] = v
+		}
+	}
+	updatedInstalled[app.AppID] = app
+	s.mutex.Lock()
+	s.installed = updatedInstalled
+	s.mutex.Unlock()
```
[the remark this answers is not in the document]
levb: Because like you noted before, pointers to the map have been given out; this way it's copy-on-write, rather than copy-on-read.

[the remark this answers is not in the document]
levb: PS. @iomodo as we discussed offline, I didn't want to deal with the "ugliness" of using sync.Map's, 2/5 leave as is for now unless you see a functional problem.

server/api/impl/store/apps.go
levb: Apps are only "installed" or builtin (also pre-installed). The manifest store is possibly going to change somewhat after discussing with @iomodo and @hanzei, tempted to add another override layer, `dev-manifests.json` that'd be combined with `manifests.json` in memory. 1/5 we keep the HA-capable "local" manifest store for future use-cases.

apps/app.go
levb: Doing it in the upcoming Docs PR

levb:
@hanzei @iomodo this is ready for review.

server/http/restapi/marketplace.go
```diff
@@ -2,84 +2,13 @@ package restapi
...
-		}
-	}
-
-	// Sort result alphabetically.
-	sort.SliceStable(result, func(i, j int) bool {
-		return strings.ToLower(result[i].Manifest.DisplayName) <
-			strings.ToLower(result[j].Manifest.DisplayName)
-	})
-
+	result := a.admin.GetListedApps(filter)
```
hanzei, line 12: Before this PR the API endpoint did send a list of apps. Now it's an map which causes a crash in the webapp. Would you mind reverting to the old data structure of a list?

levb: That was a mistake (sending the map out of the REST API), will do now.

levb: @hanzei not 100% sure how to test it, but should be reverted now

server/api/impl/admin/list.go
```diff
@@ -4,28 +4,80 @@
...
+		app, _ := adm.store.App().Get(m.AppID)
+		if app != nil {
+			marketApp.Installed = true
+			marketApp.Enabled = !app.Disabled
+		}
+		out = append(out, marketApp)
 	}
 
-	return apps, out, nil
+	// Sort result alphabetically, byu display name.
```
hanzei, line 54:
```suggestion
	// Sort result alphabetically, by display name.
```

server/api/impl/admin/manifest.go
```diff
@@ -0,0 +1,26 @@
...
+	"github.com/pkg/errors"
+
+	"github.com/mattermost/mattermost-plugin-apps/apps"
+	"github.com/mattermost/mattermost-plugin-apps/server/utils/md"
+)
+
+func (adm *Admin) AddLocalManifest(cc *apps.Context, sessionToken apps.SessionToken, m *apps.Manifest) (md.MD, error) {
+	if m.AppID == "" {
+		return "", errors.New("app ID must not be empty")
+	}
```
hanzei, line 16: Should we check for `IsValid` instead?

server/api/impl/configurator/configurator.go
```diff
@@ -1,20 +1,18 @@
...
-	"github.com/mattermost/mattermost-plugin-apps/server/api"
-	"github.com/mattermost/mattermost-plugin-apps/server/api/impl/aws"
-
 	"github.com/pkg/errors"
 
 	pluginapi "github.com/mattermost/mattermost-plugin-api"
 	"github.com/mattermost/mattermost-server/v5/model"
-)
 
-var _ api.Configurator = (*config)(nil)
```
hanzei, line 17: Don't we want to check that any longer?

server/api/impl/store/apps.go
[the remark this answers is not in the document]
hanzei: Gentle bump

server/api/impl/upstream/upawslambda/upstream.go
```diff
@@ -100,3 +118,21 @@ func callToInvocationPayload(call *apps.CallRequest) ([]byte, error) {
...
+				var err error
+				matchedPath = f.Path
+				matchedName, err = awsclient.GenerateLambdaName(app.AppID, app.Version, f.Name)
+				if err != nil {
+					return "", err
+				}
+			}
+		}
+	}
+	return matchedName, nil
```
hanzei, line 137: Should we return `utils.ErrNotFound` if no match is found?

hanzei (approved):
I've leaved some nit pick, but there is nothing blocking from my side :+1:

iomodo (approved):

server/api/impl/upstream/upawslambda/upstream.go
levb: No, we take the longest match; if you have a `/` it'll be returned as the last resort

apps/call.go
```diff
@@ -135,6 +135,23 @@ func MakeCall(url string) *Call {
 	return c
 }
 
+func (c *Call) WithOverrides(override *Call) *Call {
+	out := Call{}
+	if c != nil {
```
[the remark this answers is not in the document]
levb: That's the normal usage pattern, see https://github.com/mattermost/mattermost-plugin-apps/blob/8295243ac5174879544c989e49bf7588e24e22ea/server/api/impl/admin/install.go#L78

awsclient/generate_names.go
```diff
@@ -43,14 +49,25 @@ func getFunctionName(appID apps.AppID, version apps.AppVersion, function string)
...
-func GetAssetFileKey(appID apps.AppID, version apps.AppVersion, name string) string {
-	return fmt.Sprintf("%s%s_%s_app/%s", staticAssetsFolder, appID, version, name)
+func GenerateAssetS3Name(appID apps.AppID, version apps.AppVersion, name string) string {
+	return fmt.Sprintf("%s%s_%s_app/%s", StaticAssetsFolder, appID, version, name)
+}
+
+func GenerateS3BucketNameWithDefaults(name string) string {
+	if name != "" {
+		return name
+	}
```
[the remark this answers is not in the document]
levb: If it has a name, returns it; otherwise fills in the default. Kinda equiv to `name = configName | GenerateS3BucketName()`, a 1-liner.

apps/app.go
```diff
@@ -1,62 +1,22 @@
...
 	"github.com/pkg/errors"
+
+	"github.com/mattermost/mattermost-server/v5/model"
 )
 
 // AppID is a globally unique identifier that represents a Mattermost App.
 // Allowed characters are letters, numbers, underscores and hyphens.
 type AppID string
 
-func (id AppID) IsValid() error {
```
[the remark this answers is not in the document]
levb: Readability/typography of the file. The functions are of secondary importance; defining the types at the top helps read the types themselves, which are of much higher importance.

apps/app.go
```diff
@@ -66,41 +26,35 @@ const (
...
-	// consistently.
+	// See DefaultInstallCall, DefaultBindingCall, etc. for the defaults. The
+	// App developer can override the defaults by providing explicit Path,
+	// Expand values.
+	OnDisable        *Call `json:"on_disable,omitempty"`
+	OnEnable         *Call `json:"on_enable,omitempty"`
+	OnInstall        *Call `json:"on_install,omitempty"`
+	OnVersionChanged *Call `json:"on_version_changed,omitempty"`
+	OnUninstall      *Call `json:"on_uninstall,omitempty"`
+	Bindings         *Call `json:"bindings,omitempty"`
```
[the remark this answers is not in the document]
levb: They are in the docs PR

apps/app.go
```diff
@@ -110,20 +64,50 @@ type Manifest struct {
...
+// App describes an App installed on a Mattermost instance. App should be
+// abbreviated as `app`.
+type App struct {
+	Manifest
 
-	// Deployment manifest for hostable apps will include path->invoke mappings
-	Functions []Function
+	Disabled bool `json:"disabled,omitempty"`
+
+	// Secret is used to issue JWT
```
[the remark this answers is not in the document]
levb: In the docs PR

server/api/impl/configurator/configurator.go
levb: It is effectively checked in https://github.com/mattermost/mattermost-plugin-apps/pull/85/files#diff-42a3dc1fa0c237add93a7841342bdb1ab36a39db77a8c7adfc54a87ad654fc3cR28

PR #86, opened 2021-03-01
[MM-33377] Move AWS invoke credentials to env variables

hanzei:
#### Summary
In production, these credentials will be provided by the cloud team and configured via env variables.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-33377

### DIFF
--- a/server/plugin.go
+++ b/server/plugin.go
@@ -5,6 +5,7 @@ package main
 
 import (
 	gohttp "net/http"
+	"os"
 
 	"github.com/gorilla/mux"
 	"github.com/pkg/errors"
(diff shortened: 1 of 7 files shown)


server/plugin.go
```diff
@@ -64,7 +65,16 @@ func (p *Plugin) OnActivate() error {
...
+	accessKey := os.Getenv("APPS_INVOKE_AWS_ACCESS_KEY")
+	if accessKey == "" {
+		mm.Log.Warn("APPS_INVOKE_AWS_ACCESS_KEY is not set. AWS apps won't work.")
+	}
+	secretKey := os.Getenv("APPS_INVOKE_AWS_SECRET_KEY")
+	if secretKey == "" {
+		mm.Log.Warn("APPS_INVOKE_AWS_SECRET_KEY is not set. AWS apps won't work.")
+	}
+
+	awsClient := aws.NewAWSClient(accessKey, secretKey, &mm.Log)
```
iomodo, line 77: if `APPS_INVOKE_AWS_ACCESS_KEY` and `APPS_INVOKE_AWS_SECRET_KEY` are empty `aws.NewAWSClient` may still create a credentials. It is reading from the "default env vars" `AWS_ACCESS_KEY_ID` and `AWS_SECRET_ACCESS_KEY`. So either we get rid of the previous behaviour or correct the messages or get rid of this new env vars.

iomodo (changes requested):

server/plugin.go
hanzei: I'm inclined to get rid of get rid of the previous behavior and don't use `AWS_ACCESS_KEY_ID` and `AWS_SECRET_ACCESS_KEY`. Given that we are using two sets of credentials, I think it's better to be explicit. @iomodo What do you think?

iomodo: Agree

hanzei: Changed via c37e9d8

iomodo (approved):

levb (approved):

PR #91, opened 2021-03-03
Change call type submit and call response ok to not be empty strings

mickmister:
#### Ticket Link

Fixes https://mattermost.atlassian.net/browse/MM-31981

### DIFF
--- a/apps/call.go
+++ b/apps/call.go
@@ -11,7 +11,7 @@ type CallType string
 
 const (
 	// CallTypeSubmit (default) indicates the intent to take action.
-	CallTypeSubmit = CallType("")
+	CallTypeSubmit = CallType("submit")
 	// CallTypeForm retrieves the form definition for the current set of values,
 	// and the context.
 	CallTypeForm = CallType("form")
@@ -46,8 +46,7 @@ type CallResponseType string
 const (
 	// CallResponseTypeOK indicates that the call succeeded, and returns
 	// Markdown and Data.
-	// TODO update OK to be ["ok", "" default], update redux, webapp?
-	CallResponseTypeOK = CallResponseType("")
+	CallResponseTypeOK = CallResponseType("ok")
 
 	// CallResponseTypeOK indicates an error, returns Error.
 	CallResponseTypeError = CallResponseType("error")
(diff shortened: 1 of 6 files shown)


levb (approved):
No cases where we now need to add defaulting? Or default == failure?

mickmister:
@levb These constants aren't used anywhere except the hello app. I've added default cases for the hello app.

The frontend is continuing to treat an empty string call response type as `ok`, while also supporting `ok`. The frontend defaults to using call type `submit` if some piece of code calling `doAppCall` did not set the type.

server/examples/go/hello/builtin_hello/helloapp.go
```diff
@@ -95,7 +95,7 @@ func (h *helloapp) OneWay(call *apps.Call) error {
 }
 
 func (h *helloapp) Install(c *apps.Call) *apps.CallResponse {
-	if c.Type != apps.CallTypeSubmit {
+	if c.Type != apps.CallTypeSubmit && c.Type != "" {
```
larkox, line 98: This is a call from the proxy. Can we assume this will be well formed (and therefore will never receive a empty string type)?

apps/call.go
```diff
@@ -11,7 +11,7 @@ type CallType string
 
 const (
 	// CallTypeSubmit (default) indicates the intent to take action.
-	CallTypeSubmit = CallType("")
+	CallTypeSubmit = CallType("submit")
```
larkox, line 14: If we want to default, I would do it at proxy level. My approach would be:
- Assume calls from the client are well formed (the client is our code and should be correct)
- Assume responses from the apps may be wrong. Default to OK if no type is received.
- Send always proper responses to the client.

We could add some defaulting in the trip client->proxy->app to be safe, but one part of me want to leave it like that to find errors ASAP.

larkox (changes requested):
I think if we are going to default the information, we should do it in the proxy, not leave the app or the client uncertainties.

apps/call.go
mickmister: Response type is now defaulted to `ok`. I agree with no defaulting from `client->proxy->app`, to catch errors early.

server/examples/go/hello/builtin_hello/helloapp.go
mickmister: I think we should get rid of all empty string checks, and instead return an error when there is an unexpected call type (like an empty one).

I had this check in here to make sure all the different pieces would work on everyone's computer (webapp inconsistency due to not being merged yet or something). But we should plan for the future and not really factor that in.

larkox (approved):

PR #94, opened 2021-03-05
Extract Call request struct from Call struct

mickmister:
#### Summary

To differentiate between the Call that the App configures in bindings etc., and the Call requests that come form the client, we've made a new `CallRequest` struct that has all fields whereas the `Call` struct now just has the fields that the App sets.

#### Ticket Link

Fixes https://mattermost.atlassian.net/browse/MM-33578


larkox (approved):

levb:
@mickmister  Do you mind pointing the PR at https://github.com/mattermost/mattermost-plugin-apps/tree/call-type-submit for clarity? Can re-point to master once the ^^ is merged, or merge them together?

mickmister:
@levb This PR is already pointing to [call-type-submit](https://github.com/mattermost/mattermost-plugin-apps/tree/call-type-submit?rgh-link-date=2021-03-08T13%3A26%3A49Z). Idk what's going on with the commits. Both branches have @hanzei's circleci additions so the diff shouldn't include those AFAIK. @hanzei any idea what's going on with the diff?

hanzei:
`git diff origin/call-type-submit  origin/call-request` seems fine to me. Maybe that is a issue with GitHub presenting the dataß

apps/call.go
```diff
@@ -33,12 +33,19 @@ const (
...
-	Context    *Context               `json:"context,omitempty"`
-	RawCommand string                 `json:"raw_command,omitempty"`
-	Expand     *Expand                `json:"expand,omitempty"`
+	Path   string      `json:"path,omitempty"`
+	Expand *Expand     `json:"expand,omitempty"`
+	State  interface{} `json:"state,omitempty"`
+}
+
+type CallRequest struct {
+	*Call
```
hanzei, line 42: Why is a `Call` embedded as a pointer? This can cause issues like `CallRequest.Path` panicing because `Call` is `nil`, so I want to make sure we do that for a good reason.

mickmister: @hanzei If we go the route of embedding a `Call` struct instead of a pointer, each function constructing a `CallRequest` needs to dereference the pointer to the `Call` it is embedding. That doesn't change much about the nil pointer issue, but I do agree that there's no good reason for it to be a pointer.

We do get the benefit of having a `json.Unmarshal`'d `CallRequest` avoiding this nil pointer issue. The `Path` would be empty in that case, which may cause other problems.

I'll change it to embed the struct and not the pointer

levb: 3/5 @hanzei is right, we should embed the struct, not a pointer

levb:
LGTM, but I recommend embedding `Call` into `CallRequest` before merging.

hanzei:
@mickmister Merging `all-type-submit` into `call-request` might help with the diff.

levb:
@hanzei I think we are good to go as soon as you say so.

PR #97, opened 2021-03-07
Customize base commands and sub commands

mickmister:
#### Summary

In order to support customizing an App's base command, the binding structure of what the App needs to return is changing (breaking change for all Apps). This also allows us to potentially support more than one base command per App if we choose to support that in the future.

The original structure was:

```json
[
  {
    "location": "/command",
    "bindings": [
      {
        "location": "some-sub-command",
        "description": "Run the sub command",
        "icon": "Some sub command icon"
      }
    ]
  }
]
```

This is what App's will need to return now:

```json
[
  {
    "location": "/command",
    "bindings": [
      {
        "location": "the-app-id", // ignored because the base command is always the app id for now
        "description": "Run the base command",
        "icon": "Some base command icon",
        "bindings": [
          {
            "location": "some-sub-command",
            "description": "Run the sub command",
            "icon": "Some sub command icon"
          }
        ]
      }
    ]
  }
]
```

#### Ticket Link

Fixes https://mattermost.atlassian.net/browse/MM-33579


server/api/impl/proxy/bindings.go
```diff
@@ -89,7 +96,29 @@ func (p *Proxy) scanAppBindings(app *apps.App, bindings []*apps.Binding, locPref
...
 
-		if !fql.IsTop() {
+		if locPrefix == apps.LocationCommand {
+			b.Location = apps.Location(app.Manifest.AppID)
+			b.Label = string(app.Manifest.AppID)
+		}
+
+		if fql.IsTop() {
+			if locationsUsed[appB.Location] {
+				// TODO Log to the app that there are two bindings on the top level with the same location
```
levb, line 106: we should be careful logging from `GetBindings`, it'd be a flood of log messages.

Recommend that we remove the TODO, scan for and carefully reconsider any previous logging in `GetBindings`, maybe even add a comment at the top to be careful and not to include logging here without serious consideration for potential volume.

levb (approved):

server/api/impl/proxy/bindings.go
mickmister: @levb

- I changed the existing ones to DEBUG
- removed the comments here
- added a comment above the function saying to avoid excessive logging

mickmister: Actually just removed the existing ones too

PR #98, opened 2021-03-10
Expand bindings call

larkox:
#### Summary
Expand bindings call. Also avoid panic when the manifest is not available, and simplify the get bindings call.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-33447


server/api/impl/proxy/bindings.go
```diff
@@ -33,33 +31,51 @@ func mergeBindings(bb1, bb2 []*apps.Binding) []*apps.Binding {
...
+		if manifest.Bindings != nil {
+			bindingsCall = *manifest.Bindings
 		}
 		bindingsCall.Context = &appCC
 
-		bindings, err := upstream.GetBindings(up, bindingsCall)
+		resp := p.Call(debugSessionToken, &bindingsCall)
+
+		if resp == nil || resp.Type != apps.CallResponseTypeOK {
+			// TODO Log error (chance to flood the logs)
```
levb, line 62: 1/5 log at Debug level?

levb (approved):
Fix lint?

server/api/impl/proxy/bindings.go
larkox: My fear is that any error here will fail on each binding get, and therefore we will have one line of error every time a user in the system changes channel, which will create lots of messages, even for debug level.

server/api/impl/proxy/bindings.go
```diff
@@ -33,33 +31,51 @@ func mergeBindings(bb1, bb2 []*apps.Binding) []*apps.Binding {
...
 }
 
-func (p *Proxy) GetBindings(cc *apps.Context) ([]*apps.Binding, error) {
+func (p *Proxy) GetBindings(debugSessionToken apps.SessionToken, cc *apps.Context) ([]*apps.Binding, error) {
 	allApps := p.store.App().GetAll()
 
 	all := []*apps.Binding{}
 	for _, app := range allApps {
-		appID := app.Manifest.AppID
+		manifest, err := p.store.Manifest().Get(app.AppID)
```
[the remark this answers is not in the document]
larkox: I just realized this is already "done" in a different place, but your concern is still valid.
@levb thoughts on this?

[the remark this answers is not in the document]
levb: I think my PR fixes this so that the Manifest store is not used at "runtime" of an App. So I do not have any concerns here, temporarily as it is.

server/api/impl/proxy/bindings.go
levb: ticket it? We need some feedback to the developer... `/apps debug-bindings` maaaybe good enough though.

catalintomai (approved):

PR #99, opened 2021-03-11
Remove type from call and consider type on path for helloapp

larkox:
#### Summary
Remove type from call and consider type on path for helloapp

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-33178

#### Related Pull Requests
Redux: https://github.com/mattermost/mattermost-redux/pull/1406
Webapp: https://github.com/mattermost/mattermost-webapp/pull/7679
Mobile: https://github.com/mattermost/mattermost-mobile/pull/5221


server/examples/go/hello/http_hello/functions.go
```diff
@@ -19,9 +20,6 @@ func (h *helloapp) GetBindings(w http.ResponseWriter, req *http.Request, claims
 }
 
 func (h *helloapp) Install(w http.ResponseWriter, req *http.Request, claims *api.JWTClaims, c *apps.CallRequest) (int, error) {
-	if c.Type != apps.CallTypeSubmit {
```
catalintomai, line 22: do we have to remove this? Has anything changed?

catalintomai (approved):

server/examples/go/hello/http_hello/functions.go
larkox: Since calls don't have type anymore, there is no need to check the type here. Since install is called from the server, it does not go under a subpath (e.g. `/install/submit`) but directly to the endpoint.

Does it make sense? Or should all calls (including bindings and install) go to the subpath?

@levb thoughts?

levb (approved):

PR #103, opened 2021-03-15
Restructure package layout

levb:
- Eliminated /server/api, moved packages to the top level
- Implemented `GetStatic` in the `Upstream` interface
- Combined `Admin` and `Proxy` services into 1 `Proxy`
- Combined `upstream`'s `OneWay` and `Roundtrip` into `Roundtrip(..., async bool)`


server/store/apps.go
```diff
@@ -12,23 +12,33 @@ import (
...
+	InitBuiltin(...*apps.App)
+	Save(app *apps.App) error
+}
+
 // appStore combines installed and builtin Apps.  The installed Apps are stored
 // in KV store, and the list of their keys is stored in the config, as a map of
 // AppID->sha1(App).
 type appStore struct {
-	*Store
+	*Service
```
catalintomai, line 33: 1/5 - find more readable `Store` than `Service` in a storage context ( we use `Service all across now and it becomes harder to follow).

server/store/subscriptions.go
```diff
@@ -7,21 +7,22 @@ import (
...
 
 	"github.com/mattermost/mattermost-plugin-apps/apps"
-	"github.com/mattermost/mattermost-plugin-apps/server/api"
+	"github.com/mattermost/mattermost-plugin-apps/server/config"
 	"github.com/mattermost/mattermost-plugin-apps/server/utils"
 )
 
-type SubStore struct {
-	*Store
+type Subscription interface {
```
catalintomai, line 14: 2/5 - why not keep the previous SubscriptionStore, AppStore, etc - find it more readable that reusing the entity names for interface names as well.

catalintomai (approved):

larkox (approved):

server/store/subscriptions.go
levb: Will change back. Note that the pointers to service interfaces are usually named after the "substance", i.e. `proxy proxy.Service` and then `proxy.Call` etc.

levb: @catalintomai is it better now?

catalintomai: Thanks @levb, I just find them more readable.

levb:
@larkox @mickmister TestBindings is flaky, I couldn't see why. Thought it might have been because of the map order randomness, but it failed again once I fixed the order. And then succeeded on a re-run. Please take a look, maybe you see something I don't?

https://app.circleci.com/pipelines/github/mattermost/mattermost-plugin-apps/740/workflows/a1ec41f6-705f-456f-a0fe-4b8ab900b608/jobs/2407

PR #105, opened 2021-03-17
Fix expand, uninstall and send survey

larkox:
#### Summary
Several errors breaking the app, and fix the send survey behaviour for http hello

#### Ticket Link
None


larkox:
For some reason I haven't find out yet, the uninstall flow still does not finish correctly when trying to delete the app. It is specially weird because when the function starts, it gets the app with the Get function, but at the end is not able to find it doing practically the same operation.

server/api/impl/proxy/expand.go
```diff
@@ -156,7 +159,10 @@ func stripUser(user *model.User, level apps.ExpandLevel) *model.User {
 }
 
 func stripChannel(channel *model.Channel, level apps.ExpandLevel) *model.Channel {
-	if channel == nil || level == apps.ExpandAll {
+	if channel == nil {
```
catalintomai, line 162: do we need this split?

server/api/impl/admin/install.go
```diff
@@ -48,7 +48,7 @@ func (adm *Admin) InstallApp(cc *apps.Context, sessionToken apps.SessionToken, i
 	client := model.NewAPIv4Client(conf.MattermostSiteURL)
 	client.SetToken(string(sessionToken))
 
-	if app.GrantedPermissions.Contains(apps.PermissionActAsBot) && app.BotUserID == "" {
```
levb, line 51: I thought this was protecting from double-registering the bot account and the OAuth app? Or were these checks redundant? OT: We need an E2E test for Install/Uninstall. It's time.

server/api/impl/admin/install.go
```diff
@@ -61,7 +61,7 @@ func (adm *Admin) InstallApp(cc *apps.Context, sessionToken apps.SessionToken, i
 		app.BotAccessToken = token.Token
 	}
 
-	if app.GrantedPermissions.Contains(apps.PermissionActAsUser) && app.OAuth2ClientID == "" {
```
levb, line 64: ^^

server/api/impl/proxy/expand.go
```diff
@@ -128,7 +128,10 @@ func (e *expander) ExpandForApp(app *apps.App, expand *apps.Expand) (*apps.Conte
 }
 
 func stripUser(user *model.User, level apps.ExpandLevel) *model.User {
-	if user == nil || level == apps.ExpandAll {
+	if user == nil {
```
levb, line 131: 👍

server/examples/go/hello/send_survey.go
```diff
@@ -128,14 +127,15 @@ func NewSendSurveyPartialFormResponse(c *apps.CallRequest, callType apps.CallTyp
...
 	bot := mmclient.AsBot(c.Context)
-	userID := c.GetValue(fieldUserID, c.Context.ActingUserID)
-
-	// TODO this should be done with expanding mentions, make a ticket
-	if strings.HasPrefix(userID, "@") {
-		user, _ := bot.GetUserByUsername(userID[1:], "")
-		if user != nil {
-			userID = user.Id
+	userID := c.Context.ActingUserID
+	if c.Values[fieldUserID] != nil {
```
levb, line 131: ouch, we need a helper function here (later is fine)

levb (approved):

server/api/impl/proxy/expand.go
larkox: I prefer having it for parallelism with the other functions, and not to make the same mistake in the future if we want to sanitize the channel somehow.

server/api/impl/admin/install.go
larkox: The part that takes care of the double-registering is the "ensure" function. With the previous approach you were considering that the previous value was always an existing app, and will not create a new one. In that case, if the admin deleted by mistake the oauth app, and wanted to fix it reinstallin, it will not work, because the app id (or bot id) exist in the app, even if it does not exist in the system.

larkox:
Found and fixed the final bug with uninstall. Also added the websocket event to update the bindings for the person invoking the uninstall command.

catalintomai (approved):

server/api/impl/admin/install.go
catalintomai: > I thought this was protecting from double-registering the bot account and the OAuth app? Or were these checks redundant? OT: We need an E2E test for Install/Uninstall. It's time.

cc: @iomodo

PR #112, opened 2021-03-24
Update interactivity docs and helloworld app

mickmister:
#### Summary

This PR includes updates for documentation on Call/CallRequest structure changes and call type URL changes. It also includes an error response, and a field-specific error response.

It also adds some fields to the helloworld app's form, to match what's being used in the docs. 0/5 on what fields this app is supposed to support. Maybe it should be kept simple as it was with one field.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-30273
https://mattermost.atlassian.net/browse/MM-34160

#### Rendered

[docs/02-example-post-menu.md](https://github.com/mattermost/mattermost-plugin-apps/blob/hello-docs/docs/02-example-post-menu.md)
[docs/03-example-interactivity.md](https://github.com/mattermost/mattermost-plugin-apps/blob/hello-docs/docs/03-example-interactivity.md)


larkox (approved):

levb (changes requested):
The idea of the hello, world app is to keep it extremely simple, so please remove all error handling? Also, it is on purpose that its functionality is limited, why add Post Menu?

levb:
2/5 we should try to have more simple, targeted "Hello" apps rather than bundle more functionality into 1.

mickmister:
@levb I was mainly just committing the changes I used to generate the data used in the docs, but we can use a different app for that, like the test app https://github.com/larkox/mattermost-app-test . I agree that we should keep the hello app as simple as possible, like the plugin starter template.

hanzei:
@levb What is needed in terms of docs here?

levb (approved):

levb:
@hanzei only a review/approval by @justinegeffen

mickmister:
@levb @justinegeffen Pretty much all the changes in this PR are just changing the JSON data in the docs. 3/5 I think we should merge this as-is, as I think it's high impact for external documentation for other devs. @justinegeffen Please leave comments for anything you think may need to change, and I can change in a followup PR as necessary.

PR #113, opened 2021-03-24
MM-31118: 3rd party OAuth2 support

levb:
#### Ticket
https://mattermost.atlassian.net/browse/MM-31118

image

To demo:
- Follow [Google Calendar Plugin README](https://github.com/mattermost/mattermost-plugin-google-calendar/tree/30c9e3dbe2a6e4a81ab4217666e7dcd03ffe12e8#installing) to set up your OAuth2 app on the Google side.
- In a separate window, run `go run ./server/examples/go/hello-3rdparty`
- In Mattermost,
  - `/apps debug-add-manifest --url http://localhost:8080/manifest.json`
  - `/apps install --app-id hello3rdparty` (any secret)
  - `/hello3rdparty configure --client-secret VALUE --client-id VALUE`
  - `/hello3rdparty connect`
  - `/hello3rdparty send`

#### Breaking changes
- static should be invoked as `.../apps/APPID/static/NAME` (was something else)
- KV schema changed, prior data should be wiped out

#### Changes
- Added `http/gateway` that supports `.../apps/AppID/(static|oauth2)` paths, moved the static handler there
- Added `hello3rdparty` example App
- Removed `mock_config` - `config.TestService` is a simpler functional equivalent.
- Added `AppPath` to `Context`, re-factored the new `Context` creation.
- Simplified initialization of mmclient
- shortened internal key names for KV REST API, see https://github.com/mattermost/mattermost-plugin-apps/pull/113/files?file-filters%5B%5D=.go&file-filters%5B%5D=.json&file-filters%5B%5D=.md&file-filters%5B%5D=.png&file-filters%5B%5D=.sum#diff-a3021be373b09ef7dfe9949ad06ae9d2fc9c4a47c6afb7c415e50f9203a6dfd8L26

#### Next
- Add 3rd party OAuth as a `Permission`
- Add App's OAuth paths to `ExpandedContext` 
- Add a Config/Token store REST API (instead of using KV awkwardly)
- Add 3rd party OAuth config/tokens to `ExpandedContext`


larkox (approved):
LGTM

server/config/constants.go
```diff
@@ -31,8 +31,9 @@ const (
 	AppsPath = "/apps"
 
 	// OAuth2 sub-paths.
-	PathOAuth2         = "/oauth2"          // convention for Mattermost Apps, comes from OAuther
-	PathOAuth2Complete = "/oauth2/complete" // convention for Mattermost Apps, comes from OAuther
+	PathOAuth2             = "/oauth2"
+	PathMattermostRedirect = "/mattermost/redirect"
```
catalintomai, line 35: do we need this const?

apps/call.go
```diff
@@ -58,6 +58,7 @@ type CallRequest struct {
 	Values map[string]interface{} `json:"values,omitempty"`
```
catalintomai, line 58: Is the corresponding ticket: https://mattermost.atlassian.net/browse/MM-31118 ?

server/http/gateway/gateway.go
```diff
@@ -0,0 +1,41 @@
...
+	pluginapi "github.com/mattermost/mattermost-plugin-api"
+
+	"github.com/mattermost/mattermost-plugin-apps/apps"
+	"github.com/mattermost/mattermost-plugin-apps/server/appservices"
+	"github.com/mattermost/mattermost-plugin-apps/server/config"
+	"github.com/mattermost/mattermost-plugin-apps/server/proxy"
+	"github.com/mattermost/mattermost-plugin-apps/server/utils/httputils"
+)
+
+type gateway struct {
```
catalintomai, line 17: J/C - why carving out the `gateway` package from the `restapi` one?

server/config/constants.go
levb: I still need to sort out OAuth2 for Mattermost, and what callback (and redirect?) URLs should be reserved for those. The callback path is being registered and stored persistently in Mattermost when we create an OAuth2 Application for the App. The redirect const here is a reminder.

apps/call.go
levb: yes.

server/http/gateway/gateway.go
levb: This is a great question, I struggled with it myself.

On the one hand, I am very tempted to roll all the `http/*` packages into 1, esp. considering that "dialog" is a goner soon. On the other hand, there are 2 distinct buckets - proxying App interactions (gateway) and REST APIs, mostly for the Apps. There is some clarity in separation. IMO the `Call` API should become something like a POST to a `.../apps/AppID[/call]/PATH` and we may deprecate the `Call` REST API.

Long story short, 0/5 we either roll all http into 1 package (like we did proxy/admin), or we keep the 3 sub-packages for clarity?

levb: If we roll them, 1/5 a separate PR?

levb:
@catalintomai @larkox see the Next section in the updated description. I'd like to finish it as a P1, but as a separate PR.

catalintomai (approved):

PR #114, opened 2021-03-24
[MM-32599] Implement webhook handler for App Framework

jfrerich:
#### Demo:
- Run the app as `go run ./server/examples/go/hello-webhooks`
- Navigate to a Channel in Mattermost that you want to demo in (post webhooks to) and run:
```
/apps debug-add-manifest --url http://localhost:8080/manifest.json
/apps install --app-id hello-webhooks
/hello-webhooks info 
```
- The last command displays the next command to run, should look like
```
/hello-webhooks send https://levb.ngrok.io/plugins/com.mattermost.apps/apps/hello-webhooks/webhook/hello?secret=XYZ
```
- Running the command posts a webhook message, that prints into the channel

#### Summary
- added [http-webhook](./server/examples/go/http-webhook) example.
- webhook HTTP endpoint: {PluginURL}/apps/{AppID}/webhook/{Path}?secret=XYZ
- webhook Call Path: /webhook/{Path}
- webhook size is limited to the configured file attachment size, default 75Mb
- Added a new `PermissionRemoteWebhooks` 
- Added `WebhookSecret` to `App`, visible in `ExpandedContext` when `App` is expanded. It is automatically generated on install, if the app's permissions contain "remote_webhooks".
- pass payload inside the appCall as values["data"]

### TODO - Create Tickets
- [x] Update documentation (https://mattermost.atlassian.net/browse/MM-34622)
- [x] add `/apps webhook reset-secret AppID` command (https://mattermost.atlassian.net/browse/MM-34623)
- [x] add tests (https://mattermost.atlassian.net/browse/MM-34624)
- [x] add JWT validation - (https://mattermost.atlassian.net/browse/MM-34625)

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-32599


server/command/install.go
```diff
@@ -16,9 +16,11 @@ import (
 func (s *service) executeInstall(params *params) (*model.CommandResponse, error) {
 	appID := ""
 	appSecret := ""
+	appWebhookSecret := ""
 	fs := pflag.NewFlagSet("", pflag.ContinueOnError)
 	fs.StringVar(&appID, "app-id", "", "ID of the app")
 	fs.StringVar(&appSecret, "app-secret", "", "App secret")
+	fs.StringVar(&appWebhookSecret, "app-webhook-secret", "", "App webhook secret")
```
catalintomai, line 23: do we have to use a secret different from the app secret? (or can we reuse it if not provided in the command)?

server/http/restapi/webhook.go
```diff
@@ -0,0 +1,68 @@
...
+		},
+		Type: apps.CallTypeSubmit,
+		Call: apps.Call{
+			Path: "/webhook-incoming",
+		},
+	}
+	_ = a.proxy.Call("", &call)
+}
+
+func (a *restapi) validSecret(appID, secret string) bool {
```
catalintomai, line 64: s/`validSecret`/`isValidSecret` ?

server/http/restapi/webhook.go
```diff
@@ -0,0 +1,68 @@
...
+	values["payload"] = c
+
+	call := apps.CallRequest{
+		Values: values,
+		Context: &apps.Context{
+			AppID: apps.AppID(appID),
+		},
+		Type: apps.CallTypeSubmit,
+		Call: apps.Call{
+			Path: "/webhook-incoming",
```
larkox, line 58: Should we allow custom webhook calls like we do with Bindings and Install?

server/http/restapi/webhook.go
```diff
@@ -0,0 +1,68 @@
...
+	}
+
+	secret := queryVars["secret"][0]
+	if !a.validSecret(appID, secret) {
+		httputils.WriteBadRequestError(w, errors.New("webhook secret is not valid"))
+		return
+	}
+
+	var c interface{}
+	err := json.NewDecoder(req.Body).Decode(&c)
```
larkox, line 42: Probably an edge case, but what if the body is not JSON coded? Should we send the req.Body in a more "raw" way?

server/http/restapi/webhook.go
```diff
@@ -0,0 +1,68 @@
...
+	if err != nil {
+		httputils.WriteBadRequestError(w, err)
+		return
+	}
+
+	values := make(map[string]interface{})
+	values["payload"] = c
```
larkox, line 49: We probably should add also the query values (minus the secret).

server/command/install.go
larkox, line 23: Should the webhook secret be autogenerated from mattermost? If I am not mistaken, expanding the app will send also the webhook secret with the call, so if the app needs it, it can take it from there.

larkox (changes requested):
A few potential improvements.

server/http/restapi/webhook.go
jfrerich: done.

server/command/install.go
jfrerich: Good question.  We could use the same secret.  I'm not sure if he have used that implementation in another plugin

server/http/restapi/webhook.go
jfrerich: I can certainly add that, but @levb and I discussed getting this PR working with Zendesk and abstracting pieces later.  We should probably make sure that we agree with the `default` path name `webhook-incoming` for this PR

jfrerich: What do y'all think?

server/http/restapi/webhook.go
jfrerich: Do you have an idea in mind? I assume there is a method available to check if `req.Body` is valid `json`?  If `req.Body` is not `json`, send as a raw `string`?

server/command/install.go
jfrerich: I can generate the secret automatically. @levb and I discussed this also.  I like the idea, but we scoped the PR as this simple implementation where the secret is provided by the installer.  

If autogenerated, that would be great.  We would need to provide a method for retrieving the webhook secret (or better yet the link for the endpoint).  Maybe the zendesk DM after install?  

What if the user wants to change the webhook secret? I think that is only available by doing another `install`

jfrerich: Ideally, I'd like to provide a `/zendesk setup-webhook` command to go and create the webhook without having to go into `zendesk` itself and manually create

server/http/restapi/webhook.go
larkox: The only thing that comes to mind is that anyone is out there using XML instead of JSON for encoding stuff. Haven't seen it on the tools we have used, but who knows...

server/command/install.go
hanzei: What is a use case for setting up  a new webhook secret? Would a slash command like `/zendesk setup-webhook --force` to re-setup the webhook be fine?

hanzei: Should we fore now generate the secret and output it after the install process for the admin?

server/http/dialog/install.go
```diff
@@ -44,6 +44,15 @@ func NewInstallAppDialog(m *apps.Manifest, secret, pluginURL string, commandArgs
 			HelpText:    "TODO: How to obtain the App Secret",
 			Default:     secret,
 		})
+		elements = append(elements, model.DialogElement{
+			DisplayName: "App webhook secret:",
```
hanzei, line 48: Don't we need that for AWS apps?

server/http/restapi/webhook.go
```diff
@@ -0,0 +1,67 @@
...
```
hanzei, line 48: Do we need any headers?

server/command/install.go
larkox: The reason in general for regenerating secrets is that it has been compromised, and has to be reset for security reasons. (Or security policies to reset every X time). I think it is a needed feature down the line, but 0/5 on how urgent it is.

levb: 1/5 keep it manually entered for now, maybe pre-initialize on install (use App.Secret?). Then have a special command to change it.

We can always migrate common admin commands to the framework over time, so `/zendesk webhook secret` may eventually become `/apps webhook secret --app zendesk`. Anyway, simpler is better for now.

server/http/dialog/install.go
levb: we do

server/http/restapi/restapi.go
```diff
@@ -43,7 +43,7 @@ func Init(router *mux.Router, mm *pluginapi.Client, conf config.Service, proxy p
 	subrouter.HandleFunc(config.KVPath+"/", a.handleKV(a.kvList)).Methods("GET")
 	subrouter.HandleFunc(config.KVPath+"/{key}", a.handleKV(a.kvHead)).Methods("HEAD")
 	subrouter.HandleFunc(config.KVPath+"/{key}", a.handleKV(a.kvDelete)).Methods("DELETE")
-
+	subrouter.HandleFunc(config.Webhook+"/{app_id}/{path}", a.handleWebhook).Methods("POST")
```
levb, line 46: this should all go into the `gateway` package for now, with the route `[plugin url]/apps/appid/webhooks`

server/http/restapi/webhook.go
levb: Why do we ever need to decode the body? Why not just pass through the bytes?

server/http/restapi/webhook.go
levb: We could do this:
1. read in req.Body as []byte
2. If content-type==application/json, parse it and inject it into "values" as such (AT THE TOP LEVEL, aka values = object)
3. pass the HTTP request values (all: URL, header) in the context (`Context.Request`?)

server/http/restapi/webhook.go
levb: see my response ^^ https://github.com/mattermost/mattermost-plugin-apps/pull/114/files#r601442953

server/command/install.go
levb: As @hanzei pointed out, AppSecret is specific to the HTTP apps. I like @larkox's suggestion of a generated secret, but it's more code; I'm ok with a manually entered, separate secret for now.

server/http/restapi/webhook.go
levb: For the App, all webhooks are "incoming", whether they come from Mattermost or a 3rd party system.

2/5 we refer to 3rd party systems as "remote (3rd party)" and in the code/config just "remote".

levb: Customizing the Expand of the webhook call is going to be essential in the future, but ATM would create difficulties since Expand does not enforce bot permissions.

It may be easier if less performant to use the REST API in webhook handlers.

Webhook handlers should be invoked with a `Notify`, async.

server/http/restapi/webhook.go
jfrerich: @levb, I'm not quite sure that I understand the comment.  Right now I decode the payload into `c`, the empty interface variable. That value is the then used as part of the larger `CallResponse` object

server/http/restapi/webhook.go
jfrerich: Do you want me to create a new `Context.Request` of type `interface{}` and just fill it with the entire request?  Then the App would handle webhook the request however it wants?  That seem pretty flexible, and would remove any complexity placed on the proxy app.

larkox: I am not sure whether it should go in the context or as a value. I feel that the Context.Request would only appear on webhooks, and therefore it does not feel so much as a call context, but as a value.

server/http/dialog/install.go
levb: The Webhook Secret is now auto-generated if the App needs one.

server/command/install.go
levb: The secret is generated on install now

server/http/restapi/webhook.go
levb: All the HTTP stuff can go in as Values later, if needed (Query, Header)

levb (approved):

catalintomai (approved):

server/http/dialog/install.go
```diff
@@ -25,7 +26,7 @@ func NewInstallAppDialog(m *apps.Manifest, secret, pluginURL string, commandArgs
 	intro := md.Bold(
 		md.Markdownf("Application %s requires the following permissions:", m.DisplayName)) + "\n"
 	for _, permission := range m.RequestedPermissions {
-		intro += md.Markdownf("- %s\n", permission.Markdown())
+		intro += md.Markdownf("- %s\n", permission)
```
hanzei, line 29: Nice one :+1:

server/http/gateway/oauth2.go
```diff
@@ -3,23 +3,17 @@ package gateway
...
 	"github.com/mattermost/mattermost-plugin-apps/server/utils"
 	"github.com/mattermost/mattermost-plugin-apps/server/utils/httputils"
 )
 
 func (g *gateway) remoteOAuth2Connect(w http.ResponseWriter, req *http.Request, sessionID, actingUserID string) {
-	vars := mux.Vars(req)
-
-	appID := vars["app_id"]
-	if appID == "" {
+	if appIDVar(req) == "" {
```
hanzei, line 11: Nit pick:
```suggestion
	appID := appIDVar(req) 
	if appID == "" {
```

server/http/gateway/webhook.go
```diff
@@ -0,0 +1,49 @@
...
+	}
+	app, err := g.proxy.GetInstalledApp(appID)
+	if err != nil {
+		httputils.WriteError(w, err)
+		return
+	}
+	if secret != app.WebhookSecret {
+		httputils.WriteError(w, utils.NewInvalidError("webhook secret mismatched"))
+		return
+	}
```
hanzei, line 32: I quite like this solution. Simple and enough for our needs right now.

server/proxy/install.go
```diff
@@ -172,7 +175,6 @@ func (p *Proxy) ensureBot(manifest *apps.Manifest, actingUserID string, client *
 			}
 		}
 	}
-
```
hanzei, line 175: Nit pick: Why drop this line?

apps/manifest.go
```diff
@@ -12,6 +12,12 @@ import (
 // Where static assets are.
 const StaticFolder = "static"
 
+// Root Call path for incoming webhooks from remote (3rd party) systems. Each
+// webhook URL should be in the form:
+// "{PluginURL}/apps/{AppID}/webhook/{PATH}...?secret=XYZ", and it will invoke a
+// Call with "/webhook/{PATH}"."
+const PathWebhook = "/webhook"
```
hanzei, line 19: Should we make this configuration like `Bindings`? Maybe we need a followup ticket.

server/proxy/proxy.go
```diff
@@ -86,6 +88,45 @@ func (p *Proxy) Notify(cc *apps.Context, subj apps.Subject) error {
...
+	creq := &apps.CallRequest{
+		Call: apps.Call{
+			Path: path.Join(apps.PathWebhook, webhookPath),
+		},
+		Context: p.conf.GetConfig().SetContextDefaultsForApp(app.AppID, &apps.Context{
+			ActingUserID: app.BotUserID,
+		}),
+		Values: map[string]interface{}{
+			"data": datav,
+			"path": webhookPath,
```
hanzei, line 118: Do we really need to put the `path` into `Values`? `Call.Path` does contain it anyway. 1/5 that we should not duplicate information in a `Call`?

server/proxy/proxy.go
```diff
@@ -86,6 +88,45 @@ func (p *Proxy) Notify(cc *apps.Context, subj apps.Subject) error {
...
+	if err != nil {
+		return err
+	}
+
+	var datav interface{}
+	err = json.Unmarshal(data, &datav)
+	if err != nil {
+		// if the data can not be decoded as JSON, send it "as is", as a string.
+		datav = string(data)
+	}
```
hanzei, line 106: What value does that unmarshaling the JSON add? Does it make the developers life easier?

server/proxy/proxy.go
```diff
@@ -86,6 +88,45 @@ func (p *Proxy) Notify(cc *apps.Context, subj apps.Subject) error {
...
+
+	up, err := p.upstreamForApp(app)
+	if err != nil {
+		return err
+	}
+
+	var datav interface{}
+	err = json.Unmarshal(data, &datav)
+	if err != nil {
+		// if the data can not be decoded as JSON, send it "as is", as a string.
```
hanzei, line 104: Should we document that behaviour?

server/http/gateway/webhook.go
```diff
@@ -0,0 +1,49 @@
...
+	if secret == "" {
+		httputils.WriteError(w, utils.NewInvalidError("webhook secret was not provided"))
+		return
+	}
+	app, err := g.proxy.GetInstalledApp(appID)
+	if err != nil {
+		httputils.WriteError(w, err)
+		return
+	}
+	if secret != app.WebhookSecret {
```
hanzei, line 29: Should we make use of https://golang.org/pkg/crypto/subtle/ here?

hanzei (approved):
Nice, I quite like the secret handling :+1: 

LGTM, a few nit picks below

larkox (approved):
LGTM! Great work!

server/http/gateway/gateway.go
```diff
@@ -30,12 +30,24 @@ func Init(router *mux.Router, mm *pluginapi.Client, conf config.Service, proxy p
 	subrouter := router.PathPrefix(config.PathApps).Subrouter()
 
 	// Static
-	subrouter.HandleFunc("/{app_id}/"+apps.StaticFolder+"/{name}",
-		httputils.CheckAuthorized(mm, p.static)).Methods(http.MethodGet)
+	subrouter.HandleFunc("/{appid}/"+apps.StaticFolder+"/{name}",
+		httputils.CheckAuthorized(mm, g.static)).Methods(http.MethodGet)
+
+	// Incoming remote webhooks
```
levb, line 36: @jfrerich @larkox double-check in Zendesk/SN

jfrerich:
@lev Tickets have been created for the TODO items in the PR description. Merging now

apps/manifest.go
levb: @hanzei yes we should file a ticket, but I intentionally did not since I think we should figure out a greater scheme of attaching "Expand" and "State" to calls.

hanzei:
Documented in https://developers.mattermost.com/integrate/apps/functionality/external-webhooks/

PR #115, opened 2021-03-25
[MM-32330] E2E webapp tests

mickmister:
#### Summary

This PR adds a test app that can have its responses controlled via http endpoints that the cypress test calls. The cypress test can then decide what the app should return, and run assertions on the request that the app received. I tried to have the cypress test itself host an express server to act as the app, but the cypress code runs in the browser and cannot act as a server.

- [e2e/mockappserver/routes.go](https://github.com/mattermost/mattermost-plugin-apps/blob/e2e-webapp-tests/e2e/mockappserver/routes.go)
- [e2e/cypress/integration/bindings/channel_header_spec.ts]( https://github.com/mattermost/mattermost-plugin-apps/blob/e2e-webapp-tests/e2e/cypress/integration/bindings/channel_header_spec.ts)

Some test setup steps still need to be done, like ensuring the proxy plugin is installed, and ensuring the test/mock app is installed.

#### Ticket Link

Fixes https://mattermost.atlassian.net/browse/MM-32330


larkox (approved):

hanzei:
Is there a followup ticket to add webapp E2E tests to CI?

levb (approved):
fix lint, please

e2e/package.json
```diff
@@ -0,0 +1,51 @@
...
+    "mochawesome-merge": "4.1.0",
+    "mochawesome-report-generator": "5.1.0",
+    "moment-timezone": "0.5.31",
+    "mysql": "2.18.1",
+    "path": "0.12.7",
+    "pg": "8.3.3",
+    "recursive-readdir": "2.2.2",
+    "shelljs": "0.8.4",
+    "uuid": "8.3.0",
+    "yargs": "16.0.3"
```
hanzei, line 40: Do we need /all/ of these just to run the cypress tests? Or can we rely on `cypress` to bring what it needs?

e2e/package.json
```diff
@@ -0,0 +1,51 @@
...
+  },
+  "scripts": {
+    "cypress:open": "cross-env TZ=Etc/UTC cypress open",
+    "cypress:run": "cross-env TZ=Etc/UTC cypress run --browser chrome",
+    "cypress:run:firefox": "cross-env TZ=Etc/UTC cypress run --browser firefox",
+    "cypress:run:edge": "cross-env TZ=Etc/UTC cypress run --browser edge",
+    "cypress:run:electron": "cross-env TZ=Etc/UTC cypress run --browser electron",
```
hanzei, line 47: `cross-env` is in maintenance mode(https://www.npmjs.com/package/cross-env). Given that all staff devs are using linux/macOS, can we drop `cross-env`?

e2e/package.json
```diff
@@ -0,0 +1,51 @@
...
+    "path": "0.12.7",
+    "pg": "8.3.3",
+    "recursive-readdir": "2.2.2",
+    "shelljs": "0.8.4",
+    "uuid": "8.3.0",
+    "yargs": "16.0.3"
+  },
+  "scripts": {
+    "cypress:open": "cross-env TZ=Etc/UTC cypress open",
+    "cypress:run": "cross-env TZ=Etc/UTC cypress run --browser chrome",
```
hanzei, line 44: 0/5 should we simplify to `open`, `run` etc. similar as in https://github.com/mattermost/mattermost-plugin-incident-collaboration/blob/master/tests-e2e/package.json?

e2e/package.json
```diff
@@ -0,0 +1,51 @@
...
+    "cypress:run:firefox": "cross-env TZ=Etc/UTC cypress run --browser firefox",
+    "cypress:run:edge": "cross-env TZ=Etc/UTC cypress run --browser edge",
+    "cypress:run:electron": "cross-env TZ=Etc/UTC cypress run --browser electron",
+    "test:ci": "node run_tests.js",
```
hanzei, line 48: That command errors with
```
$ npm run test:ci 

> test:ci
> node run_tests.js

node:internal/modules/cjs/loader:927
  throw err;
  ^

Error: Cannot find module '/home/sumacheb/src/mattermost/plugins/apps/e2e/run_tests.js'
    at Function.Module._resolveFilename (node:internal/modules/cjs/loader:924:15)
    at Function.Module._load (node:internal/modules/cjs/loader:769:27)
    at Function.executeUserEntryPoint [as runMain] (node:internal/modules/run_main:76:12)
    at node:internal/main/run_main_module:17:47 {
  code: 'MODULE_NOT_FOUND',
  requireStack: []
}
npm ERR! code 1
npm ERR! path /home/sumacheb/src/mattermost/plugins/apps/e2e
npm ERR! command failed
npm ERR! command sh -c node run_tests.js

npm ERR! A complete log of this run can be found in:
npm ERR!     /home/sumacheb/.npm/_logs/2021-03-30T09_24_48_017Z-debug.log

```

e2e/package.json
```diff
@@ -0,0 +1,51 @@
...
+    "uniq-meta": "grep -r \"^// $META:\" cypress | grep -ow '@\\w*' | sort | uniq"
```
hanzei, line 49: Do we need that command?

e2e/cypress/integration/bindings/channel_header_spec.ts
```diff
@@ -0,0 +1,192 @@
...
+    ]);
+}
+
+const setBindings = (bindings) => {
+    const data = {
+        type: 'ok',
+        data: bindings,
+    };
+
+    return cy.request('POST', 'http://localhost:8065/plugins/com.mattermost.apps/e2e-testapp/bindings/set-response', data);
```
hanzei, line 175: Should we use variables for the siteURL and the plugin id?

e2e/package.json
```diff
@@ -0,0 +1,51 @@
...
+  }
```
hanzei, line 50: Do we need a `.eslintrc`?

hanzei (changes requested):
Great job!

I commented on a few things that stood out to me

e2e/package.json
mickmister: Done

e2e/package.json
mickmister: Removed this npm command as it has nothing to do with this project

e2e/package.json
mickmister: Nope

e2e/package.json
mickmister: Sure, added one

mickmister:
@hanzei I've replied to most of your comments. I haven't added any further commits yet, but I've resolved locally what I said I've resolved

hanzei:
Awesome, thanks @mickmister! Looking forward to reviewing the commits once they are pushed.

e2e/package.json
mickmister: @hanzei I added these dependencies back in, as they are needed to import `'mattermost-webapp/e2e/cypress/support.ts`. This is the same setup as the demo plugin. We could go the route of copying functions from the webapp repo, but I think this way is better as all of the utility functions are defined in one place.

I'm not sure what the best way to do this is, as we are adding a lot of "unnecessary" dependencies here.

server/examples/go/hello/send_survey.go
```diff
@@ -62,7 +62,6 @@ func NewSendSurveyFormResponse(c *apps.CallRequest) *apps.CallResponse {
 					Name:                 fieldUserID,
 					Type:                 apps.FieldTypeUser,
 					Description:          "User to send the survey to",
-					IsRequired:           true,
```
hanzei, line 65: Is this changes intended?

server/examples/go/hello/send_survey.go
```diff
@@ -129,14 +128,9 @@ func NewSendSurveyPartialFormResponse(c *apps.CallRequest, callType apps.CallTyp
...
-	if c.Values[fieldUserID] != nil {
-		option := apps.SelectOption{}
-		b, _ := json.Marshal(c.Values[fieldUserID])
-		err := json.Unmarshal(b, &option)
-		if err != nil {
-			return "", err
-		}
-		userID = option.Value
+	submission := extractSurveyFormValues(c)
+	if submission.UserID.Value != "" {
```
hanzei, line 132: Isn't a userID required?

server/plugin.go
```diff
@@ -99,9 +97,6 @@ func (p *Plugin) OnActivate() error {
 	}
 	// app store
 	appstore := p.store.App
-	if pluginapi.IsConfiguredForDevelopment(mmconf) {
-		appstore.InitBuiltin(builtin_hello.App())
-	}
```
hanzei, line 104: Why isn't this needed any longer?

hanzei (changes requested):

server/plugin.go
mickmister: @hanzei The built-in app does not have a bot, and so the features that require a bot (submitting any form) do not work, so it is sort of useless to have there at its current state. The function that creates the bot (below) requires a user session token, which is not available upon plugin startup, since there is no user interaction going on.

https://github.com/mattermost/mattermost-plugin-apps/blob/02428ac9dd1423231989d1d56fdecfe7d43da1bf/server/proxy/install.go#L139

There were a few options to choose from, in order to write an e2e test that performs some app's action. Hosting the test app or the `helloworld` app that runs in a separate port/process was not an option. Here some some options:

- Have the built-in app create its own bot on init. Since there is no user session token, we would be doing so outside of the `proxy.ensureBot` function. In doing so, we would be duplicating code and risking having these pieces of code being out of sync.
- Use the http-hello app, and keep the builtin app installed. This has the issue of the bindings' HTML ids colliding in the webapp. So the e2e test cannot objectively find the right element.
  - One solution to this would be append the app id to the binding locations when serving up the bindings, but this makes the command names change from `send` to `send-http-hello`. I chose not to do this for this reason.
  - Another solution would be to update the webapp to include the app id in the HTML id, as `${app_id}_${location}`. I'm submitting a PR to include this as this should be done anyway.
- Remove the built-in app, and only use the http-hello app in the e2e test. This has the drawback of not having a default app installed. I ultimately chose this option.

There are probably other practical options, but this is what I considered when writing the test

catalintomai:
@mickmister, to confirm wrt to "I tried to have the cypress test itself host an express server to act as the app, but the cypress code runs in the browser and cannot act as a server.", - reusing+extending the one at `.../mattermost-webapp/e2e/webhook_serve.js` does not work, right?

mickmister:
@catalintomai We can use the `webhook_serve.js` file as the foundation for the test app. I'm currently using the hello app, as it is already in place and implements some basic features.

Even if it is trivial to introduce the javascript server as the test app into CI, it would still take some time for us to implement the install flow etc. into the test app. 1/5 that we should introduce the javascript test app after this PR because of this.

catalintomai:
@mickmister , I think it is ok to proceed incrementally. I just believe in aligning to smth. that the QA team owns - makes maintenance easier long-term.

server/plugin.go
hanzei: The lack of a bot account for the built-in app seems like a bug to me. I was expecting the app to have them. `actingUserID` is only used for notification, which is not needed in this case. `ensureBot` could be rewritten to optionally accept a userID.

jfrerich (approved):
Awesome work @mickmister.  Lot's of setup work!

hanzei (approved):

hanzei:
@mickmister and I synced up offline. The behavior changes are as they make E2E testing easier.

PR #117, opened 2021-03-28
MM-31118: further improvements for Remote OAuth

levb:
Further extending on #113 
Ticket: https://mattermost.atlassian.net/browse/MM-31118

### BREAKING CHANGES
- `appservices.KV` prefix is now passed in the path, as in `.../kv/{prefix}/{key}`.
- the apps plugin KV prefix, `state` separator, sub separator is now `.` (to avoid conflicts with base64).
- hashed KV keys are back to `md5` to preserve space in the 50-char key.
cc @jfrerich @DHaussermann @mickmister you may want to `debug-clean` first.

##### OAuh
- Added `RemoteOAuth2 OAuthApp` to `App`.
- Added `GetOAuth2RedirectURL` and `OnOAuth2Complete` to `Manifest`, with defaults.
- Added `OAuth2 OAuth2Context` to `ExtendedContext`.
- Added "oauth2_app", "oauth2_user" to `Expand`.
- Added "remote_oauth2" `Permission`.
- Added `CreateOAuth2State` (user), `StoreOAuth2App` (bot, **should this be sysadmin?**), `StoreOAuth2User` (user), `GetOAuth2User` (user) REST APIs, and corresponding methods in `appservices`. Note that for the time being we need to pass in AppID explicitly; the idea is to derive it from the OAuth session, but that needs more changes, including in `mattermost-server`.
- Added `store.OAuthStore` to support OAuth2 implementation.
- Added `GetRemoteOAuth2RedirectURL` and `CompleteRemoteOAuth2` to `proxy` to keep `http/gateway` to HTTP/JSON-> go mapping.
- `CompleteRemoteOAuth2` validates the state before calling app's `OnOAuth2Complete`.

##### Misc
- eliminated `InInstallApp` struct.
- added `store.AppKVStore`, `appservices` should only concern with validation and business logic.
- eliminated re-encoding of values in the (app) kv PUT `restapi`.
- moved Call validation into `proxy.Call` from `restapi`.
- renamed and adjusted the e2e tests.
- made APIs take `sessionID`, `actingUserID` consistently.
- errors _caused_ by `utils.ErrXXX` are now mapped to HTTP status codes.
- added `Permissions.IsValid` that checks for consistency.
- consolidated REST API paths in `mmclient`.
- refacored, and hopefully fixed some corner-cases in `proxy/expand.go`. Added (most?) app permission checks.
- nits and cosmetics, changed all remaining `<><>` to `<>/<>` to make their look more impactful.

### TODO
- [x] `StoreOAuth2App` should be sysadmin, not bot.
- [x] Update docs
- [x] check all JSON tags
- [x] How do we add to JS API client? cc @catalintomai?
- [ ] Need a `/hello-oauth disconnect`
- [x] _Maybe_ provide `appid /connect` and `/disconnect` Call handlers, unless overwritten by the App?


apps/context.go
```diff
@@ -75,8 +76,12 @@ type ExpandedContext struct {
...
-	PathOAuthComplete = "/oauth2/remote/complete"
-)
+type OAuth2Context struct {
+	// Expanded with "oauth2_app". Config must be previously stored with
+	// mmclient.StoreOAuth2App
+	OAuth2App
+	RedirectURL string `json:"redirect_url,omitempty"`
+	CompleteURL string `json:"complete_url,omitempty"`
+
+	User interface{} `json:"user,omitempty"`
```
hanzei, line 86: What kind of `User` is this?

apps/permissions.go
```diff
@@ -46,8 +51,28 @@ func (p Permission) Markdown() md.MD {
...
 	default:
 		m = "unknown permission: " + string(p)
 	}
 	return md.MD(m)
 }
+
+func (p Permissions) IsValid() error {
+	for _, pp := range []Permissions{
+		{PermissionRemoteOAuth2, PermissionActAsBot, PermissionActAsUser},
+		{PermissionUserJoinedChannelNotification, PermissionActAsBot},
```
hanzei, line 65: What are these groups?

apps/context.go
```diff
@@ -75,8 +76,12 @@ type ExpandedContext struct {
...
-// Paths for OAuth endpoint that the App needs to set up remote (3rd party) OAuth
-const (
-	PathOAuthRedirect = "/oauth2/remote/redirect"
-	PathOAuthComplete = "/oauth2/remote/complete"
-)
+type OAuth2Context struct {
+	// Expanded with "oauth2_app". Config must be previously stored with
+	// mmclient.StoreOAuth2App
+	OAuth2App
+	RedirectURL string `json:"redirect_url,omitempty"`
```
larkox, line 83: Just to be clear. The redirect URL is the place I have to send someone to start the OAuth flow, right?
I post the link to the redirect URL, this goes to a mattermost endpoint that ask the app for the OAuth link to start the flow (called also redirect URL), and while giving that information, I store the state.

The complete URL is the URL where I should redirect the user after the flow is finished, right? When I create the link to start the flow, I should do it with the complete URL as the redirect at the end of the flow. When that endpoint is hit, the proxy will forward to the app the query values, and the app will check the state, create the token, and store the user.

The naming seems a bit weird right now, because there are several "redirect" elements here and there. Any idea how to make it more clear? I feel the redirectURL here in particular could change to something like "StartURL", since what is doing is start the flow.

larkox (approved):
In general looks fine, just a bit potential improvements on naming.

apps/context.go
levb: A generic 3rd party "user" object, up to the App. In hello-oauth it's `oauth2.Token`

apps/permissions.go
levb: (A, B, C, ...) - A requires (depends on) B, C, ...

apps/context.go
levb: I agree. I wish the OAuth2 package (and spec?) used the client-centric term "Complete" for the "final redirect"

`start`? `connect`? and `complete`

larkox: I like the idea of `start` and `connect`, but I am struggling with `complete`.

`start` starts the process on mattermost side
`connect` is the url to the OAuth provider. Mattermost will redirect from start to the connect url.
`complete` is the url to which the OAuth provider will redirect after finishing. That is a good fit for complete naming, but it is also the `redirect url` as per OAuth spec. 0/5 on which direction to go.

levb: Current flow:
1. "connect link" HTTP `/apps/appid/oauth2/remote/redirect` (A)
2. served by PP, invokes `GetOAuth2RedirectURL` call (B) which returns a "redirect URL" (C), issues a 302 there
3. upon completion, the browser is redirected to `/apps/appid/oath2/remote/complete` (D)
4. state verified, served by PP, invokes `OnOAuth2Complete` (E)
5. app stores the token as "User" (cc @hanzei)

Proposed naming:
- A: `/apps/appid/oauth2/remote/connect`
- B: `"oauth2_connect"`
- C: connect URL
- D: no change
- E: `"oauth2_complete"`

apps/permissions.go
hanzei: Makes sense. Please add a comment about that once your work on the docs.

catalintomai (approved):

catalintomai:
wrt to "How do we add to JS API client?", there is no ClientPP equivalent for JS fwik, so Zendesk app would have to call the `mmclient.Path*` API directly. cc: @jfrerich

hanzei (approved):

PR #119, opened 2021-03-29
Apps metadata for client

mickmister:
#### Summary

This PR adds a `AppMetadataForClient` struct to send back the bot user id and bot username to the client.

#### Ticket Link

Fixes https://mattermost.atlassian.net/browse/MM-31083


larkox (approved):

apps/call.go
```diff
@@ -131,6 +131,9 @@ type CallResponse struct {
 
 	// Used in CallResponseTypeForm
 	Form *Form `json:"form,omitempty"`
+
+	// Used to provide info about the App to client, e.g. the bot user id
+	AppMetadata AppMetadataForClient `json:"app_metadata"`
```
levb, line 136: 1/5 I'd rather avoid adding private fields (i.e. irrelevant to the app) to the CallResponse. Can we wrap `CallResponse` into a `ProxyCallResponse` perhaps?

0/5 I am also contemplating that `apps.Request` and `apps.Response` maybe better names.

levb (changes requested):

apps/call.go
mickmister: > 1/5 I'd rather avoid adding private fields (i.e. irrelevant to the app) to the CallResponse. Can we wrap `CallResponse` into a `ProxyCallResponse` perhaps?

So we would use `CallResponse` everywhere that we are handling what the app returned, but convert to a `ProxyCallResponse` when we are packaging it up for the UI?

> 0/5 I am also contemplating that `apps.Request` and `apps.Response` maybe better names.

Would this change apply to the redux types as well? @larkox What are your thoughts on this?

levb: redux types are used (should be used?) by the apps, so we'd need to have both in redux, yes; but `Proxy...` is not for App developers.

larkox: Long term ideal solution:
- Have two types file for Apps: package exported types (for apps developers) and internally used types (not exported outside of the package).
- ProxyResponse be a struct with the Response Embedded
- ProxyResponse to have a custom Marshaller to flatten the struct (I don't like the idea of having to handle in the client something like `response = {appData: {...}, response: {...}}`)

How does this sound?

mickmister: @larkox Wrapping the `CallResponse` inside the `ProxyCallResponse` doesn't require a special Marshaller. We can use an embedded struct like so:

```go
type ProxyCallResponse struct {
	*CallResponse

	AppMetadata *AppMetadataForClient `json:"app_metadata"`
}
```

apps/call.go
```diff
@@ -133,6 +133,25 @@ type CallResponse struct {
...
+// data for the client, such as information about the App's bot account.
+//
+// Apps will use the CallResponse struct to respond to a CallRequest, and the proxy will
+// decorate the response using the ProxyCallResponse to provide additional information.
+type ProxyCallResponse struct {
+	*CallResponse
+
+	// Used to provide info about the App to client, e.g. the bot user id
+	AppMetadata *AppMetadataForClient `json:"app_metadata"`
+}
```
mickmister, line 146: @levb We're introducing more pointers here. Should we instead always use a struct for `CallResponse` throughout the codebase instead of a pointer? The client expects `AppMetadata` to be optional, since it makes its own transient `CallResponse`s to create local errors, so that should actually be a pointer.

levb: 2/5 Just embed the fields themselves, no need for a new struct?

mickmister: @levb The PR is already merged on mobile that uses this struct https://github.com/mattermost/mattermost-mobile/blob/06be22d6f72af8aad42af3dd3a0ef18bd54108c0/app/mm-redux/types/apps.ts#L76-L91

levb: Ouch... OK. Can we file a ticket to rename this all later, since it doesn't affect dev experience?

levb (approved):
Approve, but file a ticket to flatten the struct

apps/call.go
mickmister: @levb If we wanted to change it, we should change it now before any of the clients are released. Wouldn't be a PITA to change it after the first release?

mickmister: @larkox What are your thoughts on this?

apps/call.go
larkox: For some reason I thought the marshaller would not flatten the list. LGTM.

apps/call.go
larkox: I like having a separated field, in case we need to add more stuff in the future. I don't want to pollute much more the response (we already have at top level markdown, error, navigate to url, use external browser... I like the app_metadata at least being in its own compartment). But 0/5.

larkox (approved):

apps/call.go
levb: @larkox Hmm, we control that namespace (`App`), so as long as we don't create conflicts, why do we need the extra level? I prefer that we consider that every "exported" type (even for internal use) comes with a "complexity cost". (code has a much lower complexity cost since it's somewhat self-explanatory, and contained).

mickmister:
@DHaussermann FYI nothing has functionally changed since you originally checked out this branch

DHaussermann (approved):
Tested and passed
This was tested along with client side functionality that leverages the data for bots.

No further functional testing is needed.

LGTM!

PR #123, opened 2021-03-30
[MM-33020] Make app ID validation more strict

hanzei:
### Summary
Given that we use app IDs in the store where key length is limited we should put a stricter limit on the length. Also, 3 characters is now the minimum limit.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-33020

### DIFF
--- a/apps/app.go
+++ b/apps/app.go
@@ -68,7 +68,7 @@ type ListedApp struct {
 
 const (
 	MinAppIDLength = 3
-	MaxAppIDLength = 25
+	MaxAppIDLength = 32
 )
 
 func (id AppID) IsValid() error {
--- a/apps/app.go
+++ b/apps/app.go
@@ -66,11 +66,18 @@ type ListedApp struct {
 	Labels    []model.MarketplaceLabel `json:"labels,omitempty"`
 }
 
-const MaxAppID = 32
+const (
+	MinAppIDLength = 3
+	MaxAppIDLength = 25
+)
 
 func (id AppID) IsValid() error {
-	if len(id) > MaxAppID {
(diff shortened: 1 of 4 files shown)


apps/app.go
```diff
@@ -66,11 +66,18 @@ type ListedApp struct {
 	Labels    []model.MarketplaceLabel `json:"labels,omitempty"`
 }
 
-const MaxAppID = 32
+const (
+	MinAppIDLength = 3
+	MaxAppIDLength = 25
```
levb, line 71: Why 25? And the comment says, up to 20?

catalintomai (approved):

apps/app.go
hanzei: I wasn't aware of the fact that the app id is not longer used directly. With that in mind I'm happy to revert back to 32 bytes. What do you think @levb ?

levb: @hanzei should work.

levb:
@hanzei @justinegeffen we should update the docs with this

hanzei:
@levb I didn't found any existing docs on the manifest properties. Do we need to write them from scratch?

levb (approved):

levb:
@hanzei godoc

hanzei:
@levb Is https://github.com/mattermost/mattermost-plugin-apps/pull/123/files#diff-f41d83db23b9875fe2c189732d707ee45e4f9b2187b5418e2d70f3d7cb55cd02R11-R12 enough?

levb:
Ooops, maybe jumped the gun cc @DHaussermann but unit tests seem to cover the changes.

PR #127, opened 2021-04-01
[MM-30403] Expand users mentioned in post

iomodo:
#### Summary
This PR adds support of the mentioned user expand.

#### Ticket Link

  Fixes https://mattermost.atlassian.net/browse/MM-30403

### DIFF
--- a/server/proxy/mention.go
+++ b/server/proxy/mention.go
@@ -41,7 +41,6 @@ func getMentionsEnabledFields(post *model.Post) model.StringArray {
 
 	ret = append(ret, post.Message)
 	for _, attachment := range post.Attachments() {
-
 		if attachment.Pretext != "" {
 			ret = append(ret, attachment.Pretext)
 		}
--- a/server/proxy/mention.go
+++ b/server/proxy/mention.go
@@ -1,0 +1,126 @@
+package proxy
+
+import (
+	"sync"
+
+	"github.com/mattermost/mattermost-server/v5/mlog"
+	"github.com/mattermost/mattermost-server/v5/model"
+	"github.com/mattermost/mattermost-server/v5/store"
+	"github.com/mattermost/mattermost-server/v5/utils/markdown"
+	"github.com/pkg/errors"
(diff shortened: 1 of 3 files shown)


levb (changes requested):
As discussed on the call, the primary focus should be providing the expand map for the Call Values; whether or not we need it on the Post is TBD, maybe as a separate "Expand" clause?

server/proxy/mention.go
```diff
@@ -0,0 +1,125 @@
...
+	}
+	users := e.processText(buf)
+	mentions = join(mentions, users)
+
+	return mapToSlice(mentions)
+}
+
+// Given a post returns the values of the fields in which mentions are possible.
+// post.message, preText and text in the attachment are enabled.
+func getMentionsEnabledFields(post *model.Post) model.StringArray {
```
larkox, line 39: Should we consider also Embedded bindings?

server/proxy/mention.go
```diff
@@ -0,0 +1,125 @@
...
+
+	possibleMentions := model.PossibleAtMentions(text)
+	mentionChan := make(chan *mentionMapItem, len(possibleMentions))
+
+	var wg sync.WaitGroup
+	for _, mention := range possibleMentions {
+		wg.Add(1)
+		go func(mention string) {
+			defer wg.Done()
+			user, nErr := e.mm.User.GetByUsername(mention)
```
larkox, line 69: Are we using the plugin client to get the users? Any chance this may be exposing users to users with not enough permissions? (e.g. guest users).

larkox (changes requested):

justinegeffen:
@iomodo, would you be open to updating the docs with these changes? They can be found here: https://github.com/mattermost/mattermost-developer-documentation/tree/master/site/content/integrate/apps.

larkox (changes requested):
My previous comments still apply

hanzei:
Closing this PR for now

PR #131, opened 2021-04-06
[MM-34491] Don't overwrite root command label

hanzei:
#### Summary
Respect slash command root label instead of using the app ID.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-34491

### DIFF
--- a/docs/01-anatomy-hello.md
+++ b/docs/01-anatomy-hello.md
@@ -81,7 +81,7 @@ The Hello App creates a Channel Header button, and adds a `/helloworld send` com
 			"bindings": [
 				{
 					"icon": "http://localhost:8080/static/icon.png",
-                    "label": "helloworld",
+					"label": "helloworld",
 					"description": "Hello World app",
 					"hint":        "[send]",
 					"bindings": [
--- a/docs/01-anatomy-hello.md
+++ b/docs/01-anatomy-hello.md
@@ -81,6 +81,7 @@ The Hello App creates a Channel Header button, and adds a `/helloworld send` com
 			"bindings": [
 				{
 					"icon": "http://localhost:8080/static/icon.png",
+                    "label": "helloworld",
 					"description": "Hello World app",
 					"hint":        "[send]",
 					"bindings": [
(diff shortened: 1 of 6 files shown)


larkox (approved):

DHaussermann:
@hanzei this seems to be working on mobile though the first deploy needs a debug-clean. 

The slash command binding is now working with the `display_name` and the App ID is no longer appearing. 
It may make sense to live this PR open until we see this behavior on webapp as well.  

Side note: Huge thanks for the addition of the auto-complete for using the debug commands. This is very helpful.

docs/01-anatomy-hello.md
```diff
@@ -81,6 +81,7 @@ The Hello App creates a Channel Header button, and adds a `/helloworld send` com
 			"bindings": [
 				{
 					"icon": "http://localhost:8080/static/icon.png",
+                    "label": "helloworld",
```
mickmister, line 84: Nit: indentation. Same with other json changes

mickmister (approved):

levb (approved):

server/proxy/bindings.go
```diff
@@ -136,11 +136,6 @@ func (p *Proxy) scanAppBindings(app *apps.App, bindings []*apps.Binding, locPref
 			continue
 		}
 
-		if locPrefix == apps.LocationCommand {
```
levb, line 139: @hanzei Can we still default to AppID if the location/label are missing, or are you >0/5 on encouraging hierarchical AppIDs?

docs/01-anatomy-hello.md
hanzei: Good catch :+1:

server/proxy/bindings.go
hanzei: I'm opposed to using the AppID by default for two reasons:
1. 3/5 I would like to think about the AppID as "a technical detail". It's a technical necessity and should not be user facing.
2. 1/5 Default values like that are not obvious. I would like to rather error with a meaningful message to developer then doing "magic".

hanzei:
@levb I'm happy to continue the discussion about what default values we should set, but I'm going to merge this PR to unblock the other ones.

server/proxy/bindings.go
levb: SG

PR #135, opened 2021-04-07
Fix get bindings when expanding the user

larkox:
#### Summary
We were using bad the CheckAuthorization function in several places with almost no real consequence (the values were not used) but on the bindings when we tried to expand the user.

#### Ticket Link
None


hanzei (approved):
good catch :+1: 

Should we also change the argument name of `CheckAuthorized` from `sessionID` to `token`?

larkox:
The thing that is actually added here is a SessionID, so we should change token to sessionID probably. Will do.

levb (approved):

hanzei (approved):

PR #136, opened 2021-04-07
Add headers for static assets

iomodo:
#### Summary
This PR adds headers to the static assets

#### Ticket Link

  Fixes https://mattermost.atlassian.net/browse/MM-34648


larkox (approved):

larkox:
I am not savvy enough about this, but I wonder if there is any header that may become "dangerous" that we want to filter out.

hanzei (changes requested):
I'm with @larkox here. Copying all headers might have side effects that we don't know about. We should have an allow list instead.

hanzei (approved):

PR #138, opened 2021-04-09
Add API for clients to check the bots and oauth apps ids

larkox:
#### Summary
Add API for clients to check the bots and oauth apps ids

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-33158

#### Related pull requests
Webapp: https://github.com/mattermost/mattermost-webapp/pull/7850

### DIFF
--- a/server/http/restapi/get_ids.go
+++ b/server/http/restapi/get_ids.go
@@ -5,7 +5,7 @@ import (
 	"net/http"
 )
 
-func (a *restapi) handleGetBotIDs(w http.ResponseWriter, r *http.Request) {
+func (a *restapi) handleGetBotIDs(w http.ResponseWriter, r *http.Request, _, _ string) {
 	apps := a.proxy.GetInstalledApps()
 	ids := []string{}
 	for _, app := range apps {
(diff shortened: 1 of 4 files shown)


server/utils/utils.go
```diff
@@ -60,3 +61,8 @@ func LoadSession(mm *pluginapi.Client, sessionID, actingUserID string) (*model.S
 	}
 	return session, nil
 }
+
+func DumpObject(c interface{}) {
```
catalintomai, line 65: Is this one for future debugging sessions?

larkox: Yes. It was not my intention to commit this change, it is something I was using locally.

Do you want me to revert this change?

catalintomai: Might be useful for others as well (and avoids the need for everyone to constantly add and remove pre-checkin similar methods). Maybe just add a comment re: its purpose.

catalintomai (approved):

hanzei (approved):
LGTM

0/5 Maybe the API endpoints should take into account that the API might have to be enhanced in the future. Do we need a struct that contains the data instead of a list?

larkox:
I prefer the simplest the better, but you are right that having this in a struct may make this more expandable. Worst case scenario, we can always open a new api endpoint if we need more information.

server/http/restapi/get_ids.go
```diff
@@ -0,0 +1,30 @@
+package restapi
+
+import (
+	"encoding/json"
+	"net/http"
+)
+
+func (a *restapi) handleGetBotIDs(w http.ResponseWriter, r *http.Request) {
+	apps := a.proxy.GetInstalledApps()
```
levb, line 9: We should require user authentication on these APIs, or did I miss it?

levb (changes requested):

hanzei (approved):

levb (approved):

PR #139, opened 2021-04-12
Add e2e app test

iomodo:
#### Summary
This PR adds e2e test for the test-app
run `/make test-app-e2e`


server/config/service.go
```diff
@@ -112,6 +112,9 @@ func (s *service) Reconfigure(stored StoredConfig, services ...Configurable) err
 }
 
 func (s *service) StoreConfig(sc StoredConfig) error {
+	if strings.HasPrefix(*s.mattermostConfig.ServiceSettings.SiteURL, "http://localhost:") {
```
iomodo, line 115: This is done because `s.mm.Configuration.SavePluginConfig` method is not working on an e2e environment.

iomodo:
@levb @catalintomai kind reminder to review this.

levb:
@catalintomai this seems relevant, but is behind, and should perhaps be done with the `test-e2e` framework? What shall we do with the PR?

catalintomai:
@levb, I'd say let's close it. It seems fragile anyway and as you pointed out, we can come up with a better approach.

PR #143, opened 2021-04-14
Set minimum version

larkox:
#### Summary
Set plugin minimum version.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-34829


levb (approved):

levb:
Got to fix the e2e test, @hanzei or @catalintomai would know how

plugin.json
```diff
@@ -3,7 +3,7 @@
     "name": "Cloud Apps",
     "description": "Cloud Apps Registry and API proxy.",
     "version": "0.1.0",
-    "min_server_version": "5.26.0",
+    "min_server_version": "5.35.0",
```
hanzei, line 6: Per https://community-release.mattermost.com/core/pl/qz9gedh74jrufkmd7oaobh11ye the April 21th Cloud release wont be based on v5.35.0. Given that we target that release (right?) we need to stick to `5.34.0`

hanzei (changes requested):

plugin.json
levb: Well, we do depend on a later build, it appears, so don't we have a bigger problem then? Or do we continue to compile with 5.35, but _require_ 5.34 in JSON?

hanzei: Yes, we can require 5.34 here
```suggestion
    "min_server_version": "5.34.0",
```

hanzei:
Pushed to p2 as we can't require 5.35 right now.

DHaussermann (approved):
Tested and passed
- Plugin will **not** start on 5.34
- Plugin will start on 5.35

LGTM!

hanzei:
Merging as QA has approved and it's a simple change.

PR #142, opened 2021-04-14
Manage permissions on bot creation

mickmister:
#### Summary

This PR changes some things about the bot user when it is created:

- The bot is created regardless if `act_as_bot` was requested
- The bot user's access token is only created if `act_as_bot` was requested
- The bot user's roles are updated to be able to create/read posts in any channel, only created if `act_as_bot` was requested

I'm mainly making this PR to show that we can solve the post permission issue without making the bot a sysadmin. It made sense to implement both tickets in this PR since I was touching post permissions and app permissions already.

#### Ticket Link

Fixes https://mattermost.atlassian.net/browse/MM-34577
Fixes https://mattermost.atlassian.net/browse/MM-34195

PR #146, opened 2021-04-16
Test InstallApp and ensureBot

mickmister:
#### Summary

Tests for PR https://github.com/mattermost/mattermost-plugin-apps/pull/142

I've introduced some `Client4` mocking in this PR. I don't want it to hold up https://github.com/mattermost/mattermost-plugin-apps/pull/142, so I made this a separate PR.

#### Ticket Link

Fixes https://mattermost.atlassian.net/browse/MM-34577
Fixes https://mattermost.atlassian.net/browse/MM-34195


server/clients/client4.go
```diff
@@ -0,0 +1,34 @@
...
+	UpdateUserRoles(userID, newRoles string) (bool, *model.Response)
+	CreateUserAccessToken(userID, name string) (*model.UserAccessToken, *model.Response)
+	GetOAuthApp(clientID string) (*model.OAuthApp, *model.Response)
+	CreateOAuthApp(*model.OAuthApp) (*model.OAuthApp, *model.Response)
+}
+
+/*
+	asAdmin := model.NewAPIv4Client(conf.MattermostSiteURL)
+	asAdmin.SetToken(session.Token)
+*/
```
hanzei, line 20: Remove?

server/clients/client4.go
```diff
@@ -0,0 +1,34 @@
...
+
+type ClientService interface {
+	NewClient4(siteURL string) Client4
+}
```
hanzei, line 24: Is that level on direction really needed? Are additional clients expected to be added?

server/proxy/install_test.go
```diff
@@ -0,0 +1,211 @@
...
+
+	appStore.EXPECT().Get(appID).Times(2).Return(app, nil)
+	appStore.EXPECT().Save(app).Return(nil)
+
+	upstreams := map[apps.AppID]upstream.Upstream{}
+	up := mock_upstream.NewMockUpstream(ctrl)
+	cr := &apps.CallResponse{
+		Markdown: "Install done",
+	}
+	bb, _ := json.Marshal(cr)
```
hanzei, line 76: Why not check the error to double check?

hanzei (approved):
Urg, the amount of mocking that needs to be done for the test seems very inflexible to me. I don't have an objection to merging this, but it got me thinking how tests can be written without depending on so much.

mickmister:
@levb Do you think we should be adopting tests like this (mocking Client4 methods individually etc)? If not, I think this PR should be closed

levb:
@mickmister 0/5. Seems like a lot of work, and we have the `test-e2e` that allows us to exercise these paths, no?

mickmister:
Closing as it doesn't seem to be providing any value. We can reference this PR if we'd like to add more unit testing to the framework.

PR #148, opened 2021-04-16
Add explicit app permission type for bot posting in all channels

mickmister:
#### Summary

https://github.com/mattermost/mattermost-plugin-apps/pull/142 made it so the bot is given a role to be able to post in any channel. This PR makes it so this role is only applied when the app specifically requests permission to do so.

#### Ticket Link

Fixes https://mattermost.atlassian.net/browse/MM-34920

### DIFF
--- a/apps/permissions.go
+++ b/apps/permissions.go
@@ -20,6 +20,13 @@ const (
 	// receive permissions to any resources, need to be added explicitly.
 	PermissionActAsBot Permission = "act_as_bot"
 
+	// PermissionPostAllAsBot means that the Bot User will be assigned the role
+	// of `system_post_all`. This makes it so the bot can create posts in direct message
+	// and group message channels. Without this permission, the bot must be explicitly added to
+	// public and private channels by the app, and the bot would not be able to create posts
+	// in direct message or group message channels.
+	PermissionPostAllAsBot Permission = "post_all_as_bot"
+
 	// PermissionActAsUser means that the app is allowed to connect users'
 	// OAuth2 accounts, and then use user API tokens.
 	PermissionActAsUser Permission = "act_as_user"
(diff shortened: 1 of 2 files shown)


apps/permissions.go
```diff
@@ -20,6 +20,13 @@ const (
 	// receive permissions to any resources, need to be added explicitly.
 	PermissionActAsBot Permission = "act_as_bot"
 
+	// PermissionPostAllAsBot means that the Bot User will be assigned the role
+	// of `system_post_all`. This makes it so the bot can create posts in direct message
+	// and group message channels. Without this permission, the bot must be explicitly added to
```
catalintomai, line 25: "This makes it so the bot can create posts in direct message and group message channels" - what other permissions the bot user gets by being assigned the `system_post_all` role?

levb: `PermissionReadWriteAllChannels` perhaps?

justinegeffen:
@mickmister, could you let me know where this content needs to be updated in the docs? They're located here: https://github.com/mattermost/mattermost-developer-documentation/tree/master/site/content/integrate/apps. Thank you!

mickmister:
@justinegeffen This PR will likely not be merged. Though if something were to be put into the docs related to this, it would be that:

Bots made by the Apps framework need to be added to teams and channels to be able to create/read posts in those channels. An App can add a bot to a channel when necessary by using the current acting user's token. cc @jfrerich for a reference of js code we can link.

Most use cases will allow for posting as the acting user, rather than the bot. For use cases like subscriptions/webhooks, the App should add the bot to the channel using the acting user's token that is creating the subscription.

(Shouldn't necessarily be documented, but) We haven't finalized the system of how bots should be able to post in certain channels. For instance, bots cannot create posts in DMs or GMs under any circumstances, because they cannot be added to those channels, and they don't have global permissions to post in an arbitrary channel.

larkox (approved):

PR #158, opened 2021-04-21
Refresh bindings when OAuth2 process is complete

mickmister:
#### Summary

~~This PR adds a new endpoint for Apps to call when they want to refresh a specific user's bindings.~~

~~`POST /api/v1/refresh_bindings?user_id=(some_user_id)`~~

This PR dispatches the `refresh_bindings` websocket message when an OAuth2 connect process is complete.

This fix requires no changes to webapp, server, or mobile. The ServiceNow and Zendesk apps do not need to be updated either.

#### Ticket Link

Fixes https://mattermost.atlassian.net/browse/MM-35076

#### Related Pull Requests

https://github.com/mattermost/mattermost-app-servicenow/pull/22

### DIFF
--- a/server/http/restapi/restapi.go
+++ b/server/http/restapi/restapi.go
@@ -38,6 +38,7 @@ func Init(router *mux.Router, mm *pluginapi.Client, conf config.Service, proxy p
 	subrouter.HandleFunc(config.PathCall, a.handleCall).Methods("POST")
 	subrouter.HandleFunc(mmclient.PathSubscribe, a.handleSubscribe).Methods("POST")
 	subrouter.HandleFunc(mmclient.PathUnsubscribe, a.handleUnsubscribe).Methods("POST")
+	subrouter.HandleFunc(mmclient.PathRefreshBindings, a.handleRefreshBindings).Methods("POST")
 
 	// Bot and OAuthApps checks
 	subrouter.HandleFunc(mmclient.PathBotIDs,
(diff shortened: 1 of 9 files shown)


levb (approved):
1/5 We need a dynamic bindings hello example, separate PR

server/http/restapi/restapi.go
```diff
@@ -38,6 +38,7 @@ func Init(router *mux.Router, mm *pluginapi.Client, conf config.Service, proxy p
 	subrouter.HandleFunc(config.PathCall, a.handleCall).Methods("POST")
 	subrouter.HandleFunc(mmclient.PathSubscribe, a.handleSubscribe).Methods("POST")
 	subrouter.HandleFunc(mmclient.PathUnsubscribe, a.handleUnsubscribe).Methods("POST")
+	subrouter.HandleFunc(mmclient.PathRefreshBindings, a.handleRefreshBindings).Methods("POST")
```
hanzei, line 41: Can we simplify things for an app developer here by assuming that the bindings need to be reloaded every time an OAuth flow completes?

hanzei (changes requested):
Such API endpoint seems excessive and raises security concerns. I'm wondering if really need it.

server/http/restapi/restapi.go
catalintomai: Maybe the path here is to start with giving the developer the functionality first and move it to the framework once we understand the usage/needs better.

catalintomai (approved):

server/http/restapi/restapi.go
hanzei: How would the transition from the API to the framework work?

mickmister:
I'm not sure where docs should go for this, but something like this would suffice.

> When the OAuth2 process for connecting to an external service completes, the user will receive a websocket message that automatically refresh their bindings. There is currently no support for refreshing the bindings when disconnecting from the external service.

hanzei (approved):
Nice work :+1:

hanzei:
@mickmister The docs should go to https://developers.mattermost.com/integrate/apps/using-third-party-api/

levb (approved):

DHaussermann (approved):
Tested and passed. 
Tested this on webapp and mobile. Once a client connects(not yet supported on mobile) Both webapp and mobile clients will see the create option without the need to change channels.

Tested all 3 types of bindings (command. post menu, and channel header) on both mobile and webapp.

LGTM!

PR #161, opened 2021-04-21
replaced hello icon

levb:


examples/go/hello/bindings.go
```diff
@@ -21,7 +21,7 @@ func Bindings() []*apps.Binding {
 				{
 					Location:    "send",
 					Label:       "Survey a user",
-					Icon:        "https://raw.githubusercontent.com/mattermost/mattermost-plugin-jira/master/assets/icon.svg",
+					Icon:        "https://github.com/mattermost/mattermost-plugin-apps/raw/master/assets/profile.png",
```
hanzei, line 24: 0/5 should we use static assets here?

hanzei (approved):

mickmister (approved):
Non-blocking suggestion to use static assets as Ben suggested

levb:
@hanzei @mickmister I skipped the "static" route because TBH I was lazy and I really want to get rid of http-hello to start with and replace it with a 2-3 more standalone smaller examples. All separate PR(s).

PR #163, opened 2021-04-22
[MM-35054] Clean user context on Call and GetBindings endpoints

mickmister:
#### Summary

This PR makes sure the Context taken from user requests are vetted:
- A new struct `ContextFromUserAgent` is introduced that contains only the values we expect from the frontend
- All values in `ContextFromUserAgent` are verified for access control of the acting user
    - If any of the values are invalid, we return an error and do not process the request further
- Any values present in the `Context` outside of the fields in `ContextFromUserAgent` are discarded

#### Ticket Link

Fixes https://mattermost.atlassian.net/browse/MM-35054

Known TODOs:
- Handle root post id cleaning


server/proxy/proxy.go
```diff
@@ -178,3 +187,67 @@ func (p *Proxy) upstreamForApp(app *apps.App) (upstream.Upstream, error) {
...
+	case ctx.TeamID != "":
+		teamID = ctx.TeamID
+
+		_, err := p.mm.Team.GetMember(teamID, userID)
+		if err != nil {
+			return nil, errors.Wrapf(err, "failed to get team membership. user=%v team=%v", userID, teamID)
+		}
+
+	default:
+		return nil, errors.Errorf("no post, channel, or team context provided. user=%v", userID)
```
mickmister, line 245: 0/5 on how to handle this case. We don't currently have any use cases where we are submitting a call outside of the context of a given team, so I figured it should be required to provide at least a post id, channel id, or team id in the request.

server/proxy/proxy.go
```diff
@@ -178,3 +187,67 @@ func (p *Proxy) upstreamForApp(app *apps.App) (upstream.Upstream, error) {
 		return nil, utils.NewInvalidError("not a valid app type: %s", app.AppType)
 	}
 }
+
+func (p *Proxy) CleanUserCallContext(userID string, cc *apps.Context) (*apps.Context, error) {
+	cc = &apps.Context{
+		ContextFromUserAgent: cc.ContextFromUserAgent,
+	}
```
mickmister, line 194: Unwanted fields are discarded here by redefining the context as just the user agent fields

server/proxy/proxy.go
```diff
@@ -178,3 +187,67 @@ func (p *Proxy) upstreamForApp(app *apps.App) (upstream.Upstream, error) {
...
+		if err != nil {
+			return nil, errors.Wrapf(err, "failed to get channel membership. user=%v channel=%v", userID, channelID)
+		}
+
+		c, err := p.mm.Channel.Get(channelID)
+		if err != nil {
+			return nil, errors.Wrapf(err, "failed to get channel. channel=%v", channelID)
+		}
+
+		teamID = c.TeamId
```
hanzei, line 219: That seems way to forgiving to me. If a teamID or a chanelID from something the client doesn't have access to is provided, we should just error and not try to correct the input.

server/proxy/proxy.go
```diff
@@ -178,3 +187,67 @@ func (p *Proxy) upstreamForApp(app *apps.App) (upstream.Upstream, error) {
...
+
+	case cc.ChannelID != "":
+		channelID = cc.ChannelID
+
+		_, err := p.mm.Channel.GetMember(cc.ChannelID, userID)
```
hanzei, line 224: I both a `ChannelID` and a `PostID` are provided `GetMember` is done twice. Can we drop one?

server/proxy/proxy.go
```diff
@@ -178,3 +187,67 @@ func (p *Proxy) upstreamForApp(app *apps.App) (upstream.Upstream, error) {
 		return nil, utils.NewInvalidError("not a valid app type: %s", app.AppType)
 	}
 }
+
+func (p *Proxy) CleanUserCallContext(userID string, cc *apps.Context) (*apps.Context, error) {
```
hanzei, line 191: As pointed out below I think it should be a simple `CheckCallContext(userID string, cc *apps.Context) error` function.

server/proxy/enable.go
```diff
@@ -51,13 +61,23 @@ func (p *Proxy) DisableApp(cc *apps.Context, app *apps.App) (md.MD, error) {
...
+	}
+
 	app.Disabled = true
-	resp := p.Call("", "", &apps.CallRequest{
-		Call:    *app.OnDisable,
-		Context: cc,
-	})
-	if resp.Type == apps.CallResponseTypeError {
-		p.mm.Log.Warn("OnDisable failed, app disabled anyway", "err", resp.Error(), "app_id", app.AppID)
+
```
hanzei, line 70: Is the comment at https://github.com/mattermost/mattermost-plugin-apps/blob/951f54b169dcda31aca45f71f189606036a5027e/apps/manifest.go#L62 still valid?

server/proxy/enable.go
```diff
@@ -51,13 +61,23 @@ func (p *Proxy) DisableApp(cc *apps.Context, app *apps.App) (md.MD, error) {
...
+	var message md.MD
+	if app.OnDisable != nil {
```
hanzei, line 72: Should we fall back to a default call if none is provided similar to `OnInstall` or 		https://github.com/mattermost/mattermost-plugin-apps/blob/951f54b169dcda31aca45f71f189606036a5027e/server/proxy/oauth2.go#L70?

server/proxy/enable.go
```diff
@@ -22,23 +23,32 @@ func (p *Proxy) EnableApp(cc *apps.Context, app *apps.App) (md.MD, error) {
...
 	}
 
-	resp := p.Call("", "", &apps.CallRequest{
-		Call:    *app.OnEnable,
-		Context: cc,
-	})
-	if resp.Type == apps.CallResponseTypeError {
-		p.mm.Log.Warn("OnEnable failed, app enabled anyway", "err", resp.Error(), "app_id", app.AppID)
+	var message md.MD
+	if app.OnEnable != nil {
```
hanzei, line 38: Same question here

server/proxy/enable.go
```diff
@@ -12,6 +12,7 @@ import (
 	"github.com/mattermost/mattermost-plugin-apps/server/utils/md"
 )
 
+// EnableApp is never called. Context vetting needs to be done by the caller via p.CleanUserCallContext, though we're temporarily doing it here instead.
```
hanzei, line 15: 0/5 is the comment `EnableApp is never called` really helpful? It might be obsolete soon.

apps/context.go
```diff
@@ -34,25 +27,39 @@ type Context struct {
...
+	// Data accepted from the user agent
+	ContextFromUserAgent
+
+	// More data as requested by call.Expand
+	ExpandedContext
+}
+
+// ContextFromUserAgent is a subset of fields from Context that are accepted from the user agent
+// The values are vetted, and all fields present in the provided Context that are not in ContextFromUserAgent are discarded when the Call comes from an acting user.
+type ContextFromUserAgent struct {
```
hanzei, line 45: I really like this solution :100:

server/proxy/service.go
```diff
@@ -33,6 +33,7 @@ type Proxy struct {
 
 type Service interface {
 	Call(sessionID, actingUserID string, creq *apps.CallRequest) *apps.ProxyCallResponse
+	CleanUserCallContext(actingUserID string, cc *apps.Context) (*apps.Context, error)
```
hanzei, line 36: 2/5 I don't like that the permission check is split between the proxy and the caller. That makes it hard to verify that permission are checked.

I'm advocating for moving `CleanUserCallContext` out of the proxy and letting the `caller´ deal with checking the permission.

server/http/restapi/call.go
```diff
@@ -10,17 +10,31 @@ import (
...
+	res, err := a.handleCall(sessionID, actingUserID, call)
+	if err != nil {
+		httputils.WriteError(w, err)
+		return
 	}
 
 	httputils.WriteJSON(w, res)
 }
+
+func (a *restapi) handleCall(sessionID, actingUserID string, call *apps.CallRequest) (*apps.ProxyCallResponse, error) {
```
hanzei, line 29: Nit pick: Does a separate function make the code more readable?

server/http/restapi/bindings.go
```diff
@@ -3,27 +3,45 @@ package restapi
...
+	bindings, err := a.handleGetBindings(sessionID, actingUserID, cc)
 	if err != nil {
 		httputils.WriteError(w, err)
 		return
 	}
 
 	httputils.WriteJSON(w, bindings)
 }
+
+func (a *restapi) handleGetBindings(sessionID, actingUserID string, cc *apps.Context) ([]*apps.Binding, error) {
```
hanzei, line 33: Same question here

hanzei (changes requested):
Nice work :+1: Also thanks for adding some tests


This will have a great negative effect on performance. But I don't think there is an easy way around it that.

jupenur:
There's one additional concern I probably didn't voice before: the Call API could also be used to spoof Notify. This partially addresses it, because after these changes it would no longer be possible to spoof SubjectUserJoinedChannel for channels the user can't join, but it would it would still be possible to spoof things like SubjectUserLeftChannel without actually leaving a channel.

Not sure if the Notify issue also should be addressed in this PR or even how to address it, but it's probably something that needs to be addressed in any case.

server/proxy/proxy.go
mickmister: @hanzei So the logic is currently:

```
if post id is provided:
    - Verify user is a member of the post's channel
        - If they are a member, extract the channel id from the post, and extract the team id from the post's channel. Do not even check what channel id or team id the client provided.
        - If they are not a member, throw an error
else if channel is provided:
    - Verify user is a member of the provided channel
        - If they are a member, extract the team id from the channel. Do not check what team id the client provided.
        - If they are not a member, throw an error
else if team id is provided:
    - Verify user is a member of the provided team
        - If they are a member of the team, use the team id.
        - If they are not a member of the team, throw an error
else: (no ids provided at all)
    - Throw an error
```

Are you suggesting that we instead do something like:

```
if post id is provided:
    - Verify user is a member of the post's channel
        - If they are a member, use this post id and do not use the post's channel id
        - If they are not a member, throw an error
if channel id is provided (no else here):
    - Verify user is a member of the provided channel
        - If they are a member, use this channel and do not use the channel's team id
        - If they are not a member, throw an error
if team id is provided (no else here):
    - Verify user is a member of the provided team
        - If they are a member, use this team
        - If they are not a member, throw an error
```

This second strategy could result in data inconsistencies, but that would be the client's fault. With the first (current) strategy we would be potentially using data that goes against what the client provided. The client should never provide a post id that does not match the channel id it provided but it definitely could (especially if the client is not the webapp or mobile app, i.e. not our code). I don't think we have a use case for providing a channel id that is different than the post's channel id.

Instead of "filling in" the values based on other values (filling in the channel id based on the provided post), are you suggesting we check to see if the channel id provided matches the post and error out if that's not the case?

server/proxy/proxy.go
mickmister: @hanzei That's not true, because this case statement won't run if the post id was provided

server/proxy/enable.go
mickmister: The functions exist but they are not "supported" as those functions are never called.

server/proxy/enable.go
mickmister: I was thinking the same. Are you suggesting to remove the entire comment, or just that sentence?

server/http/restapi/call.go
mickmister: @hanzei @levb I wrote it this way so that it was easier to test. It made it so I didn't need to use requests or writers in the test. That being said, the test is actually less useful since it covers less code. Do you suggest I modify the test to use requests and writers instead of extracting this function?

mickmister:
> the Call API could also be used to spoof Notify

@levb Is there a mechanism in place to differentiate between user-agent provided calls, and proxy plugin created calls (like Notify)? It would be something that the app itself needs to check, but I don't think there is a way to objectively reject calls that come from the user-agent that are formed to act as Notify calls.

server/proxy/proxy.go
hanzei: @mickmister Can we assume that every (valid) call request has either
1. Only a `ChannelID`
2. A `PostID` and a `ChannelID`
3. A `PostID` and a `ChannelID` and a `TeamID`
4. A `ChannelID` and a `TeamID`


If so, can we validate these three cases? My main motivation is to reduce the number of RPC calls.

mickmister: @hanzei Currently call requests will always have a channel id and team id (except for the call for fetching bindings due to the PRs below not being merged yet). The "optional" fields are the post id and the root post id, because sometimes the call has to do with a given post, and sometimes the call is just for a given channel and not a specific post.

> If so, can we validate these three cases? My main motivation is to reduce the number of RPC calls.

I think the current implementation in this PR is attempting to minimize the number of RPC calls. I'm not sure how it could be improved. I'm happy to discuss a solution in the MM Apps channel if you'd like.

https://github.com/mattermost/mattermost-webapp/pull/7902
https://github.com/mattermost/mattermost-mobile/pull/5316

server/proxy/proxy.go
mickmister: @hanzei How can we discard unwanted fields if we only return an error?

server/proxy/enable.go
mickmister: @hanzei I honestly don't know. @levb Do you think we should notify the app even if it didn't declare that it wants to be notified on enable/disable?

server/proxy/service.go
mickmister: Moved the function to the `restapi` package

server/http/restapi/call.go
mickmister: @hanzei I've put the functions back the way they were, and modified the tests to use http requests

mickmister:
@levb To avoid spoofing of "notify" by a user call to `/api/v1/call`, I'm thinking we should be sending a secret to the app for calls that are not made by a user. If the secret is not present for a notify call, the app should reject the request. What do you think?

mickmister:
@jupenur @levb @hanzei Validations are now removed from this PR. The PR is now just stripping unwanted values from the user's context. Removing notify calls will be in a separate PR.

jupenur (approved):

server/proxy/enable.go
levb: no

levb (approved):

levb:
@mickmister I'd like you to implement naming changes, but approving for all else

server/http/restapi/call.go
```diff
@@ -5,22 +5,93 @@ import (
...
+		if err != nil {
+			return nil, errors.Wrapf(err, "failed to get channel membership. user=%v channel=%v", userID, cc.ChannelID)
+		}
+
+		c, err := mm.Channel.Get(channelID)
+		if err != nil {
+			return nil, errors.Wrapf(err, "failed to get channel. channel=%v", channelID)
+		}
+
+		teamID = c.TeamId
```
hanzei, line 78: Do we need to do that? are inconsistent `cc.ChannelID` and `cc.TeamID` something the apps plugin need to catch?

hanzei: This continues the discussion from https://github.com/mattermost/mattermost-plugin-apps/pull/163#discussion_r618512481. I'm wondering if we are fine with data inconsistencies coming from the user agent, which would grant us a performance boost, because we can drop this call.

server/proxy/proxy.go
hanzei: Just overwrite the struct behind the pointer.

hanzei (approved):
Non blocking: I would prefer if we shield the back end from malicious user input using separate data structs instead of clearing it. E.g. `CallRequest` contains a `ContextFromUserAgent` instead of `Context`.

mickmister:
> `CallRequest` contains a `ContextFromUserAgent` instead of `Context`.

So then we need a separate struct for the `CallRequest` sent to the app right? I'm fine with that but can be done in a separate PR right?

mickmister:
@hanzei Please see my previous comment, I meant to mention you in it

server/http/restapi/call.go
mickmister: >  are inconsistent cc.ChannelID and cc.TeamID something the apps plugin need to catch?

@hanzei I think so. A given app should be able to trust that all data passed in is correct and valid

> I'm wondering if we are fine with data inconsistencies coming from the user agent

Maybe we should fail instead of filling in the gaps of the inconsistent values.

server/http/restapi/call.go
```diff
@@ -5,22 +5,93 @@ import (
...
+	cc = a.conf.GetConfig().SetContextDefaults(cc)
+
+	call.Context = cc
+	res := a.proxy.Call(sessionID, actingUserID, call)
 	httputils.WriteJSON(w, res)
 }
+
+func cleanUserCallContext(mm *pluginapi.Client, userID string, cc *apps.Context) (*apps.Context, error) {
+	cc = &apps.Context{
+		ContextFromUserAgent: cc.ContextFromUserAgent,
```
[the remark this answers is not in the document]
mickmister: The `p.Call` method is already setting the acting user id. 0/5 on if it needs to be set before then

server/http/restapi/call.go
mickmister: Discuss further and handle in another PR?

server/http/restapi/call.go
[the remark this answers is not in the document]
mickmister: I've added a commit that sets the `ActingUserID` here

hanzei (approved):

DHaussermann (approved):
Tested and passed
- Followed instruction provided by dev to modify the requset and received an error as expected.
- Briefly regression tested

LGTM!

hanzei:
I'm going to merge this PR as it is and file a ticket on the open discussions next week. Hope your are fine with that @mickmister

PR #165, opened 2021-04-26
Introduce ReadOnly property for form fields

mickmister:
#### Summary

The `ReadOnly` field is already supported by the webapp and mobile app. This PR adds the missing `ReadOnly` field on the backend, as well as adding some constants for field types.

#### Related Pull Requests

https://github.com/mattermost/mattermost-plugin-autolink/pull/159

### DIFF
--- a/apps/field.go
+++ b/apps/field.go
@@ -1,6 +1,7 @@
 package apps
 
 type FieldType string
+type TextFieldSubtype string
 
 const (
 	// Text field. "subtype":"textarea" for multi-line text input (Modal only?).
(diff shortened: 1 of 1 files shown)


apps/field.go
```diff
@@ -22,6 +23,14 @@ const (
 
 	// A mattermost channel reference (~name).
 	FieldTypeChannel FieldType = "channel"
+
+	TextFieldSubtypeInput     TextFieldSubtype = "input"
```
hanzei, line 27: According to https://developers.mattermost.com/integrate/apps/interactivity/ the field is named `text`. Are the docs wrong?

hanzei (changes requested):
Are there client changes required for this?

levb (approved):

apps/field.go
mickmister: @hanzei The field type is `text`, but the text *subtype* can be any that are listed here: https://github.com/mattermost/mattermost-webapp/blob/74cbcb7c3f0dff7e21ecb88776b8841dd1fa55e6/components/widgets/settings/text_setting.tsx#L8

DHaussermann (approved):
Tested and passed
- Tested using Michael's new autolink app https://github.com/mattermost/mattermost-plugin-autolink/pull/159
- Confirmed that read only fields are behaving as expected
- Tested on webapp and mobile to ensure client side support for both
- No issues found 

LGTM!

apps/field.go
mickmister: It is already documented in the interactivity docs https://github.com/mattermost/mattermost-developer-documentation/blob/master/site/content/integrate/apps/interactivity/_index.md#modal-forms

hanzei: Yep, but it's documented as `text` there not `input`. Fixed in [`3cda372` (#805)](https://github.com/mattermost/mattermost-developer-documentation/pull/805/commits/3cda3721955353179c53fdc4908349e29836a143)

hanzei (approved):

PR #169, opened 2021-04-27
[MM-35089] Improve error and call logging

hanzei:
#### Summary
https://github.com/mattermost/mattermost-plugin-apps/commit/781dd61f2efa2bf48c5b206cfab226ca48a05d7a adds a `Debug` log message for any call originated from a client. I thought about also logging calls from the apps plugin, but it might be to spammy and the code to do that didn't turn out nice.

https://github.com/mattermost/mattermost-plugin-apps/commit/e6e5a13cb78d172a436a5dfc7d30629687214198 adds logging to certain interesting errors that previously only propagated to the client. I've made use of `github.com/mattermost/mattermost-plugin-api/experimental/bot/logger` by adding a small wrapper around it to make the API fit. Only the cases that are interesting for debugging and not spammy get logged. 

#### Ticket Link
Fixes https://mattermost.atlassian.net/browse/MM-35089
Fixes https://mattermost.atlassian.net/browse/MM-34807

### DIFF
--- a/server/http/restapi/call.go
+++ b/server/http/restapi/call.go
@@ -26,6 +26,7 @@ func (a *restapi) handleCall(w http.ResponseWriter, req *http.Request) {
 		"Received call response",
 		"app_id", call.Context.AppID,
 		"acting_user_id", call.Context.ActingUserID,
+		"error", res.ErrorText,
 		"type", res.Type,
 		"path", call.Path,
 	)
--- a/server/http/restapi/call.go
+++ b/server/http/restapi/call.go
@@ -22,5 +22,13 @@ func (a *restapi) handleCall(w http.ResponseWriter, req *http.Request) {
 		res.Type = apps.CallResponseTypeOK
 	}
 
+	a.mm.Log.Debug(
+		"Received call response",
+		"app_id", call.Context.AppID,
+		"acting_user_id", call.Context.ActingUserID,
+		"type", res.Type,
+		"path", res.Call.Path,
+	)
+
 	httputils.WriteJSON(w, res)
 }
--- a/server/http/restapi/call.go
+++ b/server/http/restapi/call.go
@@ -27,7 +27,7 @@ func (a *restapi) handleCall(w http.ResponseWriter, req *http.Request) {
 		"app_id", call.Context.AppID,
 		"acting_user_id", call.Context.ActingUserID,
 		"type", res.Type,
-		"path", res.Call.Path,
+		"path", call.Path,
 	)
 
 	httputils.WriteJSON(w, res)
--- a/server/logger/logger.go
+++ b/server/logger/logger.go
@@ -1,0 +1,41 @@
+package logger
+
+import (
+	apilogger "github.com/mattermost/mattermost-plugin-api/experimental/bot/logger"
+)
+
+type Logger = apilogger.Logger
+type LogContext = apilogger.LogContext
+
+type LogAPI interface {
+	Error(message string, keyValuePairs ...interface{})
+	Warn(message string, keyValuePairs ...interface{})
+	Info(message string, keyValuePairs ...interface{})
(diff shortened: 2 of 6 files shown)


server/proxy/bindings.go
```diff
@@ -87,20 +92,19 @@ func (p *Proxy) GetBindingsForApp(sessionID, actingUserID string, cc *apps.Conte
 
 	resp := p.Call(sessionID, actingUserID, bindingsRequest)
 	if resp == nil || resp.Type != apps.CallResponseTypeOK {
-		// TODO Log error (chance to flood the logs)
-		// p.mm.Log.Debug("Response is nil or unexpected type.")
-		// if resp != nil && resp.Type == apps.CallResponseTypeError {
-		// 	p.mm.Log.Debug("Error getting bindings. Error: " + resp.Error())
-		// }
+		if resp != nil && resp.Type == apps.CallResponseTypeError {
```
catalintomai, line 95: Could we combine the two `if`s in one? A bit counter-intuitive to have `resp == nil` at the top and then `resp != nil` inside.

hanzei: Not really. The best flow that I found is:
```go
resp := p.Call(sessionID, actingUserID, bindingsRequest)
if resp == nil || (resp.Type != apps.CallResponseTypeError && resp.Type != apps.CallResponseTypeOK) {
	logger.Debugf("Bindings response is nil or unexpected type.")
	return nil, nil
}

if resp.Type == apps.CallResponseTypeError {
	logger.Debugf("Error getting bindings. Error: " + resp.Error())
	return nil, nil
}
```

@catalintomai Let me know what you think

server/http/restapi/call.go
```diff
@@ -22,5 +22,13 @@ func (a *restapi) handleCall(w http.ResponseWriter, req *http.Request) {
 		res.Type = apps.CallResponseTypeOK
 	}
 
+	a.mm.Log.Debug(
+		"Received call response",
+		"app_id", call.Context.AppID,
+		"acting_user_id", call.Context.ActingUserID,
+		"type", res.Type,
+		"path", call.Path,
```
levb, line 30: 0/5 maybe include `"error",res.ErrorText`?

server/logger/logger.go
```diff
@@ -0,0 +1,41 @@
+package logger
+
+import (
+	apilogger "github.com/mattermost/mattermost-plugin-api/experimental/bot/logger"
+)
+
+type Logger = apilogger.Logger
+type LogContext = apilogger.LogContext
+
+type LogAPI interface {
```
levb, line 10: sorry, why is this necessary? Is it because of the `...`? If so, should we make `logger.Logger` compliant with `pluginapi`? I think it's done the way it is for historical reasons (was it p.API compatibility?) cc @larkox

levb (approved):

server/proxy/bindings.go
catalintomai: fwiw I like it better. 1/5.

catalintomai (approved):

mickmister (approved):

server/proxy/bindings.go
hanzei: Changed in 4c5a3b62bc866a589a3d894950a4c67b88c25b66

server/http/restapi/call.go
hanzei: Changed in 4c5a3b62bc866a589a3d894950a4c67b88c25b66

server/logger/logger.go
hanzei: It's needed because `logger.Logger` accepts an API containing `LogError`, `LogWarn` etc. and the `pluginapi` exposes `Error` and `Warn` methods. Sure, we can fix that. For now this wrapper is fine to me.

PR #170, opened 2021-04-28
Make getOAuth2User use interface{} and not map[string]interface{} to unmarshal the value

levb:


mickmister (approved):
Any reason why the function accepts an interface to populate, rather than returning a concrete type?

levb:
@mickmister it's a custom entity, can even be a string. See https://github.com/levb/mattermost-app-productboard/blob/master/connect.go#L42 as an example.

hanzei (approved):
:+1:

hanzei:
@levb Should this PR be merged after `v0.5.0` is cut?

levb:
@hanzei it shouldn't affect anyone, objects should still unmarshal as maps; but no rush.

PR #171, opened 2021-04-28
eliminated the use of md5

levb:
Ticket: https://mattermost.atlassian.net/browse/MM-35129


catalintomai (approved):

server/store/service.go
```diff
@@ -70,3 +65,19 @@ func (s *Service) hashkey(globalPrefix, namespace, prefix, id string) string {
 
 	return key
 }
+
+func hashkey(globalPrefix, namespace, prefix, id string) string {
+	namespacePrefixHash := make([]byte, 20)
+	sha3.ShakeSum128(namespacePrefixHash, []byte(namespace+prefix))
```
jupenur, line 71: My suggestion on the Jira ticket was to not hash the namespace at all. Is that not doable? I guess it comes down to the OAuth2 APIs namespacing by app ID instead of bot ID, but since all apps have a bot associated with them anyway, couldn't we use bot IDs for OAuth2, too?

The 20-byte SHAKE128 hash *should* be enough to protect namespaces against collision attacks, but in case there's ever a weakness discovered in the algorithm, we need to have a path to upgrade it. My concern is that if we don't store the unhashed keys anywhere, upgrading to a new algorithm may not be possible.

server/store/service.go
```diff
@@ -70,3 +65,19 @@ func (s *Service) hashkey(globalPrefix, namespace, prefix, id string) string {
...
+	idHash := make([]byte, 16)
+	sha3.ShakeSum128(idHash, []byte(id))
+
+	encodedPrefix := make([]byte, ascii85.MaxEncodedLen(len(namespacePrefixHash)))
+	_ = ascii85.Encode(encodedPrefix, namespacePrefixHash)
+
+	encodedID := make([]byte, ascii85.MaxEncodedLen(len(idHash)))
+	_ = ascii85.Encode(encodedID, idHash)
+
+	return globalPrefix + path.Join(string(encodedPrefix), string(encodedID))
```
jupenur, line 82: Nit: why the `path.Join`? ASCII85 encodings contain slashes anyway, so you can't use them blindly to split keys to namespaces and IDs -- you have to rely on the known length of the namespace. Seems to me like the extra slash here is just taking up space.

server/store/service.go
levb: Trying to figure it out. 
- The limit is actually 50
- The AppID is not limited in length, the bot user ID is 28. = 28
- The global prefix - need at least 1 leading byte to differentiate from all other keys (*can prefix all other keys instead) = 0
- no separator needed if using fixed 28 chars for namespace = 0
- leaves 22 bytes for app-specific prefix (type, list all keys by type) and ID
- ID is 20 bytes (16 byte hash, ascii 85)
- Apps can have a 2-letter prefixes (clear text)

Not perfect but WFM. Implementing now.

server/store/service.go
levb: thx! missed it. was a leftover from the md5 code.

server/store/service.go
jupenur: @levb if you're still short on space with the 26-character IDs, you can also zbase32-decode them and re-encode using ASCII85. That should free up another 6 characters.

levb: 1/5 the limit would feel artificial to a developer, but for most apps even a 1-letter prefix "should be enough!"

levb: FWIW, I do not consider this KV store implementation a long-term solution.

levb: Ended up implementing slightly differently, with a 2-char global prefix/namespace and a 2-char app-specific one.

server/store/service_test.go
```diff
@@ -0,0 +1,139 @@
...
+			key, err := s.hashkey(tc.globalPrefix, tc.botUserID, tc.prefix, tc.id)
+			if tc.expectedError != "" {
+				require.NotNil(t, err)
+				require.Equal(t, tc.expectedError, err.Error())
+			} else {
+				require.NoError(t, err)
+				require.Equal(t, tc.expected, key)
+				require.Equal(t, tc.expectedLen, len(key))
+
+				gp, b, p, h, _ := parseHashkey(key)
```
hanzei, line 131: Nit pick: Check that the error is `nil`

server/config/constants.go
```diff
@@ -53,16 +53,31 @@ const (
...
-// base64 URL encoding that already uses '-' and '_'.
+// KV namespace
+//
+// Keys starting with a '.' are reserved for app-specific keys in the "hashkey"
+// format. Hashkeys have the following format (see service_test.go#TestHashkey
+// for examples):
+//
+//  - global prefix of ".X" where X is exactly 1 byte (2 bytes)
+//  - bot user ID (26 bytes)
+//  - app-specific prefix, limited to 2 non-space ASCII characters, right-filled
```
hanzei, line 64: Out of scope of this PR: What is the use case of such prefix? KVList per prefix?

hanzei (approved):
Looks good to me. Thanks for adding tests :+1:

catalintomai (approved):

server/config/constants.go
mickmister: `KVList` would be costly as there is no parameter for prefix. We would have to use SQL if we wanted to efficiently loop through prefixed keys.

hanzei: Good point @mickmister :+1:

server/store/service.go
```diff
@@ -49,24 +50,63 @@ func NewService(mm *pluginapi.Client, conf config.Service) *Service {
...
+	case len(k) == 0:
+		return "", errors.New("key must not be empty")
+	case len(b) != 26:
+		return "", errors.Errorf("botUserID %q must be exactly 26 ASCII characters", botUserID)
+	case len(gns) != 2 || gns[0] != '.':
+		return "", errors.Errorf("global prefix %q is not 2 ASCII characters starting with a '.'", globalNamespace)
+	}
+
+	hashed := hashkey(gns, b, ns, k)
+	if len(hashed) > model.KEY_VALUE_KEY_MAX_RUNES {
```
jupenur, line 80: Nit: This seems like a thing that should be covered by tests, not checked at runtime. If `hashkey` is working correctly, the length is never more than `model.KEY_VALUE_KEY_MAX_RUNES`.

jupenur (approved):

hanzei:
@justinegeffen Heads up that this PR needs docs

justinegeffen:
Thanks, @hanzei. I'll sort it out. :)

PR #172, opened 2021-04-29
Remove system_post_all role from bot accounts

mickmister:
#### Summary

This PR removes the block that adds the system_post_all role on app install

### DIFF
--- a/server/proxy/install.go
+++ b/server/proxy/install.go
@@ -170,27 +170,6 @@ func (p *Proxy) ensureBot(app *apps.App, actingUserID string, client *model.Clie
 	app.BotUsername = fullBot.Username
 
 	if app.RequestedPermissions.Contains(apps.PermissionActAsBot) {
-		if user == nil {
-			user, response = client.GetUser(fullBot.UserId, "")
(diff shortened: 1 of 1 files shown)


levb (approved):

jfrerich (approved):

hanzei:
@mickmister Anything specific that needs to be tested here?

hanzei (approved):

catalintomai (approved):

DHaussermann:
Subscriptions are working which I think was the main change needed to accommodate this change. 

I think I see a regression here. I'm noticing that when you create a ticket from a post in a DM - the feedback post does not get created tp confirm your ticket creation. This post would be made on behalf of the user making the ticket so, it should still be working. 

@jfrerich any thoughts on this? 

cc @catalintomai and @hanzei also curious if we have any certainty around if this can be addressed at an App level.

DHaussermann (approved):
Tested and passed

Subscriptions work as expected. 

The issue above seems related to code not yet deployed to AWS. I will retest on the next build.

LGTM!

PR #173, opened 2021-04-29
MM-35063: fixed static path traversal vulnerability

levb:
https://mattermost.atlassian.net/browse/MM-35063


server/http/gateway/static.go
```diff
@@ -12,16 +14,19 @@ import (
...
-
 	if appID == "" {
 		httputils.WriteError(w, utils.NewInvalidError("app_id not specified"))
 		return
 	}
 
 	vars := mux.Vars(req)
-	assetName := vars["name"]
+	assetName := path.Clean(vars["name"])
+	if assetName == "." || strings.Contains(assetName, "/..") {
```
catalintomai, line 24: wouldn't `path.Clean` remove the "/.." ? https://play.golang.org/p/Vnfn7OawTG8

levb: Not always (see https://play.golang.org/p/WPI-alRj-Cn) but there was a typo, and it should just be a prefix check anyway. Done.

catalintomai (approved):

hanzei (approved):
LGTM. Maybe add a test?

mickmister (approved):

jupenur (changes requested):
This is not sufficient and does not for example cover the case from the Jira ticket: ServiceNow running in HTTP mode. The reason is many servers and HTTP routers treat escaped paths as unescaped, including `github.com/gorilla/mux`. That means a double-escaped path passed to the apps plugin can still cause issues.

An additional check needs to be added for paths with escaping.

levb:
@jupenur: I added the unescaping. Please review.

jupenur (approved):
I was expecting just one additional round of unescaping and then passing the original escaped value forward after validation. But this should work, too.

PR #174, opened 2021-04-30
Add js hello-world example

hanzei:
#### Summary
This PR adds the code from https://developers.mattermost.com/integrate/apps/quick-start-js/ to this repo for easier discovery and version control.

It also fixes a couple of issue:
- Renamed the go example to `hello-world` for conistency
- Fix AWS lambda js example manifest
- Formating

#### Ticket Link
None


examples/js/hello-world/app.js
```diff
@@ -0,0 +1,139 @@
...
+    res.sendFile(__dirname + '/icon.png');
+});
+
+app.use(express.json());
+
+app.post('/send/submit', (req, res) => {
+    const call = req.body;
+
+    let message = 'Hello, world!';
+    const user_message = call.values.message;
```
mickmister, line 97:
```suggestion
    const formMessage = call.values.message;
```

or maybe `submittedMessage`

examples/js/hello-world/app.js
```diff
@@ -0,0 +1,139 @@
...
+        then((channel) => {
+            const post = {
+                channel_id: channel.id,
+                message,
+            };
+
+            // Create a post
+            options.body = JSON.stringify(post);
+
+            fetch(mattermost_site_url + '/api/v4/posts', options);
```
mickmister, line 131: No error handling here?

examples/js/hello-world/app.js
```diff
@@ -0,0 +1,139 @@
...
+        });
```
mickmister, line 132: Would you be opposed to using `async/await` here?

```suggestion
    const channel = await fetch(mattermost_site_url + '/api/v4/channels/direct', options).then((mm_res) => mm_res.json());
        
    const post = {
        channel_id: channel.id,
        message,
    };

    // Create a post
    options.body = JSON.stringify(post);

    fetch(mattermost_site_url + '/api/v4/posts', options);
```

examples/js/hello-world/app.js
mickmister: Also, I think it would be helpful to have the example return `{markdown: 'Some ephemeral response'}` as it is very easy for the developer to implement, and showcases something specific to the framework. This can be in addition to the public post made here. I'm aware that the Go example does not make use of the ephemeral post.

If we remove the public post created here and use the `markdown` option to create an ephemeral post instead, the whole example becomes much simpler too.

mickmister (changes requested):
Mostly just questions for discussion. Thanks for implementing this @hanzei

examples/js/hello-world/app.js
levb: Hello apps generally skip on error handling to improve the readability.

Perhaps comments are warranted everywhere, and 2/5 maybe a separate `hello-errors?` @justinegeffen

levb (approved):
LGTM once @mickmister's comments are addressed.

examples/js/hello-world/app.js
```diff
@@ -0,0 +1,139 @@
...
+            Authorization: 'BEARER ' + call.context.bot_access_token,
+            'Content-Type': 'application/json',
+        },
+        body: JSON.stringify(users),
+    };
+
+    // Get the DM channel between the user and the bot
+    const mattermost_site_url = call.context.mattermost_site_url;
+
+    fetch(mattermost_site_url + '/api/v4/channels/direct', options).
```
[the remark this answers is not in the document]
hanzei: I'm leaning toward no. A developers has to start using `fetch` anyway as the apps specific API endpoints aren't exposed in `Client4`. My suggestion would be to start using a "real" app drive once we have it.

examples/js/hello-world/app.js
hanzei: @mickmister That snipped errors when I run it locally:
```
SyntaxError: await is only valid in async functions and the top level bodies of modules
```

Is `fetch` any async function?

examples/js/hello-world/app.js
hanzei: @mickmister I like the public post because it shows the use of the bot access token. 2/5 to keep it. In terms of error handling: I agree with @levb that it doesn't make the example more readable.

examples/js/hello-world/app.js
mickmister: Yes `fetch` returns a promise, making it an async function as far as the caller is concerned. This error above occurs because we are attempting to use the `await` keyword inside of a function that was not declared as `async`. This change needs to be made:

```diff
- app.post('/send/submit', (req, res) => {
+ app.post('/send/submit', async (req, res) => {
```

examples/js/hello-world/app.js
mickmister: Can we also include the `markdown` response (in one of the handlers), or is that not applicable here? I just think it's definitely something we want to emphasize because it's available and very easy to implement on the app's side.

examples/js/hello-world/app.js
mickmister: Any function that is `then`-able (meaning it has a `then` method that you can chain) returns a promise. Therefore, any function that is `then`-able can alternatively be called with `await`, instead of using `.then()`.

examples/js/hello-world/app.js
```diff
@@ -94,9 +94,9 @@ app.post('/send/submit', (req, res) => {
     const call = req.body;
 
     let message = 'Hello, world!';
-    const user_message = call.values.message;
-    if (user_message) {
-        message += ' ...and ' + user_message + '!';
+    const submitted_message = call.values.message;
+    if (submitted_message) {
+        message += ' ...and ' + submitted_message + '!';
```
mickmister, line 99: Any reason why we are using snake_case here? My original request was made because I was suggesting to use camelCase, though I didn't make that clear in the request. The typical JS convention is definitely to use camelCase (especially for variable names), though snake_case is usually used for json payloads being sent over the wire.

examples/js/hello-world/app.js
hanzei: Good point. I've added a `markdown` response in be76159.

examples/js/hello-world/app.js
hanzei: I _thought_ snake_case was used in the webapp, but that isn't the case actually. I've converted to camelCase.

hanzei:
@mickmister I've synced the go and js example both here and in https://github.com/mattermost/mattermost-developer-documentation/pull/812. Please take a look.

examples/js/hello-world/app.js
```diff
@@ -115,23 +115,26 @@ app.post('/send/submit', (req, res) => {
...
 
-            // Create a post
-            options.body = JSON.stringify(post);
+    const post = {
+        channel_id: channel.id,
+        message,
+    };
+
+    // Create a post
+    options.body = JSON.stringify(post);
```
mickmister, line 129: It seems weird that we're using the same `options` var here but not a big deal

examples/js/hello-world/app.js
```diff
@@ -115,23 +115,26 @@ app.post('/send/submit', (req, res) => {
...
+
+    fetch(mattermostSiteURL + '/api/v4/posts', options);
```
mickmister, line 131: We'll want to wait for this to be finished before calling `res.json`, otherwise we will run into the floating promises issue.

```suggestion
    await fetch(mattermostSiteURL + '/api/v4/posts', options);
```

examples/js/hello-world/app.js
```diff
@@ -115,23 +115,26 @@ app.post('/send/submit', (req, res) => {
...
 
-    fetch(mattermost_site_url + '/api/v4/channels/direct', options).
-        then((mm_res) => mm_res.json()).
-        then((channel) => {
-            const post = {
-                channel_id: channel.id,
-                message,
-            };
+    const channel = await fetch(mattermostSiteURL + '/api/v4/channels/direct', options).
+        then((res) => res.json())
```
mickmister, line 121:
```suggestion
        then((res) => res.json());
```

mickmister (approved):

examples/js/hello-world/app.js
```diff
@@ -0,0 +1,142 @@
...
+                            },
+                        ],
+                    },
+                ],
+            },
+        ],
+    });
+});
+
+app.post(['/send/form', '/send-modal/submit'], (req, res) => {
```
mickmister, line 67: Cool, I didn't know about this syntax

justinegeffen:
@hanzei - can we merge this?

hanzei:
@justinegeffen Yes

PR #175, opened 2021-04-30
include the expanded oauth2 user on oauth2complete

jfrerich:
#### Summary
When the user completes oauth2, the call from the proxy plugin should send the expanded user access_token so the 3rd party app can immediately make calls.

#### Ticket Link
N/A


mickmister (approved):
Makes sense to me :+1:

levb (approved):

PR #178, opened 2021-05-01
Change description text

aaronrothschild:
- Changed description text so admins understand what it does and don't disable it accidentally
- Added a link to Apps documentation in the header.


hanzei (approved):
LGTM

PR #180, opened 2021-05-05
MM-35389: Use the server's httpservice for outgoing http

levb:
#### Ticket Link
https://mattermost.atlassian.net/browse/MM-35389


server/store/manifest.go
```diff
@@ -86,7 +78,7 @@ func (s *manifestStore) initGlobal(awscli aws.Client, bucket string, manifestsFi
 		case len(parts) == 2 && parts[0] == "file":
 			data, err = os.ReadFile(filepath.Join(assetPath, parts[1]))
 		case len(parts) == 2 && (parts[0] == "http" || parts[0] == "https"):
-			data, err = httputils.GetFromURL(loc)
+			data, err = httpOut.GetFromURL(loc, false)
```
hanzei, line 81: Is that the expected behaviour in a dev environment? Does it make sense to support URLs to localhost even for production, i.e. when a customer runs there app locally alongside the MM server?

server/store/manifest.go
```diff
@@ -61,16 +60,9 @@ func (s *manifestStore) InitGlobal(awscli aws.Client, bucket string) error {
 	}
 	defer f.Close()
 
-	return s.initGlobal(awscli, bucket, f, assetPath)
-}
-
-// initGlobal reads in the list of known (i.e. marketplace listed) app
-// manifests.
```
hanzei, line 68: Move the comment to `InitGlobal`?

hanzei (changes requested):
Not quite sure what the expected behaviour is here. Questions below to get the discussion going.

server/store/manifest.go
levb: I fixed the dev mode support; as to production-local use case 2/5 that will be signaled with an appropriate flag, alongside devmode.

server/httpout/service.go
```diff
@@ -0,0 +1,49 @@
...
+
+func NewService(conf config.Service) Service {
+	return &service{
+		HTTPService: httpservice.MakeHTTPService(conf.GetMattermostConfig()),
+		conf:        conf,
+	}
+}
+
+func (s *service) Configure(x config.Config) {
+	s.HTTPService = httpservice.MakeHTTPService(s.conf.GetMattermostConfig())
```
hanzei, line 38: If a `Service` gets it's configuration form `s.conf` anyway, why does it need to be a `config.Configurable`? Should it create it's `HTTPService` on the fly?

server/httpout/service.go
```diff
@@ -0,0 +1,49 @@
...
+}
+
+func (s *service) GetFromURL(url string, trusted bool) ([]byte, error) {
+	client := s.MakeClient(trusted)
+	resp, err := client.Get(url)
+	if err != nil {
+		return nil, err
+	}
+	defer resp.Body.Close()
```
hanzei, line 47: Nit:
```suggestion
	defer _ = resp.Body.Close()
```

server/upstream/uphttp/upstream.go
```diff
@@ -57,7 +63,7 @@ func (u *Upstream) invoke(fromMattermostUserID string, call *apps.CallRequest) (
 
 // post does not close resp.Body, it's the caller's responsibility
 func (u *Upstream) post(fromMattermostUserID string, url string, msg interface{}) (*http.Response, error) {
-	client := u.getClient()
+	client := u.httpOut.MakeClient(true)
```
hanzei, line 66: To double check: This is fine because there is another layer of security (JWT) on top of the http connection, right? Would we still recommend customers to not use `localhost`?

hanzei (approved):
A few question, but nothing blocking

srkgupta (approved):
Thanks @levb. The changes LGTM from security perspective.

server/httpout/service.go
levb: That would certainly simplify things, but I figured I should follow the server pattern in case there's caching ever done in the implementation.

server/upstream/uphttp/upstream.go
levb: The app's RootURL is declared and accepted by the admin upon its install, so we can trust it.

server/httpout/service.go
levb: needs to be a func() then

server/httpout/service.go
hanzei: Good point on caching.

1/5 that it's preferred to blank out variables that are not used.
```suggestion
func (s *service) Configure(_ config.Config) {
	s.HTTPService = httpservice.MakeHTTPService(s.conf.GetMattermostConfig())
```

server/upstream/uphttp/upstream.go
hanzei: The `RootURL` is only accepted indirectly. What the admin adds is a _manifest_ pointing to an URL. Currently, it's not checked if the manifest is under the `RootURL`, right?

levb: Added HTTPRootURL to the consent screen. The manifest can come from anywhere, really, but the Root URL should be consented to.
image

aaronrothschild: Where on the screenshot is the "Rooturl" permission you are referring to @levb ?

levb: @aaronrothschild Access Remote HTTP API...

levb:
Since @hanzei already approved, I'm going to merge it to unblock another PR.

hanzei (commented):
LGTM :+1:

PR #181, opened 2021-05-09
Initial on-prem differences

levb:
- package `aws` renamed to `awsclient`
- provisioning and naming AWS functions are moved to `awsapps`
- introduced `Config.CloudMode` (false == onprem)
- Marketplace app manifests are only initialized in the cloud mode
- Added `/install --url` in on-prem mode
- Refactored command initialization

On-prem mode uses `MM_APPS_AWS_ACCESS_KEY`, `MM_APPS_AWS_SECRET_KEY`, and `MM_APPS_S3_BUCKET` variables to access AWS.


server/plugin.go
```diff
@@ -136,9 +126,12 @@ func (p *Plugin) OnActivate() error {
...
 
-	err = p.proxy.SynchronizeInstalledApps()
-	if err != nil {
-		p.mm.Log.Error("failed to update apps", "err", err.Error())
+	if conf.CloudMode {
+		err = p.proxy.SynchronizeInstalledApps()
+		if err != nil {
+			p.mm.Log.Error("failed to update apps", "err", err.Error())
+		}
+		p.mm.Log.Debug("updated the installed apps metadata")
```
catalintomai, line 134: remove duplicate (see below)?

server/config/config.go
```diff
@@ -79,3 +90,49 @@ func (c Config) SetContextDefaultsForApp(appID apps.AppID, cc *apps.Context) *ap
...
+
+	// use CloudLambdaAccessEnvVar for now to detect the Cloud mode
+	cloudLambdaAccessKey := os.Getenv(awsapps.CloudLambdaAccessEnvVar)
+	c.CloudMode = false
+	if cloudLambdaAccessKey != "" {
+		c.CloudMode = true
+	}
+
+	if c.CloudMode {
+		c.AWSLambdaAccessKey = os.Getenv(awsapps.CloudLambdaAccessEnvVar)
```
catalintomai, line 126: is `CloudMode` the MM Cloud (as opposed to Self-Managed AWS/Cloud ) mode? If yes, would it make sense to have `AWSLambdaAccessKey ` -> `APPS_LAMBDA_ACCESS_KEY` (or smth. that has no MM tie) instead of `MM_APPS_LAMBDA_ACCESS_KEY` ? (I look at the `MM` prefix here as the one that ties the var to the MM cloud, as opposed to the MM general namespace)

apps/awsapps/s3.go
```diff
@@ -0,0 +1,39 @@
...
+	"github.com/mattermost/mattermost-plugin-apps/apps"
+)
+
+const (
+	// S3BucketEnvVar is the environment variable containing the S3 bucket name
+	// used to host Apps' assets.
+	S3BucketEnvVar = "MM_APPS_S3_BUCKET"
+
+	// DefaultS3Bucket is the default S3 bucket name.
+	DefaultS3Bucket = "mattermost-apps-bucket"
```
catalintomai, line 17: Since S3 bucker names are globally unique, I wonder if we need a reserved name for the "default" MM Cloud one and a different one for the non-MM AWS Cloud ones (not sure if we can include some client-specific info here).

server/config/config.go
levb: Mattermost server uses `MM_` all over the place, so I assumed that this (`MM_APPS_`) is the right prefix for the new "public" settings.

For the cloud we are presently using `APPS_INVOKE_AWS_(ACCESS|SECRET)_KEY`, I left these unchanged for now. Once I find the explicit trigger for the cloud, we can use the same (MM_APPS_-prefixed) variables everywhere.

apps/awsapps/s3.go
levb: Right. I wonder if we should get rid of the default altogether since it is specific to MM Cloud.

server/config/config.go
catalintomai: ok, so `CloudMode` is for all clouds (MM and non-MM) and we'll use `MM_APPS_` prefix for the final/unique set of variables.

levb: hmmm, no :)

"CloudMode" is exclusively the Mattermost Cloud (should I rename it?) - the reverse is all "on-prem" which includes HTTP and AWS support ATM.

levb: Ok, it is `MattermostCloudMode` now; all else is "on-prem".

catalintomai: Thanks, I got confused about the "Cloud" part, which for me included on-prem "AWS" (which to me is cloud as well - could contain Azure, other hosting types in the future), and would need similar `*ACCESS|SECRET*` keys (we could reuse the same var names, `MM_APPS` prefixed for both MM and non-MM clouds). 0/5 - but with current naming it might become harder later to distinguish between AWS and non-AWS clouds (MM or non-MM).

levb: I agree, the `AWS` is missing from the names, wanted to keep them short but "type of cloud" is indeed important. Changing now.

levb: Ah, it was `MM_APPS_S3_BUCKET` that was already there that I was trying to be consistent with. @hanzei might know what's involved in renaming it if we want a universal `MM_APPS_AWS_` prefix.

levb: ok, since we use one set of credentials for S3 and Lambda, I renamed them to `MM_APPS_AWS_(ACCESS|SECRET)_KEY`. I'm keeping `MM_APPS_S3_BUCKET` as is for now since it may be a dependency, but `S3` is already AWS-specific.

catalintomai (approved):

server/config/config.go
```diff
@@ -79,3 +90,49 @@ func (c Config) SetContextDefaultsForApp(appID apps.AppID, cc *apps.Context) *ap
...
+	}
+
+	c.DeveloperMode = pluginapi.IsConfiguredForDevelopment(mmconf)
+
+	// use CloudAccessEnvVar for now to detect the Cloud mode
+	cloudAccessKey := os.Getenv(awsapps.CloudAccessEnvVar)
+	c.MattermostCloudMode = false
+	if cloudAccessKey != "" {
+		c.MattermostCloudMode = true
+	}
```
hanzei, line 123: https://github.com/mattermost/mattermost-server/blob/c90c371fd4016345e4fcc2c619d3b5c47edaeecc/model/license.go#L86 should be used to decide if cloud is used

server/plugin.go
```diff
@@ -94,14 +86,13 @@ func (p *Plugin) OnActivate() error {
...
 	// manifest store
-	conf := p.conf.GetConfig()
 	mstore := p.store.Manifest
 	mstore.Configure(conf)
-	// TODO: uses the default bucket name, do we need it customizeable?
-	manifestBucket := apps.S3BucketNameWithDefaults("")
-	err = mstore.InitGlobal(p.aws, manifestBucket, p.httpOut)
-	if err != nil {
-		return errors.Wrap(err, "failed to initialize the global manifest list from marketplace")
+	if conf.MattermostCloudMode {
```
hanzei, line 91: How would on-prem customers who use AWS to deploy there Apps listed?

cc @DHaussermann as we talked about this offline in the team meeting

hanzei (changes requested):
Thanks for taking ownership @levb  :+1: 

It's unclear to me how a customer would run there app in AWS based on these changes. The refactoring looks good to me.

server/config/config.go
levb: Done

server/plugin.go
levb: I have been assuming that `appsctl` would provide the functionality and the customer would use it with the appropriate credentials to manage their deployment.

levb:
@DHaussermann @hanzei I restored the `install <appid>` and the marketplace listing initialization functionality as discussed in the meeting this morning, I also kept the `install --url` in the onprem mode. cc @catalintomai

DHaussermann:
@levb I see an error here when trying to use the plugin in cloud mode. 
```
{"level":"error","ts":1620845720.558461,"caller":"mlog/sugar.go:25","msg":"failed to load local manifest for http-hello: not found","plugin_id":"com.mattermost.apps"}
{"level":"error","ts":1620845720.563808,"caller":"mlog/sugar.go:25","msg":"failed to load local manifest for test: not found","plugin_id":"com.mattermost.apps"}
{"level":"error","ts":1620845720.962682,"caller":"mlog/sugar.go:25","msg":"failed to load app com.mattermost.zendesk: key app.724b93d339d77f7a324443fc81f4989e9a093154 not found","plugin_id":"com.mattermost.apps"}
{"level":"error","ts":1620845720.966579,"caller":"mlog/sugar.go:25","msg":"failed to load app http-hello: key app.196f1be555785a200755180cb59ebb932002d0f0 not found","plugin_id":"com.mattermost.apps"}
{"level":"error","ts":1620845720.971104,"caller":"mlog/sugar.go:25","msg":"failed to load app test: key app.07cb3c27e9ae2eaa99822bb1335bbaab4aef5689 not found","plugin_id":"com.mattermost.apps"}
{"level":"error","ts":1620845720.976199,"caller":"mlog/log.go:247","msg":"Unable to activate plugin","plugin_id":"com.mattermost.apps","error":"failed to initialize own command handling: invalid autocomplete data in command: Named argument should not be before positional argument"}
{"level":"error","ts":1620845721.068128,"caller":"mlog/sugar.go:25","msg":"failed to load local manifest for http-hello: not found","plugin_id":"com.mattermost.apps"}
{"level":"error","ts":1620845721.073192,"caller":"mlog/sugar.go:25","msg":"failed to load local manifest for test: not found","plugin_id":"com.mattermost.apps"}
```
I don't understand why I did not see this before restarting the plugin with a cloud license applied. 

0/5 Do you merge master into this? @hanzei has a change that makes install a positional argument and not a named one. Maybe this is causing the issue?

server/command/service.go
```diff
@@ -30,52 +30,139 @@ type service struct {
...
+	}
 
-		autoComplete.AddCommand(c.autoComplete)
+	if conf.MattermostCloudMode {
+		// install only by ID (from the marketplace) in cloud mode
+		h.autoComplete.Arguments = append(h.autoComplete.Arguments, &model.AutocompleteArg{
+			HelpText: "ID of the app to install",
+			Type:     model.AutocompleteArgTypeText,
+			Data: &model.AutocompleteTextArg{
+				Hint: "secret string",
```
hanzei, line 119: Copy past error?

server/command/service.go
```diff
@@ -30,52 +30,139 @@ type service struct {
...
+	} else {
+		// install from URL in the on-prem mode
+		h.autoComplete.Arguments = append(h.autoComplete.Arguments, &model.AutocompleteArg{
+			Name:     "url",
+			HelpText: "URL of the App's manifest",
+			Type:     model.AutocompleteArgTypeText,
+			Data: &model.AutocompleteTextArg{
+				Hint: "URL",
+			},
+			Required: true,
```
hanzei, line 132: Why is an URL requires for on-prem?

server/command/service.go
```diff
@@ -30,52 +30,139 @@ type service struct {
...
+					},
+				},
+			},
+		},
+	}
 
-		autoComplete.AddCommand(c.autoComplete)
+	if conf.MattermostCloudMode {
+		// install only by ID (from the marketplace) in cloud mode
+		h.autoComplete.Arguments = append(h.autoComplete.Arguments, &model.AutocompleteArg{
```
hanzei, line 115: Does this command needs to be added in on-prem as well?

server/command/service.go
```diff
@@ -30,52 +30,139 @@ type service struct {
...
+			RoleID:   model.SYSTEM_ADMIN_ROLE_ID,
+			Arguments: []*model.AutocompleteArg{
+				{
+					Name:     "app-secret",
+					HelpText: "(HTTP) App's JWT secret used to authenticate incoming messages from Mattermost.",
+					Type:     model.AutocompleteArgTypeText,
+					Data: &model.AutocompleteTextArg{
+						Hint: "secret string",
+					},
+				},
```
hanzei, line 108: This need to be appended last to fix the autocomplete issue:
```
ERRO[2021-05-13T06:03:30.3800595+02:00] Unable to restart plugin on upgrade.          caller="mlog/log.go:251" err_details="failed to initialize own command handling: invalid autocomplete data in command: Named argument should not be before positional argument" err_where=installExtractedPlugin http_code=500 ip_addr="::1" method=POST path=/api/v4/plugins request_id=j8933jqcgirubpji87amtbpezr user_id=sdz3twtfbjbytd9uqf3xyrp6ay
```

server/command/install.go
```diff
@@ -11,28 +11,55 @@ import (
...
+	var appID apps.AppID
+	conf := s.conf.GetConfig()
+	if conf.MattermostCloudMode {
+		if len(params.current) == 0 {
+			return errorOut(params, errors.New("you must specify the app id"))
+		}
+		appID = apps.AppID(params.current[0])
+	} else {
+		switch {
+		case len(params.current) > 0:
```
hanzei, line 37: Does this also catch cases like `/apps install --url  http://localhost:3000/manifest`?

hanzei (changes requested):
There are a couple of outstanding issues with the commands.

I'm fine with using `manifests.json` as a way to list apps for on-prem customers, but I see it as a short term workaround. Do we need something like `/apps install --s3 com.mattermost.zendesk:0.2.1`?

awsclient/iam.go
```diff
@@ -38,7 +38,7 @@ type rolePrincipal struct {
 	Service string
 }
 
-func (c *client) MakeLambdaFunctionDefaultPolicy() (string, error) {
+func (c *client) EnsureLambdaRoleAndPolicy(log Logger) (string, error) {
```
levb, line 41: this really ought to be in `awsapps` but some other time, maybe once I figure out the API Gateway provisioning

server/command/service.go
levb: No longer relevant - it's now the only way in on-prem

levb:
@catalintomai @hanzei @DHaussermann @justinegeffen  8e802f1 contains the bulk of most recent changes.

The OnPrem mode now supports `install --url` only, and errors out if the manifest is for an AWS Lambda app.
The Cloud mode supports AWS Lambda apps only.
(all this should be ignorable in dev)

I fixed `appsctl` so I was able to successfully deploy the new `hello-lambda` app to my AWS account with relative ease, and it worked when I faked the "cloud mode" and included it in the catalog. However, I am stuck figuring out the API Gateway provisioning, so I don't have the steps of how to run Lambda-as-HTTP yet.

apps/awsapps/lambda.go
```diff
@@ -0,0 +1,36 @@
...
+	"github.com/mattermost/mattermost-plugin-apps/apps"
+)
+
+const MaxLambdaName = 64
+
+// LambdaName generates function name for a specific app, name can be 64
+// characters long.
+func LambdaName(appID apps.AppID, version apps.AppVersion, function string) string {
+	// Sanitized any dots used in appID and version as lambda function names can not contain dots
+	// While there are other non-valid characters, a dots is the most commonly used one
```
justinegeffen, line 18:
```suggestion
	// While there are other non-valid characters, a dot is the most commonly used one.
```

apps/awsapps/provision_data.go
```diff
@@ -16,7 +16,12 @@ import (
...
 
-const bundleStaticAssetsFolder = "static/"
+type Logger interface {
+	Error(message string, keyValuePairs ...interface{})
+	Warn(message string, keyValuePairs ...interface{})
+	Info(message string, keyValuePairs ...interface{})
+	Debug(message string, keyValuePairs ...interface{})
+}
 
 // ProvisionData contains all the necessary data for provisioning an app
```
justinegeffen, line 26:
```suggestion
// ProvisionData contains all the necessary data for provisioning an app.
```

examples/go/hello-lambda/Makefile
```diff
@@ -0,0 +1,34 @@
+.PHONY: all
+## all: builds and runs the app locally
+all: run
+
+.PHONY: build-linux
+## build-linux: build the executable for linux
```
justinegeffen, line 6:
```suggestion
## build-linux: Build the executable for Linux.
```

examples/go/hello-lambda/Makefile
```diff
@@ -0,0 +1,34 @@
+.PHONY: all
+## all: builds and runs the app locally
```
justinegeffen, line 2:
```suggestion
## all: Builds and runs the app locally.
```

examples/go/hello-lambda/Makefile
```diff
@@ -0,0 +1,34 @@
...
+all: run
+
+.PHONY: build-linux
+## build-linux: build the executable for linux
+build-linux:
+	GOOS=linux GOARCH=amd64 go build -o dist/hello-lambda
+
+.PHONY: build
+## build: build the executable for the local OS
```
justinegeffen, line 11:
```suggestion
## build: Build the executable for the local OS.
```

examples/go/hello-lambda/Makefile
```diff
@@ -0,0 +1,34 @@
...
+build:
+	go build -o dist/hello-lambda
+
+.PHONY: run
+## run: runs the app locally
```
justinegeffen, line 16:
```suggestion
## run: Runs the app locally.
```

examples/go/hello-lambda/Makefile
```diff
@@ -0,0 +1,34 @@
...
+run: build
+	LOCAL=true ./dist/hello-lambda 
+
+.PHONY: dist
+## dist: creates the bundle file
```
justinegeffen, line 21:
```suggestion
## dist: Creates the bundle file.
```

examples/go/hello-lambda/Makefile
```diff
@@ -0,0 +1,34 @@
...
+dist: build-linux
+	cp -r static dist; cp manifest.json dist/; cd dist/; zip -qr go-function hello-lambda; zip -r bundle.zip go-function.zip manifest.json static/
+
+.PHONY: clean
+## clean: deletes all
```
justinegeffen, line 26:
```suggestion
## clean: Deletes all.
```

examples/go/hello-lambda/Makefile
```diff
@@ -0,0 +1,34 @@
...
+clean:
+	rm -rf dist/
+
+.PHONY: help
+## help: prints this help message
```
justinegeffen, line 31:
```suggestion
## help: Prints this help message.
```

server/command/install.go
```diff
@@ -11,28 +11,51 @@ import (
...
+		if err != nil {
+			return errorOut(params, err)
+		}
+		appID = m.AppID
 	}
 
-	appID := params.current[0]
-
-	m, err := s.proxy.GetManifest(apps.AppID(appID))
+	// Get the manifest from the store, even if redundant
```
justinegeffen, line 57:
```suggestion
	// Get the manifest from the store, even if redundant.
```

server/command/install.go
```diff
@@ -11,28 +11,51 @@ import (
...
+	if conf.MattermostCloudMode {
+		if len(params.current) == 0 {
+			return errorOut(params, errors.New("you must specify the app id"))
+		}
+		appID = apps.AppID(params.current[0])
+	} else {
+		if manifestURL == "" {
+			return errorOut(params, errors.New("you must add a `--url`"))
+		}
+		// Trust the URL only in dev mode
```
justinegeffen, line 39:
```suggestion
		// Trust the URL only in dev mode.
```

server/command/service.go
```diff
@@ -30,52 +30,138 @@ type service struct {
...
+		autoComplete: &model.AutocompleteData{
+			Trigger:  "install",
+			HelpText: "Install an App.",
+			RoleID:   model.SYSTEM_ADMIN_ROLE_ID,
+		},
+	}
 
-		autoComplete.AddCommand(c.autoComplete)
+	if conf.MattermostCloudMode {
+		// install only by ID (from the marketplace) in cloud mode
```
justinegeffen, line 104:
```suggestion
		// Install only by ID (from the Marketplace) in Cloud mode.
```

server/command/service.go
```diff
@@ -30,52 +30,138 @@ type service struct {
...
+		h.autoComplete.Arguments = append(h.autoComplete.Arguments, &model.AutocompleteArg{
+			HelpText: "ID of the app to install",
+			Type:     model.AutocompleteArgTypeText,
+			Data: &model.AutocompleteTextArg{
+				Hint: "App ID",
+			},
+			Required: true,
+		})
+	} else {
+		// install from URL in the on-prem mode
```
justinegeffen, line 114:
```suggestion
		// Install from URL in the self-managed mode.
```

server/command/service.go
```diff
@@ -30,52 +30,138 @@ type service struct {
...
+	}
+	subCommands := s.allCommands()
+	var subs []string
+	for t := range subCommands {
+		subs = append(subs, t)
+	}
+	sort.Strings(subs)
+	helpText := "Available commands: " + strings.Join(subs, ", ")
+
+	// Add autocomplete for the subcommands alphabetically
```
justinegeffen, line 153:
```suggestion
	// Add autocomplete for the sub-commands alphabetically.
```

server/config/config.go
```diff
@@ -1,11 +1,17 @@
...
 
+	pluginapi "github.com/mattermost/mattermost-plugin-api"
 	"github.com/mattermost/mattermost-server/v5/model"
+	"github.com/pkg/errors"
 
 	"github.com/mattermost/mattermost-plugin-apps/apps"
+	"github.com/mattermost/mattermost-plugin-apps/apps/awsapps"
 )
 
 // StoredConfig represents the data stored in and managed with the Mattermost
```
justinegeffen, line 17:
```suggestion
// StoredConfig represents the data stored in and managed with Mattermost.
```

server/config/config.go
```diff
@@ -55,6 +62,11 @@ type Config struct {
 
 	// Maximum size of incoming remote webhook messages
```
justinegeffen, line 63:
```suggestion
	// Maximum size of incoming remote webhook messages.
```

server/config/service.go
```diff
@@ -87,30 +83,19 @@ func (s *service) reloadMattermostConfig() *model.Config {
...
 	mmconf := s.reloadMattermostConfig()
 
-	mattermostSiteURL := mmconf.ServiceSettings.SiteURL
-	if mattermostSiteURL == nil {
-		return errors.New("plugin requires Mattermost Site URL to be set")
+	newConfig := s.GetConfig()
+
+	// GetLicense silently drops an RPC error
+	// (https://github.com/mattermost/mattermost-server/blob/fc75b72bbabf7fabfad24b9e1e4c321ca9b9b7f1/plugin/client_rpc_generated.go#L864).
+	// When running in Mattermost cloud we must not fall back to the on-prem mode, so in case we get a nil retry once.
```
justinegeffen, line 90:
```suggestion
	// When running in Mattermost Cloud we must not fall back to the self-managed mode, so in case we get a nil retry once.
```

justinegeffen (commented):
Thanks, @levb! Not much feedback editorially other than some suggestions for code comments in terms of standardization in punctuation for consistency. Notable suggestion is "on-prem" to "self-managed" to be consistent with our documentation/naming. Let me know if anything is unclear or you have any questions! :)

examples/go/hello-lambda/Makefile
```diff
@@ -0,0 +1,34 @@
+.PHONY: all
```
hanzei, line 1: Should we also rename and update `js/aws_hello`? Separate PR /Ticket?

examples/go/hello-lambda/manifest-http.json
```diff
@@ -0,0 +1,13 @@
+{
+	"app_id": "hello-lambda",
+	"display_name": "Hello, AWS Lambda!",
```
hanzei, line 3: Nit pick: Version is missing

examples/go/hello-lambda/main.go
```diff
@@ -0,0 +1,82 @@
...
+//go:embed send_form.json
+var formData []byte
+
+//go:embed static
+var static embed.FS
+
+func main() {
+	localMode := os.Getenv("LOCAL") == "true"
+
+	// Serve its own manifest as HTTP for convenience in dev. mode.
```
hanzei, line 36: 1/5 that it's idiomatic to define the manifest as a go variable and rewrite based on the mode to keep things DRY.

server/command/service.go
```diff
@@ -30,52 +30,138 @@ type service struct {
...
-		httpOut: httpOut,
-	}
-	conf := configService.GetConfig()
-	subCommands := s.getSubCommands()
-	var subTrigger []string
-	for t, c := range subCommands {
-		if c.debug && !conf.DeveloperMode {
-			continue
-		}
+type params struct {
```
hanzei, line 33: 0/5 `commandParams`?

server/command/service.go
```diff
@@ -30,52 +30,138 @@ type service struct {
...
+	all := map[string]commandHandler{
+		"info": {
+			f:            s.executeInfo,
+			autoComplete: model.NewAutocompleteData("info", "", "Display debugging information"),
+		},
+		"list": {
+			f:            s.executeList,
+			autoComplete: model.NewAutocompleteData("list", "", "List installed and registered apps"),
+		},
+		"uninstall": {
```
hanzei, line 58: Should this be limited to system admins?

server/command/service.go
```diff
@@ -30,52 +30,138 @@ type service struct {
...
+	}
+
+	h.autoComplete.Arguments = append(h.autoComplete.Arguments, &model.AutocompleteArg{
+		Name:     "app-secret",
+		HelpText: "(HTTP) App's JWT secret used to authenticate incoming messages from Mattermost.",
+		Type:     model.AutocompleteArgTypeText,
+		Data: &model.AutocompleteTextArg{
+			Hint: "Secret string",
+		},
+	})
```
hanzei, line 133: Is that option really needed for Cloud?

server/command/service.go
```diff
@@ -30,52 +30,138 @@ type service struct {
...
+	uninstallAC.AddTextArgument("ID of the app to uninstall", "appID", "")
 
-		subTrigger = append(subTrigger, t)
+	all := map[string]commandHandler{
+		"info": {
+			f:            s.executeInfo,
+			autoComplete: model.NewAutocompleteData("info", "", "Display debugging information"),
+		},
+		"list": {
+			f:            s.executeList,
```
hanzei, line 55: Given that is coding is getting touched: Should `s.checkSystemAdmin` be added here for conistency?

server/proxy/proxy.go
```diff
@@ -184,6 +191,16 @@ func (p *Proxy) upstreamForApp(app *apps.App) (upstream.Upstream, error) {
...
+	}
+}
+
+func isAppTypeSupported(conf config.Config, m *apps.Manifest) error {
+	if conf.DeveloperMode {
+		return nil
 	}
+	if conf.MattermostCloudMode != (m.AppType == apps.AppTypeAWSLambda) {
+		return utils.NewForbiddenError("AWS lambda apps are supported only in Mattermost Cloud")
+	}
```
hanzei, line 204: Similar HTTP apps should only run in On-Prem. Should there be a check to enforce that?

server/plugin.go
```diff
@@ -136,11 +136,14 @@ func (p *Plugin) OnActivate() error {
 	}
 	p.mm.Log.Debug("initialized slash commands")
 
-	err = p.proxy.SynchronizeInstalledApps()
-	if err != nil {
-		p.mm.Log.Error("failed to update apps", "err", err.Error())
+	if conf.MattermostCloudMode {
```
hanzei, line 139: Does it make sense to also do that in dev mode given that the install command for AWS is available?

go.mod
```diff
@@ -3,7 +3,9 @@ module github.com/mattermost/mattermost-plugin-apps
 go 1.16
 
 require (
+	github.com/aws/aws-lambda-go v1.19.1
 	github.com/aws/aws-sdk-go v1.38.2
+	github.com/awslabs/aws-lambda-go-api-proxy v0.10.0
```
hanzei, line 8: 2/5 that this should go into a seperate `go.mod` file just for the example to make it easier to copy and run

hanzei (approved):
There are a lot of great changes in this PR :+1:

While I leaved some minor comments the overall changes look good to me.

larkox (approved):

DHaussermann:
@levb just to confirm, is this now obsolete in favor of the newer solution here https://github.com/mattermost/mattermost-plugin-apps/pull/191 ?

levb:
@DHaussermann correct, I just want to work the comments from this one into the #191

hanzei:
I moved all of my comments to https://github.com/mattermost/mattermost-plugin-apps/pull/19

levb:
> I see an error here

@DHaussermann the issue is, `init` creates a policy `arn:aws:iam::850679466851:policy/mattermost-apps-invoke-policy` which refers to the bucket, in my case:
```json
    "Version": "2012-10-17",
    "Statement": [
        {
            "Sid": "AllowS3",
            "Effect": "Allow",
            "Action": [
                "s3:GetObject"
            ],
            "Resource": [
                "arn:aws:s3:::test.lev/manifests/*",
                "arn:aws:s3:::test.lev/static/*"
            ]
        },
        {
            "Sid": "AllowLambda",
            "Effect": "Allow",
            "Action": [
                "lambda:InvokeFunction"
            ],
            "Resource": [
                "arn:aws:lambda:us-west-2:850679466851:function:hello-lambda_demo_go-function",
                "arn:aws:lambda:us-west-2:850679466851:function:com-mattermost-zendesk_0-2-1_js-function"
            ]
        }
    ]
}
```

I _think_ if you clean (may require manual cleanup) and then re-`aws init` with the new variables, it will work.

1/5 let's file separate tickets on `aws clean` not working well.

levb:
> I'm fine with using `manifests.json` as a way to list apps for on-prem customers, but I see it as a short term workaround. Do we need something like `/apps install --s3 com.mattermost.zendesk:0.2.1`?

`manifests.json` is ignored in the self-managed mode currently, is it not? I think the only way for the customer to install "our" zendesk app is to download the bundle locally (from github?) and then `appsctl provision app` to their AWS account.

We could allow an `appctl provision app --github` to close the loop (from source, or a release binary asset)

hanzei:
> > I'm fine with using `manifests.json` as a way to list apps for on-prem customers, but I see it as a short term workaround. Do we need something like `/apps install --s3 com.mattermost.zendesk:0.2.1`?
> 
> @hanzei `manifests.json` is ignored in the self-managed mode currently, is it not? I think the only way for the customer to install "our" zendesk app is to download the bundle locally (from github?) and then `appsctl provision app` to their AWS account.
> 
> We could allow an `appctl provision app --github` to close the loop (from source, or a release binary asset)

Surely, I would be great to support that. Maybe it makes sense to go further and "list" the apps somewhere, so it becomes `appctl provision app zendesk`. File a ticket?

DHaussermann:
@levb I'm not sure what you mean by...
>I think if you clean (may require manual cleanup) and then re-aws init with the new variables, it will work.

I'm thinking if I simply delete the policy it would solve the issue. (Though this does not solve the edge case of people using 2 or buckets) Also AWS allows you to simply edit the raw json. Seems like I could just manually add the target bucket this way as well?

DHaussermann:
@hanzei and @levb  I'm confused by why we're discussing this here instaed of https://github.com/mattermost/mattermost-plugin-apps/pull/191 Is this intentional?

levb:
@DHaussermann 👍  let's take this all to #191, my bad.

If you edit raw JSON you can add multiple buckets.

hanzei:
@levb Can we close this?

PR #188, opened 2021-05-13
[MM35712] MM Apps - Invalidate bindings on oAuth disconnect

jfrerich:
#### Summary
This ticket triggers a refresh bindings websocket event whenever `StoreOAuth2User` app service is called.

`StoreOAuth2User` is accessed through the rest API endpoint `/oauth2/user/<appid>` and is called by an app after a connect completion or disconnect.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-35712


jfrerich:
@DHaussermann I tested this out using zendesk and the description provided in the Jira ticket.

* /zendesk connect
* /zendesk disconnect
* /zendesk subscribe  (this will no longer show up with this code change)

DHaussermann (approved):
Tested and passed
- On oath disconnect event all binding  only available to connect user are removed without the need to change channels 
- Tested with binding in command, Channel header and Post menu
- Tested for all applicable bindings in Service Now and Zendesk
- No issues found 
LGTM!

Will retest again if there any changes requested during Dev review. 
Thanks @jfrerich This PR makes for a very nice usability improvement 🎉

mickmister:
Does this result in two binding refresh websocket events being sent to the client when the user connects?

jfrerich:
@mickmister, yes, it does because of the `p.dispatchRefreshBindingsEvent(actingUserID)` at the end of the `CompleteRemoteOAuth2` function.  That makes sense because we previously refreshed the bindings for the connect completion.  I would suggest refreshing the bindings at the AppService level and dropping the direct coupling to the restapi. The refresh could also be pushed into the `SaveUser` function inside `store/oauth2.go` 

https://github.com/mattermost/mattermost-plugin-apps/blob/a6f400df5c56f8fefbe67a14ea41d6b1e2dfffbf/server/proxy/oauth2.go#L50-L83

@lev, what are your thoughts?

server/appservices/oauth2.go
```diff
@@ -39,6 +42,7 @@ func (a *AppServices) StoreOAuth2User(appID apps.AppID, actingUserID string, ref
 	if err = a.ensureFromUser(actingUserID); err != nil {
 		return err
 	}
+	a.mm.Frontend.PublishWebSocketEvent(config.WebSocketEventRefreshBindings, map[string]interface{}{}, &model.WebsocketBroadcast{UserId: actingUserID})
```
levb, line 45: we should do it only if the value has changed

levb (changes requested):

server/appservices/oauth2.go
jfrerich: Thanks for explaining the reasoning in the meeting today.  I will make the adjustment in the PR and add the WIP label for now.  I believe there is a `changed` bit that I can use to determine if a value has changed.

hanzei (changes requested):
Blocking on @lev's comments

server/appservices/oauth2.go
jfrerich: I committed a change that compares the unmarshalled values before publishing the websocket event.  Is there a preference for comparing the bytes instead?

levb:
@jfrerich @hanzei What is the status of this PR?

hanzei:
@levb I'm talking over the PR and working on an update.

hanzei:
CI is broken ATM. https://github.com/mattermost/mattermost-plugin-apps/pull/266 will fix it.

hanzei:
I've updated the PR to only send the websocket event when the OAuth2 user changes. The event is `CompleteRemoteOAuth2` is not longer send as an app will set the OAuth2 user anyway after the authentication flow is complete. I've also updated the example app to make use of the changes and only show the `connect` or `disconnect` command if it's suitable. 

@levb @mickmister Please take another look.

jfrerich:
@hanzei, I pushed a small spelling fix to this PR. Thanks for taking this one.

hanzei:
@mickmister @levb Maybe we can get this into v0.8. Could you please take a look when you have the time?

server/proxy/oauth2.go
```diff
@@ -73,6 +73,5 @@ func (p *Proxy) CompleteRemoteOAuth2(in Incoming, appID apps.AppID, urlValues ma
 
 	p.conf.Telemetry().TrackOAuthComplete(string(appID), in.ActingUserID)
 
-	p.dispatchRefreshBindingsEvent(in.ActingUserID)
```
levb, line 76: What happened to this function? Is it still used? If it is, is it different from the explicit new `PublishWebSocketEvent` above?

levb (approved):

server/proxy/oauth2.go
hanzei: There is not difference between `dispatchRefreshBindingsEvent` and a manual `PublishWebSocketEvent`. `dispatchRefreshBindingsEvent` is still used in the proxy. https://github.com/mattermost/mattermost-plugin-apps/blob/d3ddbe8a43b48fdcfffaa437330b072f71d8631b/server/proxy/bindings.go#L207-L212

hanzei:
Merging with only one review as it's a small change

hanzei:
Documentation update: https://github.com/mattermost/mattermost-developer-documentation/pull/1210

PR #190, opened 2021-05-18
Change homepage url to point to docs

mickmister:
#### Summary

This PR changes the homepage url to point to https://mattermost.com/marketplace/app-framework

This is to match with the changes made in https://github.com/mattermost/mattermost-marketplace/pull/189#issuecomment-843396639

### DIFF
--- a/plugin.json
+++ b/plugin.json
@@ -2,7 +2,7 @@
     "id": "com.mattermost.apps",
     "name": "Apps",
     "description": "This plugin powers the Mattermost Apps Framework and is required for any Apps to run",
-    "homepage_url": "https://github.com/mattermost/mattermost-plugin-apps",
+    "homepage_url": "https://mattermost.com/marketplace/app-framework",
     "support_url": "https://github.com/mattermost/mattermost-plugin-apps/issues",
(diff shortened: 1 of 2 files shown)


plugin.json
```diff
@@ -2,7 +2,7 @@
     "id": "com.mattermost.apps",
     "name": "Apps",
     "description": "This plugin powers the Mattermost Apps Framework and is required for any Apps to run",
-    "homepage_url": "https://github.com/mattermost/mattermost-plugin-apps",
+    "homepage_url": "https://mattermost.com/marketplace/app-framework",
```
hanzei, line 5: Given that this plugin currently targets developers should we link to https://developers.mattermost.com/integrate/apps/ instead to make it easier to get started?

DHaussermann: 0/5 @hanzei yes this is aimed at developers. But we're showing this to everyone and Aaron's page has a broader context. 

Also the related PR must go to the marketplace very soon. I propose we can open a follow-up PR for further tweaks in the next day or two so the URLs stay consistent between the repo and the marketplace.

hanzei (approved):

catalintomai (approved):

plugin.json
hanzei: Makes sense to me

PR #191, opened 2021-05-24
[MM-35613] Self-managed hosting, with AWS Lambda support

levb:
This is the next iteration on https://github.com/mattermost/mattermost-plugin-apps/pull/181 (will close soon).

Due to number of changes I decided that a new clean PR is a better course of action, even for those who have already reviewed the original (sans AWS).

### DEMO/Test:
1. Set the following environment variables, use a personal AWS account, substitute the bucket/creds as appropriate:
```
export MM_APPS_AWS_REGION=us-west-2
export MM_APPS_S3_BUCKET=My-existing-bucket
export MM_APPS_PROVISION_AWS_ACCESS_KEY=****E7PF
export MM_APPS_PROVISION_AWS_SECRET_KEY=****TOKX
```

2. Initialize the AWS resources:
```
go run ./cmd/appsctl aws init --create --create-access-key
```
The output of the command will contain the "Invoke" access key, make sure you add it to the environment.

**Restart your Mattermost server with the new variables set**

3. Build `hello-lambda` bundle
```
cd ./examples/go/hello-lambda && make dist
```

4. Test the environment
```
go run ./cmd/appsctl aws test provision ./examples/go/hello-lambda/dist/bundle.zip
go run ./cmd/appsctl aws test lambda
go run ./cmd/appsctl aws test s3
```
(if the changes do not appear to work, resulting in 403/404 errors - wait for several minutes then refresh again)

5. Re-provision the `hello-lambda` again, to see the output
```
go run ./cmd/appsctl provision app ./examples/go/hello-lambda/dist/bundle.zip --update
```
Note that you should end up with the following _invoke_ policy. As more Apps are added, their functions will be added to the `AllowLambda` statement.
image

6. Install the `hello-lambda` with the command displayed in the prior step. (REFRESH if the changes do not appear to work, resulting in 403/404 errors - wait for several minutes then refresh again)

### Notes
1. new: **Cloud** vs **Self-managed** mode, cloud mode is on if the License is Cloud, or in case of `community.mattermost.com` is triggered on by the presence of otherwise deprecated `APPS_INVOKE_AWS_ACCESS_KEY` environment variable.
1. removed: `appsctl provision bucket`, the bucket must be created and provided by the customer.
2. new: `appsctl aws init [options...]`, initialize self-managed AWS environment to provision and use Apps.
3. new: `appsctl aws clean`, (best attempt, unfinished) to cleanup default entities created for Apps.
4. new: `appsctl aws test`, provision `hello-lambda` example App and test access to its resources (can be further improved).
5. new: environment variables to document, see https://github.com/mattermost/mattermost-plugin-apps/blob/74ef34fa8aa98309031ab78e2513f81f585b7e21/upstream/upaws/aws.go#L9-L23.
6. new: `/apps install aws $APP_ID $VERSION` in self-managed mode.e.g. `/apps install hello-lambda demo`
7. After an apps is uninstalled in on-prem, the manifest is still listed.

Other
- Marketplace app manifests are only initialized in cloud mode
- Combined all AWS access code (Client, Provisioner, Upstream) in `upstream/upaws`
- Refactored command initialization

#### Install commands

- `/apps install marketplace com.mattermost.servicenow` is used in Cloud to install from the Marketplace
- `/apps install http http://localhost:3000/manifest --app-secret abc` is used in on-prem to install from an HTTP server
- `/apps install aws com.mattermost.servicenow 0.1.0` is used in on-prem to install an App that was previously deployed to cloud

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-35613

### TODO
- Add `hello-lambda` README.md
- Documentation


apps/manifest.go
```diff
@@ -98,7 +98,44 @@ type Manifest struct {
 	// Cloud;
 	// - define path->function mappings, aka "routes". The function with the
 	// path matching as the longest prefix is used to handle a Call request.
-	AWSLambda []AWSLambdaFunction `json:"aws_lambda,omitempty"`
+	AWSLambda []AWSLambda `json:"aws_lambda,omitempty"`
+}
+
+// AWSLambda describes a distinct AWS Lambda function defined by the app, and
```
levb, line 104: 0/5 Makes it easier to have it here rather than try to abstract type-specific structs; may still come as a separate PR. Other AWS-related types etc are now in `upstrem/upaws`

e2e/cypress/integration/bindings/channel_header_spec.ts
```diff
@@ -14,8 +14,7 @@ const baseURL = Cypress.config('baseUrl');
 const pluginID = Cypress.config('pluginID');
 const helloManifestRoute = 'example/hello/mattermost-app.json';
 
-const addManifestCommand = `/apps debug-add-manifest --url ${baseURL}/plugins/${pluginID}/${helloManifestRoute}`;
-const installAppCommand = '/apps install http-hello --app-secret 1234';
+const installAppCommand = `/apps install --url ${baseURL}/plugins/${pluginID}/${helloManifestRoute} --app-secret 1234`;
```
levb, line 17: The E2E test now run in the "self-managed" mode, so install by URL works.

examples/go/hello-lambda/Makefile
```diff
@@ -0,0 +1,34 @@
+.PHONY: all
```
levb, line 1: Need README.MD

examples/go/hello-lambda/pong.json
```diff
@@ -0,0 +1,4 @@
+{
```
levb, line 1: Used by `appsctl aws test lambda`

examples/go/hello-lambda/static/test.txt
```diff
@@ -0,0 +1 @@
+static pong
```
levb, line 1: used by `appsctl aws test s3`

upstream/upaws/aws.go
```diff
@@ -0,0 +1,98 @@
+package upaws
```
levb, line 1: @stylianosrigas this one is relevant for you to look at

upstream/upaws/initialize.go
```diff
@@ -0,0 +1,126 @@
...
+	Bucket          string
+	PolicyARN       ARN
+	UserARN         ARN
+	GroupARN        ARN
+	ExecuteRoleARN  ARN
+	AccessKeyID     string
+	AccessKeySecret string
+}
+
+func InitializeAWS(asAdmin Client, log Logger, params InitParams) (r *InitResult, err error) {
```
levb, line 35: @stylianosrigas also here - the initialization of policies, roles, user, group.

DHaussermann:
@levb when testing in Cloud mode, the instructions above work for the most part. I saw couple issues with this...
1. It seems as though the stock permissions applied when creating a bucket do not work. When creating a bucket the default setting is block all public access. When left unchanged I saw the follwoing:
image
Making the following permission change resolved the problem I was seeing
image
Please advise if this is expected or acceptable.
2. As per point 6. above, new variable names should be explicitly documented. Seeing `"invoke"` in step 2. confused me as it seemed the variable names had not changed. (The new banes are in the terminal window but I still made a user error here)
3. The instruction set should explicitly state which environment variables are needed by the appsctl tool and which are needed by the server. I did this without updating the `bash` and the install process was failing due to the server missing some variables. 

Regression testing of Cloud mode is still ongoing. Also have not attempted to use the AWS clean commands.

cmd/appsctl/main.go
```diff
@@ -28,23 +31,43 @@ var rootCmd = &cobra.Command{
...
+		if quiet {
+			log.SetLevel(logrus.ErrorLevel)
+		}
 	},
 }
 
-func createAWSClient() (aws.Client, error) {
-	accessKey := os.Getenv("APPS_PROVISION_AWS_ACCESS_KEY")
-	secretKey := os.Getenv("APPS_PROVISION_AWS_SECRET_KEY")
+func AsProvisioner() (upaws.Client, error) {
```
hanzei, line 44: What is the difference to `AsTest`?

examples/go/hello-lambda/Makefile
hanzei: Separate ticket?

examples/go/hello-lambda/bindings.json
```diff
@@ -0,0 +1,25 @@
...
+	"type": "ok",
+	"data": [
+		{
+			"location": "/command",
+			"bindings": [
+				{
+					"icon": "/static/icon.png",
+					"label": "hello-lambda",
+					"description": "Hello Lambda app",
+					"hint":        "[send]",
```
hanzei, line 11:
```suggestion
					"hint": "[send]",
```

examples/go/hello-lambda/main.go
```diff
@@ -0,0 +1,88 @@
...
+
+func main() {
+	localMode := os.Getenv("LOCAL") == "true"
+
+	// Serve its own manifest as HTTP for convenience in dev. mode.
+	if localMode {
+		http.HandleFunc("/manifest.json", writeJSON(manifestHTTPData))
+	} else {
+		http.HandleFunc("/manifest.json", writeJSON(manifestAWSData))
+	}
```
hanzei, line 44: 1/5 that it's idiomatic to define the manifest as a go variable and rewrite based on the mode to keep things DRY.

examples/go/hello-lambda/manifest-http.json
```diff
@@ -0,0 +1,13 @@
+{
+	"app_id": "hello-lambda",
+	"display_name": "Hello, AWS Lambda!",
```
hanzei, line 3: Nit pick: Version is missing

examples/go/hello-lambda/Makefile
hanzei, line 1: Should we also rename and update `js/aws_hello`? Separate PR /Ticket?

go.mod
```diff
@@ -3,7 +3,9 @@ module github.com/mattermost/mattermost-plugin-apps
 go 1.16
 
 require (
+	github.com/aws/aws-lambda-go v1.19.1
 	github.com/aws/aws-sdk-go v1.38.2
+	github.com/awslabs/aws-lambda-go-api-proxy v0.10.0
```
hanzei, line 8: 2/5 that this should go into a seperate `go.mod` file just for the example to make it easier to copy and run

server/command/service.go
```diff
@@ -30,52 +30,138 @@ type service struct {
...
-		httpOut: httpOut,
-	}
-	conf := configService.GetConfig()
-	subCommands := s.getSubCommands()
-	var subTrigger []string
-	for t, c := range subCommands {
-		if c.debug && !conf.DeveloperMode {
-			continue
-		}
+type params struct {
```
hanzei, line 33: 0/5 `commandParams`?

server/command/service.go
```diff
@@ -30,52 +30,138 @@ type service struct {
...
+	all := map[string]commandHandler{
+		"info": {
+			f:            s.executeInfo,
+			autoComplete: model.NewAutocompleteData("info", "", "Display debugging information"),
+		},
+		"list": {
+			f:            s.executeList,
+			autoComplete: model.NewAutocompleteData("list", "", "List installed and registered apps"),
+		},
+		"uninstall": {
```
hanzei, line 58: Should the visibility be limited to system admins?

server/command/service.go
```diff
@@ -30,52 +30,138 @@ type service struct {
...
+			HelpText: "URL of the App's manifest",
+			Type:     model.AutocompleteArgTypeText,
+			Data: &model.AutocompleteTextArg{
+				Hint: "URL",
+			},
+			Required: true,
+		})
+	}
+
+	h.autoComplete.Arguments = append(h.autoComplete.Arguments, &model.AutocompleteArg{
```
hanzei, line 126: Is that option really needed for Cloud?

server/command/service.go
```diff
@@ -30,52 +30,138 @@ type service struct {
...
+			autoComplete: model.NewAutocompleteData("debug-bindings", "", "List bindings"),
+		}
+		all["debug-clean"] = commandHandler{
+			f:            s.executeDebugClean,
+			devOnly:      true,
+			autoComplete: model.NewAutocompleteData("debug-clean", "", "Delete all KV data"),
+		}
+		// TODO ticket: change to watch-manifest
+		all["debug-add-manifest"] = commandHandler{
+			f:            s.executeDebugAddManifest,
```
hanzei, line 82: Given that is coding is getting touched: Should `s.checkSystemAdmin` be added here for consistency?

server/plugin.go
```diff
@@ -73,35 +73,36 @@ func (p *Plugin) OnActivate() error {
...
 	// manifest store
-	conf := p.conf.GetConfig()
 	mstore := p.store.Manifest
 	mstore.Configure(conf)
-	// TODO: uses the default bucket name, do we need it customizeable?
-	manifestBucket := apps.S3BucketNameWithDefaults("")
-	err = mstore.InitGlobal(p.aws, manifestBucket, p.httpOut)
-	if err != nil {
-		return errors.Wrap(err, "failed to initialize the global manifest list from marketplace")
+	if conf.MattermostCloudMode {
```
hanzei, line 101: Does it make sense to also do that in dev mode given that the install command for AWS is available?

server/proxy/proxy.go
```diff
@@ -184,6 +191,16 @@ func (p *Proxy) upstreamForApp(app *apps.App) (upstream.Upstream, error) {
...
+	}
+}
+
+func isAppTypeSupported(conf config.Config, m *apps.Manifest) error {
+	if conf.DeveloperMode {
+		return nil
 	}
+	if conf.MattermostCloudMode != (m.AppType == apps.AppTypeAWSLambda) {
+		return utils.NewForbiddenError("AWS lambda apps are supported only in Mattermost Cloud")
+	}
```
hanzei, line 204: Similar HTTP apps should only run in On-Prem. Should there be a check to enforce that?

upstream/upaws/clean.go
```diff
@@ -0,0 +1,80 @@
...
+			}
+			log.Info("not found "+typ, "key", name)
+		} else {
+			log.Info("deleted "+typ, "key", name)
+		}
+		return nil
+	}
+
+	var err error
+	err = asAdmin.RemoveUserFromGroup(DefaultUserName, DefaultGroupName)
```
hanzei, line 27: Why not
```suggestion
	err := asAdmin.RemoveUserFromGroup(DefaultUserName, DefaultGroupName)
```

upstream/upaws/clean.go
```diff
@@ -0,0 +1,80 @@
...
+package upaws
+
+import (
+	"github.com/pkg/errors"
+
+	"github.com/mattermost/mattermost-plugin-apps/utils"
+)
+
+func CleanAWS(asAdmin Client, accessKeyID string, log Logger) error {
+	delete := func(typ string, name Name, del func(Name) error) error {
```
hanzei, line 13: Nice anonymous function :+1:

upstream/upaws/clean.go
```diff
@@ -0,0 +1,80 @@
...
+			if errors.Cause(err) != utils.ErrNotFound {
+				return err
+			}
+			log.Info("not found policy", "ARN", *policy.Arn)
+		} else {
+			log.Info("deleted policy", "ARN", *policy.Arn)
+		}
+	}
+
+	// TODO clean up the Lambda functions and S3 objects
```
hanzei, line 77: Ticket?

hanzei (changes requested):
Great work :+1: 

The code looks solid to me, but I'm running into the same issue as @DHaussermann in 1.

levb:
@hanzei @DHaussermann opening up the bucket to ACLs works for now, 
image

I propose that instead of making the manifest file public (which is causing the exception) we can just proxy the entire /manifests folder from the MM instance, making all manifests there visible. Separate PR?

hanzei:
@levb I'm struggling to understand what does changes with your PR. The ACLs weren't public before and should not have to, right?

levb:
@hanzei the provisioning of static assets shouldn't have changed at all, and I am not seeing them work from the server when the bucket is private. I question if they ever worked (we don't use provisioned icons in the apps at the moment)

hanzei (approved):
Tested again and works now :tada: 

Submitted by feedback in https://github.com/mattermost/mattermost-plugin-apps/pull/193 for quicker iteration

LGTM :+1:

DHaussermann:
@levb one issue that I see with this is that the `appsctl aws init` command cannot be run a 2nd time with a different bucket. If I set a different bucket and run the script again, as expected, I see `INFO[0001] using existing policy` `                         ARN="arn:aws:iam::900736139882:policy/mattermost-apps-invoke-policy"` which makes sense put the tool is aware that I am now using a different bucket. @jfrerich helped me look into this and noticed that in this case the ` "Resource":` section of the policy does not update to include the manifest and static folders for the bucket set when the command is run. 

For cases where someone maybe using multiple staging/test buckets the script should be able to accommodate adding the S3 resources to the policy.

On a minor note, there is a debug command for `add-manifest` that serves no purpose in On-Prem mode. Manifests can be stored to the KV but there is no way to use them for installation after as only the 1 step install from URL is functional.

upstream/upaws/aws.go
```diff
@@ -0,0 +1,98 @@
...
+
+import (
+	"os"
+	"text/template"
+)
+
+const (
+	AccessEnvVar          = "MM_APPS_AWS_ACCESS_KEY"           // nolint:gosec
```
stylianosrigas, line 9: @levb will this be used for APPs invoke?

upstream/upaws/aws.go
```diff
@@ -0,0 +1,98 @@
...
+			"Effect": "Allow",
+			"Principal": {
+				"Service": "lambda.amazonaws.com"
+			},
+			"Action": "sts:AssumeRole"
+		}
+	]
+}`
+
+const InitialInvokePolicyDocument = `{
```
stylianosrigas, line 48: @levb Have you tested this? You may need to add this section too
```json
{
            "Effect": "Allow",
            "Action": [
                "s3:ListBucket",
                "s3:ListBucketVersions",
                "s3:GetBucketLocation"
            ],
            "Resource": [
                ""arn:aws:s3:::{{.Bucket}}"
            ]
        }
```

server/proxy/install.go
```diff
@@ -81,17 +86,26 @@ func (p *Proxy) InstallApp(sessionID, actingUserID string, cc *apps.Context, tru
...
+		}
+
+		message = resp.Markdown
 	}
-	resp := p.Call(sessionID, actingUserID, creq)
-	if resp.Type == apps.CallResponseTypeError {
-		return nil, "", errors.Wrap(resp, "install failed")
+
+	if message == "" {
+		message = md.MD(fmt.Sprintf("Successfully install %s", app.AppID))
```
hanzei, line 104: My typo
```suggestion
		message = md.MD(fmt.Sprintf("Successfully installed %s", app.AppID))
```

upstream/upaws/clean.go
levb: What's the difference?

levb:
> For cases where someone maybe using multiple staging/test buckets the script should be able to accommodate adding the S3 resources to the policy.

@DHaussermann once the policy is initialized, they can manually edit it and add more bucket permissions.

upstream/upaws/aws.go
levb: Yes

upstream/upaws/aws.go
levb: Yes, we did (with **s, still a final round to be done). From https://docs.aws.amazon.com/AmazonS3/latest/API/API_GetObject.html

You need the s3:GetObject permission for this operation. For more information, see Specifying Permissions in a Policy. If the object you request does not exist, the error Amazon S3 returns depends on whether you also have the s3:ListBucket permission.

If you have the s3:ListBucket permission on the bucket, Amazon S3 will return an HTTP status code 404 ("no such key") error.

If you don’t have the s3:ListBucket permission, Amazon S3 will return an HTTP status code 403 ("access denied") error.

levb: Getting a 404 is fine

DHaussermann:
> Once the policy is initialized, they can manually edit it and add more bucket permissions.

Thanks Lev, giving the low likelihood that this is needed - This sounds very reasonable to me. I'll give this a try for the next test pass now 👍

cmd/appsctl/main.go
levb: Fixed, was a typo, and a very unfortunate one. Thx @hanzei for catching!

DHaussermann:
@hanzei I realize that after the merge of your work, the `apps list` command now shows you the records in the manifest when running in developer mode for on-prem. 
I have a concern that this can cause confusion when troubleshooting. For instances in the cloud the apps are only listed when the is a record in the config **and** a corresponding lambda is available so only installable apps are listed here. Unless we can also indicate in the table which apps are installable, this change may cause confusion. Any thoughts?

hanzei:
@DHaussermann For Cloud, all apps listed there are installable.

For on-prem, the lambda might get deleted and then an app can't be installed any longer. Given that this issue can also also happen while the app is running it seems like a different issue to me. Did I got your question correctly?

DHaussermann:
@hanzei yes, I believe if you understand the concern. My question would be, is it not preferable to keep the same logic for on-prem and have only installable apps listed in the table rather than showing whatever is in the plugin manifest?

server/proxy/install.go
```diff
@@ -81,17 +86,26 @@ func (p *Proxy) InstallApp(sessionID, actingUserID string, cc *apps.Context, tru
...
-		Call:    *apps.DefaultOnInstall.WithOverrides(app.OnInstall),
-		Context: cc,
+	var message md.MD
+	if app.OnInstall != nil {
+		creq := &apps.CallRequest{
+			Call:    *apps.DefaultOnInstall.WithOverrides(app.OnInstall),
+			Context: cc,
+		}
+		resp := p.Call(sessionID, actingUserID, creq)
+		if resp.Type == apps.CallResponseTypeError {
```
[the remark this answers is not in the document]
levb: Unfortunately, we'll need a little refactoring to properly differentiate "framework-introduced errors" such as 404 from the App-introduced errors. In the case on Install, 404s should be ignorable but they are not. See https://mattermost.atlassian.net/browse/MM-35152

hanzei:
@DHaussermann What is the difference between an "installable app" and one with a registered manifest?

server/plugin.go
```diff
@@ -98,7 +98,7 @@ func (p *Plugin) OnActivate() error {
 	// manifest store
 	mstore := p.store.Manifest
 	mstore.Configure(conf)
-	if conf.MattermostCloudMode {
+	if conf.MattermostCloudMode || conf.DeveloperMode {
```
hanzei, line 101: Actually, my initial comment was invalid. The marketplace install is only available in Cloud. Given that the store most likely won't work for devs, I'm inclined revert this change.
```suggestion
	if conf.MattermostCloudMode {
```

DHaussermann:
@hanzei unless I'm mistaken, for the cloud deployments - Apps are only listed in the table if there is also a lambda function with the corresponding name available. Therefore, apps are only listed in the table if they are available to install.

hanzei:
@DHaussermann That is not the case. Both for cloud and on-prem there can be apps listed that don't have a corresponding lambda function.

hanzei:
@DHaussermann I don't get which case you are describing. The lambda should work unless it get manually deleted, which IMO we should not deal with right now.

DHaussermann:
@hanzei sorry it seems maybe my concern above was not clear or I am making an incorrect assumption. 
1. My understanding is that with your changes merged in, Apps will always show in the table if they are listed in the manifest regardless of if there is also a lambda available.
2. This differs from Cloud installs where the record will show in the table if there is a record in the config **AND** the lambda is available.

Are these 2 points accurate?

hanzei:
1. Technical correct, but this should not happen. Apps running in AWS are added by `/apps install aws`. It fetches the manifest from S3 which is only present if `appsctl` was run successfully before. When should this case happen?
2. Technical not correct. If we miss configure things, the lambda might not be available.

hanzei:
@DHaussermann Is this conversation a blocker to merging the PR?

server/plugin.go
hanzei: Applying this changes as the current behavior also confused @DHaussermann

levb:
👍

server/command/install.go
```diff
@@ -11,36 +11,81 @@ import (
 
 	"github.com/mattermost/mattermost-plugin-apps/apps"
 	"github.com/mattermost/mattermost-plugin-apps/server/httpin/dialog"
-	"github.com/mattermost/mattermost-plugin-apps/server/utils"
 )
 
-func (s *service) executeInstall(params *params) (*model.CommandResponse, error) {
+func (s *service) executeInstallMarketplace(params *commandParams) (*model.CommandResponse, error) {
```
catalintomai, line 16: maybe add a comment for future readers that for this path (Marketplace) we've previously added the manifests to the store (using `InitGlobal`?).

server/plugin.go
```diff
@@ -73,74 +73,74 @@ func (p *Plugin) OnActivate() error {
...
-		p.mm.Log.Warn("APPS_INVOKE_AWS_ACCESS_KEY is not set. AWS apps won't work.")
+	conf := p.conf.GetConfig()
+	mode := "Self-managed"
+	if conf.MattermostCloudMode {
+		mode = "Mattermost Cloud"
 	}
-	secretKey := os.Getenv("APPS_INVOKE_AWS_SECRET_KEY")
-	if secretKey == "" {
-		p.mm.Log.Warn("APPS_INVOKE_AWS_SECRET_KEY is not set. AWS apps won't work.")
+	if conf.DeveloperMode {
```
catalintomai, line 81: what does it actually mean (practically) to be in both `MattermostCloudMode` and `DeveloperMode`?

server/config/service.go
```diff
@@ -87,30 +83,19 @@ func (s *service) reloadMattermostConfig() *model.Config {
...
 	mmconf := s.reloadMattermostConfig()
 
-	mattermostSiteURL := mmconf.ServiceSettings.SiteURL
-	if mattermostSiteURL == nil {
-		return errors.New("plugin requires Mattermost Site URL to be set")
+	newConfig := s.GetConfig()
+
+	// GetLicense silently drops an RPC error
+	// (https://github.com/mattermost/mattermost-server/blob/fc75b72bbabf7fabfad24b9e1e4c321ca9b9b7f1/plugin/client_rpc_generated.go#L864).
+	// When running in Mattermost cloud we must not fall back to the on-prem mode, so in case we get a nil retry once.
```
catalintomai, line 90: we can get nil the second time (and more) as well, no? Is the on-prem mode automatically the non-MattermostCloud mode?

levb: Yes, and at this point we assume "Self-managed", see https://github.com/mattermost/mattermost-plugin-apps/blob/181555e6ed4263c6e94a8a4e26706a1bc6192010/server/config/config.go#L124

hanzei: Should we log an error in that case?

server/plugin.go
hanzei: They are not mutually exclusive.

If an server is in `MattermostCloudMode` and in `DeveloperMode` it means that:
- There will be debug command like `/apps debug-list-bindings`
- `manifests.json` will be used to fetch manifest from S3
- There will only be a `/apps install marketplace` command

server/config/service.go
```diff
@@ -87,30 +83,19 @@ func (s *service) reloadMattermostConfig() *model.Config {
...
+	license := s.mm.System.GetLicense()
+	if license == nil {
+		license = s.mm.System.GetLicense()
```
hanzei, line 93:
```suggestion
		license = s.mm.System.GetLicense()
		if license == nil {
			s.mm.Log.Warn("Failed to fetch license two times. Falling back to on-prem mode.")
		}
```

server/config/service.go
hanzei: Suggestion: https://github.com/mattermost/mattermost-plugin-apps/pull/191#discussion_r647406090

DHaussermann:
Testing for this has completed. 
No further changes are needed yo on-prem mode.

I did find 1 point of concern when regression testing cloud mode. Install command - including for cloud - have been updated. I was not aware that clous install had also changed but, cloud install commands now look like `/apps install marketplace com.mattermost.servicenow`

@hanzei  this changes breaks the marketplace installation when the plugin is running in Cloud mode as the "_Install_" button is running the command without the installation type of `marketplace`

This is concerning to me because it means we need webapp changes to accommodate and will once again have a dependency on when can deploy based on when the webapp change is on Cloud. If this is the way we must implement the cloud installation command it does not block this PR. But we need a corresponding webapp PR and a fixed deployment date.   
cc @levb and @catalintomai

hanzei:
@DHaussermann We can add a temporary workaround so the old command still works. Let me do that in a separate PR if you don't mind.

DHaussermann:
@hanzei to clarify ☝️ you mean cloud mode will not require the installation type parameter (or fail-over to `marketplace` when not supplied)? 

This makes sense to me. We can then co-ordinate this change separately.

levb:
How about we merge this PR for the on-prem, and if we still need to figure out the command syntax for 0.7 we do a separate PR/ticket for it?

hanzei:
@DHaussermann Exactly.

@levb Sounds good to me.

DHaussermann (approved):
Tested and passed
- Tested deployment to my personal AWS instance
- Tested that multiple apps can be deployed and are functional
- Tested use of appctl tool as documented
- Worked with Dev on corresponding doc PR as an instruction set
- Confirmed the app runs in cloud mode when legacy AWS keys are present
- Regression tested Cloud mode
- As per @hanzei and @levb we will be decided and addressed in a separate PR before 0.7.0 release. Created https://mattermost.atlassian.net/browse/MM-36294 to address this.

LGTM!
Please merge.

hanzei:
Thanks for the testing @DHaussermann :+1: 

I'm going to merge this PR as it's a great improvement in it self and blocked some other PR right now due to code dependencies and merge conflicts.

@stylianosrigas @catalintomai Please leave any outstanding comments in this PR and I will resolve them in a separate one.

PR #193, opened 2021-05-28
Implement feedback on AWS support

hanzei:
#### Summary
This PR is based on https://github.com/mattermost/mattermost-plugin-apps/pull/191 and implements some feedback of mine. I've went through the logging calls of the whole code base and fixed some inconsistencies. 

#### Ticket Link
None


examples/go/hello-lambda/Makefile
```diff
@@ -5,7 +5,7 @@ all: run
 .PHONY: build-linux
 ## build-linux: build the executable for linux
 build-linux:
-	GOOS=linux GOARCH=amd64 go build -o dist/hello-lambda
+	CGO_ENABLED=0 GOOS=linux GOARCH=amd64 go build -o dist/hello-lambda
```
hanzei, line 8: We need to static link otherwise the AWS runtime fails to find a C library

examples/go/hello-lambda/main.go
```diff
@@ -73,7 +80,10 @@ func send(w http.ResponseWriter, req *http.Request) {
 	}
 	mmclient.AsBot(c.Context).DM(c.Context.ActingUserID, message)
 
-	json.NewEncoder(w).Encode(apps.CallResponse{})
+	json.NewEncoder(w).Encode(apps.CallResponse{
+		Type:     apps.CallResponseTypeOK,
+		Markdown: "Created a post in your DM channel.",
+	})
```
hanzei, line 86: For consistency with the other example

larkox (approved):

examples/go/hello-lambda/Makefile
levb: what's the CGO dependency?

levb (approved):
All good, but I don't understand the turning off of CGO for the lambda

examples/go/hello-lambda/Makefile
hanzei: I don't know, but 
```
$ ldd dist/com.mattermost.apps/server/dist/plugin-linux-amd64
	linux-vdso.so.1 (0x00007ffd43bf6000)
	libpthread.so.0 => /usr/lib/libpthread.so.0 (0x00007fe1cdf0c000)
	libc.so.6 => /usr/lib/libc.so.6 (0x00007fe1cdd3f000)
	/lib64/ld-linux-x86-64.so.2 => /usr/lib64/ld-linux-x86-64.so.2 (0x00007fe1cdf55000)
```

clearly shows that CGO is used _somewhere_. Maybe it's the server?

levb: Also, why did it work for me without it (in AWS?)

hanzei: @levb I have no idea

PR #197, opened 2021-06-02
[MM-30595] Add support for relative icon path to bindings and forms

hanzei:
#### Summary
With this PR apps can define `Icon` for both bindings and forms using a relative path.

Please note that no validation of the asset is done due to performance concerns. Invalid icon with show as invalid in the UI. Only malformed path containing e.g. `/../` are logged and ignored.

Tested with hello-world app.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-30595

### DIFF
--- a/apps/binding.go
+++ b/apps/binding.go
@@ -65,7 +65,7 @@ package apps
 //              "bindings": [
 //                  {
 //                      "location": "send-button",
-//                      "icon": "http://localhost:8080/static/icon.png",
+//                      "icon": "icon.png",
 //                      "label":"send hello message",
 //                      "call": {
 //                          "path": "/send-modal"
@@ -77,7 +77,7 @@ package apps
 //              "location": "/command",
 //              "bindings": [
 //                  {
-//                      "icon": "http://localhost:8080/static/icon.png",
+//                      "icon": "icon.png",
 //                      "description": "Hello World app",
 //                      "hint":        "[send]",
 //                      "bindings": [
--- a/server/proxy/proxy.go
+++ b/server/proxy/proxy.go
@@ -8,10 +8,12 @@ import (
 	"io"
 	"net/http"
 	"path"
+	"strings"
 
 	"github.com/pkg/errors"
 
 	"github.com/mattermost/mattermost-plugin-apps/apps"
+	"github.com/mattermost/mattermost-plugin-apps/server/config"
 	"github.com/mattermost/mattermost-plugin-apps/server/upstream"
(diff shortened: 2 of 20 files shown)


levb (approved):
some nits

apps/binding.go
```diff
@@ -65,7 +65,7 @@ package apps
 //              "bindings": [
 //                  {
 //                      "location": "send-button",
-//                      "icon": "http://localhost:8080/static/icon.png",
```
larkox, line 68: Should we keep some icons as absolute paths as an example of both?

larkox, line 68: Should we create "broken examples" for the sake of testing?

larkox (approved):
LGTM, just two small considerations.

server/proxy/proxy.go
```diff
@@ -61,13 +64,40 @@ func (p *Proxy) Call(sessionID, actingUserID string, creq *apps.CallRequest) *ap
...
+// If icon is an absolute URL, it's not changed.
+// Otherwise assume it's a path to a static asset and the static path URL prepended.
+func convertToStaticPath(conf config.Config, appID apps.AppID, icon string) (string, error) {
+	if !strings.HasPrefix(icon, "http://") && !strings.HasPrefix(icon, "https://") {
+		cleanIcon, err := utils.CleanStaticPath(icon)
+		if err != nil {
+			return "", errors.Wrap(err, "invalid icon path")
+		}
+
+		icon = conf.StaticURL(appID, cleanIcon)
```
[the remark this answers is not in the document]
hanzei: 2/5 that `StaticURL` should stay in `conf` as it's similar to `AppURL`. Given that I'm inclined to leave normalizeStaticPath as it is.

apps/binding.go
hanzei: The example should show the idiomatic way of using things. Hence, the icons should all be relative paths as this is the way we encourage devs to use them.

hanzei: Good point :+1:

hanzei:
Docs PR: https://github.com/mattermost/mattermost-developer-documentation/pull/840

hanzei:
@DHaussermann QA review for a downstream PR was done by @jfrerich in https://github.com/mattermost/mattermost-app-zendesk/pull/60#issuecomment-853968843. Do you still want to test this?

DHaussermann (approved):
Peer tested as part of a downstream PR

Will re-test for 0.7.0 release testing. 

LGTM!

PR #198, opened 2021-06-02
[MM-36056] Add modes info

hanzei:
#### Summary
Screenshot from 2021-06-02 13-43-06

Based on https://github.com/mattermost/mattermost-plugin-apps/pull/191

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-36056

### DIFF
--- a/server/command/info.go
+++ b/server/command/info.go
@@ -13,12 +13,15 @@ import (
 func (s *service) executeInfo(params *params) (*model.CommandResponse, error) {
 	conf := s.conf.GetConfig()
 	resp := md.Markdownf("Mattermost Apps plugin version: %s, "+
-		"[%s](https://github.com/mattermost/%s/commit/%s), built %s\n",
+		"[%s](https://github.com/mattermost/%s/commit/%s), built %s, Cloud Mode: %t, Developer Mode: %t\n",
 		conf.Version,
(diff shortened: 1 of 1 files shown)


levb (approved):

server/command/info.go
```diff
@@ -13,12 +13,15 @@ import (
 func (s *service) executeInfo(params *params) (*model.CommandResponse, error) {
 	conf := s.conf.GetConfig()
 	resp := md.Markdownf("Mattermost Apps plugin version: %s, "+
-		"[%s](https://github.com/mattermost/%s/commit/%s), built %s\n",
+		"[%s](https://github.com/mattermost/%s/commit/%s), built %s, Cloud Mode: %t, Developer Mode: %t\n",
```
larkox, line 16: Somewhere the mode string is generated (something like "Cloud Developer" or something like that), I think for some log in the initialization. Would that be more user friendly? Or do we prefer the true/false statements?

larkox (approved):
LGTM, just one consideration.

server/command/info.go
hanzei: I choose the boolean because it given you the full list of possible permutations. That help with debugging as you don't have to thing about which other options are possible.

PR #201, opened 2021-06-04
[MM-35762] Clean call path before making a request

hanzei:
#### Summary
Before making a call request check if `Path` can be used to traverse to a different URL.

Strictly speaking this isn't needed for apps running in AWS right now, but I think it's good practice to defend against path traversal attacks anyway.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-35762


server/proxy/proxy.go
```diff
@@ -42,6 +43,17 @@ func (p *Proxy) Call(sessionID, actingUserID string, creq *apps.CallRequest) *ap
 	if err != nil {
 		return apps.NewProxyCallResponse(apps.NewErrorCallResponse(err), metadata)
 	}
+
+	if creq.Path == "" || creq.Path[0] != '/' {
```
catalintomai, line 47: maybe this block can be factored out into some utility method re: `.../server/http/gateway/static.go::cleanStaticPath()`

levb, line 47: Also, what's wrong with a "/" as a call path? I know it is unusual, but why not?

levb (changes requested):

server/proxy/proxy.go
hanzei: The code is a bit different and `cleanStaticPath` is used for - well - well path of static assets. I'm fine with adding a  `cleanPath` method to keep things DRY, but I'm hesitant to mix up file paths and call paths.

catalintomai: I was referring to the code snippet of `cleanStaticPath` that was doing the same as here, not to the whole method.

levb (approved):

server/proxy/proxy.go
hanzei: Refactored the code in dd0244e161b4713bfef6650b496e7162713e5d63

larkox (approved):

srkgupta (approved):
Thanks @hanzei. The changes LGTM.

hanzei:
@DHaussermann I'm going to skip QA review here as this is a pure security issue. If the PR breaks something, it will break loud and you will notice it in release testing.

PR #202, opened 2021-06-10
[MM-36302] Update autocomplete when configuration changes

hanzei:
#### Summary
In order for the slash command autocomplete to update the command service needs to re-register the slash command.

I'm not entirely happy with the `Plugin` taking care of deciding which services need to `Reconfigure`. Rather the services register them self at the config store for a `Reconfigure`. But the change seems extends the scope of this PR.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-36302

### DIFF
--- a/server/command/service.go
+++ b/server/command/service.go
@@ -44,7 +44,7 @@ type commandHandler struct {
 	autoComplete *model.AutocompleteData
 }
 
-func (s *service) allCommands(conf config.Config) map[string]commandHandler {
+func (s *service) allSubCommands(conf config.Config) map[string]commandHandler {
 	uninstallAC := model.NewAutocompleteData("uninstall", "", "Uninstall an app")
 	uninstallAC.AddTextArgument("ID of the app to uninstall", "appID", "")
 	uninstallAC.RoleID = model.SYSTEM_ADMIN_ROLE_ID
@@ -200,7 +200,7 @@ func (s *service) Configure(conf config.Config) {
 }
 
 func (s *service) registerCommand(conf config.Config) error {
-	subCommands := s.allCommands(conf)
+	subCommands := s.allSubCommands(conf)
 	var subs []string
 	for t := range subCommands {
 		subs = append(subs, t)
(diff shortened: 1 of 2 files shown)


larkox (approved):

server/command/service.go
```diff
@@ -262,7 +276,8 @@ func (s *service) ExecuteCommand(pluginContext *plugin.Context, commandArgs *mod
 }
 
 func (s *service) handleMain(in *commandParams) (*model.CommandResponse, error) {
-	return s.runSubcommand(s.allCommands(), in)
+	conf := s.conf.GetConfig()
+	return s.runSubcommand(s.allCommands(conf), in)
```
catalintomai, line 280: why is this called `allCommands` and not `allSubCommands`  (since we are calling `runSubcommand` on the set)?

catalintomai (approved):

DHaussermann:
Release testing was done against this PR
- Bot Icons can be used
- Debug commands are hidden when dev mode is disabled 
- Tested install process successfully with Service Now 
- Zendesk may have an upgrade issue that requires further investigation

For Cloud mode:
- True when Cloud license applied OR legacy AWS keys are present
- Uses APPS_INVOKE_AWS_ACCESS_KEY and APPS_INVOKE_AWS_SECRET_KEY to invoke
- Will show listed AWS apps available for install with /apps list command 

For On-Prem mode 
- True when no cloud license and no legacy AWS keys
- Uses MM_APPS_AWS_ACCESS_KEY and MM_APPS_AWS_SECRET_KEY to invoke
- Uninstall deletes stored manifest and remove app record from the table 

One point to bring up is that I do not yet see marketplace icons as expected. This is not a show stopper for `v0.7.0` I'll follow up with @hanzei 

LGTM!

DHaussermann (approved):

server/command/service.go
hanzei: Good point, updated in 5a0d918376e50d705f1a0cdf349f90beddb0c2b3

PR #203, opened 2021-06-10
Filter bindings and commands on proxy

larkox:
#### Summary
This PR mainly solve the following issues:
- Commands with spaces on their labels
- Field names with spaces on their labels
- Field names repeated
- Options repeated
- Fields with no names
- Options with no label nor value

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-33436
https://mattermost.atlassian.net/browse/MM-33533


server/proxy/bindings.go
```diff
@@ -136,6 +137,18 @@ func (p *Proxy) scanAppBindings(app *apps.App, bindings []*apps.Binding, locPref
...
 		}
 
+		if fql.In(apps.LocationCommand) {
+			label := b.Label
+			if label == "" {
+				label = string(b.Location)
+			}
+
+			if strings.ContainsAny(label, " \t") {
+				// p.mm.Log.Debug("Binding validation error: Command label has multiple words", "app", app.Manifest.AppID, "location", b.Location)
```
hanzei, line 147: 1/5 that we should log an error

server/proxy/proxy.go
```diff
@@ -98,6 +100,104 @@ func normalizeStaticPath(conf config.Config, appID apps.AppID, icon string) (str
...
+func (p *Proxy) cleanForm(form *apps.Form) {
+	if form == nil {
+		return
+	}
+
+	toRemove := []int{}
+	usedLabels := map[string]bool{}
+	for i, field := range form.Fields {
+		if field.Name == "" {
+			toRemove = append([]int{i}, toRemove...)
```
hanzei, line 117: Why is the order reversed?

hanzei (approved):
LGTM, two non blocking questions below

server/proxy/proxy.go
larkox: So we can then remove the elements from the list without any problem in the next for. We could do a normal descending for instead of a range and it will probably be better performant, but the code would be less clear (I think).

server/proxy/bindings.go
larkox: I left it as a comment because we are not logging also a message a few lines above when the app has not allowed that binding. I can uncomment both, but I fear a logs flood in case of a "not so well developed app". What do you think?

hanzei: Given that apps are in "developers preview" right now, developers should be our focus group. Hence, if the log messages help devs, I think we should add them.

server/proxy/proxy.go
hanzei: Got it. Makes sense :+1: Maybe add a comment?

server/proxy/bindings.go
```diff
@@ -164,15 +177,26 @@ func (p *Proxy) scanAppBindings(app *apps.App, bindings []*apps.Binding, locPref
 			}
 		}
 
+		// First level of Channel Header
+		if fql == apps.LocationChannelHeader.Make(b.Location) {
+			// Must have an icon on webapp to show the icon
+			if b.Icon == "" && userAgent == "webapp" {
```
hanzei, line 183: What do we gain by only throwing a log message when a webapp user requests the bindings? Does it make sense to always throw the error as come clients (webapp) will have issue with such binding?

hanzei (approved):
One non-blocking question

server/proxy/bindings.go
larkox: In this block we are doing two things: log the error, and continue, so this binding is filtered. We could separate the conditions, but we would have an even deeper if chain.

hanzei: I'm inclined to both filter and log even if the userAgent is not `webapp`. If a dev developers an app only using mobile, them will never notice that something is off.

hanzei: Good point. Please consider my point as irrelevant.

levb (approved):

levb:
@DHaussermann I am inclined to skip the manual QA on this PR since it comes with comprehensive unit tests and is corrective in in nature.

@justinegeffen How should we handle the docs impact?

justinegeffen:
Thanks for the tag, @levb! It looks like [the bindings content](https://developers.mattermost.com/integrate/apps/bindings/) may need updating. I can't quite tell what else may need updating. If you're open to providing the updated content I'll get it published. :)

larkox:
@DHaussermann In order to test this, you can use the test app (pushed some minor changes to master today). With each thing to test, I will tell you what you should NOT see.

- Commands with spaces on their labels (`/test invalid with_multiword command`)
- Field names with spaces on their labels (`/test invalid with_multiword_field --multiword field`, you should see it as if no form was provided (just execute))
- Field names repeated (`/test invalid with_same_field_names` should show only one field)
- Options repeated (`/test invalid with_same_option_names` should show only one option)
- Fields with no name (`/test invalid with_noname_field`, you should see it as if no form was provided (just execute))
- Options with no label nor value (`/test invalid with_no_name_option --select` should show only the sv2 option)
- Channel header element with no icon on webapp (mobile should show) (`ERROR_with_no_icon` should not appear)

All of these should have relevant debug messages on the system console when the bindings are fetched.

levb:
@larkox @DHaussermann can we please try to get this in soon? It's a v. useful PR.

larkox:
@DHaussermann Friendly reminder for review

DHaussermann (approved):
Tested and passed
Deployed this along with  https://github.com/mattermost/mattermost-webapp/pull/8249
- Availability of the subcommands is working as expected
- In all cases the test steps provided by dev match the behavior I'm seeing
- No issues found

LGTM!

Thanks @larkox for providing test steps on this one.

PR #211, opened 2021-07-02
[MM-36659] Check for valid HomepageURL

hanzei:
#### Summary
Every apps needs to have a valid `HomepageURL`. This is now checked and verified. HTTP apps also need an `HTTPRootURL`.

The PR also fixes two build issue:
- The minimum go version needs to be 1.6
- All `go test` command should run against all packages, not only the ones in `./server`

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-36659

### DIFF
--- a/apps/manifest.go
+++ b/apps/manifest.go
@@ -132,11 +132,11 @@ func (f AWSLambda) IsValid() error {
 	if f.Name == "" {
 		return utils.NewInvalidError("aws_lambda name must not be empty")
 	}
-	if f.Runtime == "" {
-		return utils.NewInvalidError("aws_lambda runtime must not be empty")
-	}
 	if f.Handler == "" {
 		return utils.NewInvalidError("aws_lambda handler must not be empty")
+	}
+	if f.Runtime == "" {
+		return utils.NewInvalidError("aws_lambda runtime must not be empty")
 	}
 	return nil
 }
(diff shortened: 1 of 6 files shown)


mickmister (changes requested):

apps/manifest.go
```diff
@@ -183,6 +183,15 @@ func (m Manifest) IsValid() error {
...
 	}
 
+	if m.HomepageURL == "" {
+		return utils.NewInvalidError(errors.New("homepage_url is empty"))
+	}
+
+	_, err := url.Parse(m.HomepageURL)
+	if err != nil {
+		return utils.NewInvalidError(errors.Wrapf(err, "homepage_url invalid: %q", m.HomepageURL))
+	}
```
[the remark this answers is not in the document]
hanzei: Good point. I've added helper method to the `utils` package that also returns an error for better use.

apps/manifest.go
```diff
@@ -187,8 +186,7 @@ func (m Manifest) IsValid() error {
 		return utils.NewInvalidError(errors.New("homepage_url is empty"))
 	}
 
-	_, err := url.Parse(m.HomepageURL)
-	if err != nil {
+	if err := utils.IsValidHTTPURL(m.HomepageURL); err != nil {
```
mickmister, line 189: This function is available from `mattermost-server/model`. Do we want to import it from there?

mickmister (approved):

apps/manifest.go
hanzei: I don't think so for two reasons:

1. It's not worth including the full package from `model` for a single methods. This follows "A little copying is better than a little dependency.".
2. The `model` method doesn't return an  `error`, which I don't like, because we can't return a meaningful on to the user then.

catalintomai (approved):

hanzei:
@DHaussermann Given the good test coverage I'm inclined to merge this PR without QA review. The test steps would be:

1. Edit the manifest of an app of contain an invalid HomepageURL like `ssl://example.org`
2. Install the app
3. Notice that the installation fails with an error.

hanzei:
@mickmister Would you please merge this PR after QA is complete?

levb:
@DHaussermann inclined to merge without manual QA

levb (approved):

DHaussermann (approved):
Agreed @levb I can circle back to this in release testing as a precaution.
Please merge.

LGTM!

PR #217, opened 2021-07-22
[MM-3384] wrap manifest json data read error with better description

jfrerich:
#### Summary
When a use tries installing an app with an invalid manifest data, the following error is presented.
image

This PR wraps the error with a more informative error.
image

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-37339

### DIFF
--- a/server/command/install.go
+++ b/server/command/install.go
@@ -74,7 +74,7 @@ func (s *service) executeInstallHTTP(params *commandParams) (*model.CommandRespo
 
 	m, err := apps.ManifestFromJSON(data)
 	if err != nil {
-		return errorOut(params, err)
+		return errorOut(params, errors.Wrap(err, "unable to read `manifest json` from "+manifestURL))
 	}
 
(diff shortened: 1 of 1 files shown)


server/command/install.go
```diff
@@ -74,7 +74,7 @@ func (s *service) executeInstallHTTP(params *commandParams) (*model.CommandRespo
 
 	m, err := apps.ManifestFromJSON(data)
 	if err != nil {
-		return errorOut(params, err)
+		return errorOut(params, errors.Wrap(err, "unable to read `manifest json` from "+manifestURL))
```
levb, line 77: `"unable to decode " + manifestURL` seems more intention revealing and succinct?

levb (changes requested):

server/command/install.go
jfrerich: My intention was to reveal to the user that an expected `manifest json` structure was expected, but agree it is possibly over descriptive.  Updated to your suggestion.

levb (approved):

larkox (approved):

PR #218, opened 2021-07-23
Add i18n and add translation strings to the command package

larkox:
#### Summary
This would be the first step for the localisation of the plugin.

It follows the same pattern as the Jitsi plugin.

Only the command package and some minor strings in other places are in the translations. We should create tickets to convert all user facing strings to i18n, package by package.

This has nothing to do with App localization. Future work will be done on that front.

Documentation on how to contribute in the translations should be added.

#### Ticket Link
None


mmclient/http.go
```diff
@@ -12,8 +13,8 @@ type httpClient struct {
 	mm *model.Client4
 }
 
-func NewHTTPClient(mm *pluginapi.Client, conf config.Config, sessionID, actingUserID string) (Client, error) {
-	client, err := utils.ClientFromSession(mm, conf.MattermostSiteURL, sessionID, actingUserID)
+func NewHTTPClient(mm *pluginapi.Client, conf config.Config, sessionID, actingUserID string, i18nBundle *i18n.Bundle) (Client, error) {
```
levb, line 16: 1/5 we should consolidate `mm`, `conf`, `i18n`, and soon `log` into a struct, and pass it around/inherit in the services (it'd also implement a `Configure()`

Separate PR.

levb (approved):
We should tool for creating/wrapping international errors better, separate PR.

mmclient/http.go
mickmister: @levb Do we need tickets for this and your concern above?

.gitignore
```diff
@@ -15,3 +15,6 @@ dist/
 /server/httpin/restapi/data/
 
 node_modules
+
+# translatios
```
mickmister, line 19:
```suggestion
# translations
```

mickmister:
@larkox Do we need a ticket for documentation for contributing to translations as noted in the description? Also will this be integrated with the translation servers?

assets/i18n/active.en.json
```diff
@@ -0,0 +1,39 @@
...
+  "apps.command.error.invalidArguments": "invalid arguments to command.Handler. Please contact your system administrator",
+  "apps.command.error.mustBeAdmin": "you need to be a system admin to run this command",
+  "apps.command.error.needSubCommand": "expected a (sub-)command",
+  "apps.command.error.noSubCommand": "no subcommand specified, nothing to do",
+  "apps.command.info.rest": "{{.URL}}, built {{.BuildDate}}, Cloud Mode: {{.CloudMode}}, Developer Mode: {{.DeveloperMode}}",
+  "apps.command.info.version": "Mattermost Apps plugin version: {{.Version}}, ",
+  "apps.command.install.appID": "you must specify the app id",
+  "apps.command.install.aws.error.manifest": "failed to get manifest from S3",
+  "apps.command.install.aws.error.version": "you must specify the app version",
+  "apps.command.install.error.appID": "you must specify the app id",
```
mickmister, line 18: `you must specify the app id` shows up twice here. Should it be made into one entry?

build/custom.mk
```diff
@@ -1,5 +1,20 @@
...
 
+default: all
+
+i18n-extract-server:
+	@goi18n extract -format json -outdir assets/i18n/ server/
+	@for x in assets/i18n/active.*.json; do echo $$x | sed 's/active/translate/' | sed 's/^/touch /' | bash; done
+	@goi18n merge -format json -outdir assets/i18n/ assets/i18n/active.*.json
+	@echo "Please update your assets/i18n/translate.*.json files and execute \"make i18n-merge-server\""
+	@echo "If you don't want to change any locale file, simply remove the assets/i18n/translate.??.json file before calling \"make i18n-merge-server\""
+	@echo "If you want to add a new language (for example french), simple run \"touch assets/i18n/active.fr.json\" and then run the \"make i18n-extract-server\" again"
```
mickmister, line 11:
```suggestion
	@echo "If you want to add a new language (for example french), simply run \"touch assets/i18n/active.fr.json\" and then run the \"make i18n-extract-server\" again"
```

utils/utils.go
```diff
@@ -79,10 +81,14 @@ func LoadSession(mm *pluginapi.Client, sessionID, actingUserID string) (*model.S
...
-func ClientFromSession(mm *pluginapi.Client, mattermostSiteURL, sessionID, actingUserID string) (*model.Client4, error) {
+func ClientFromSession(mm *pluginapi.Client, mattermostSiteURL, sessionID, actingUserID string, i18nBundle *i18n.Bundle) (*model.Client4, error) {
+	loc := i18nBundle.GetUserLocalizer(actingUserID)
 	session, err := LoadSession(mm, sessionID, actingUserID)
 	if err != nil {
-		return nil, errors.Wrap(err, "failed to load session")
+		return nil, errors.Wrap(err, i18nBundle.LocalizeDefaultMessage(loc, &nsi18n.Message{
+			ID:    "apps.command.list.table.header",
+			Other: "failed to load session",
+		}))
```
mickmister, line 91: This message id doesn't seem right. I don't see `failed to load session` in the i18n json file either

server/command/service.go
```diff
@@ -265,29 +269,60 @@ func (s *service) ExecuteCommand(pluginContext *plugin.Context, commandArgs *mod
...
 	}
 
 	conf := s.conf.GetMattermostConfig().Config()
 	enableOAuthServiceProvider := conf.ServiceSettings.EnableOAuthServiceProvider
 	if enableOAuthServiceProvider == nil || !*enableOAuthServiceProvider {
-		return errorOut(params, errors.Errorf("the system setting `Enable OAuth 2.0 Service Provider` needs to be enabled in order for the Apps plugin to work. Please go to %s/admin_console/integrations/integration_management and enable it.", commandArgs.SiteURL))
+		return s.errorOut(params, errors.New(s.i18n.LocalizeWithConfig(loc, &nsi18n.LocalizeConfig{
+			DefaultMessage: &nsi18n.Message{
+				ID:    "apps.command.enableOAuth",
+				Other: "the system setting `Enable OAuth 2.0 Service Provider` needs to be enabled in order for the Apps plugin to work. Please go to {{.URL}} and enable it.",
```
mickmister, line 286: Can we capitalize this sentence and the next one pertaining to bot accounts?

mickmister (approved):
@larkox Awesome work getting the framework in place, and translating everything! I have a few non-blocking requests/questions

server/command/service.go
larkox: We shouldn't. They are passed to error out, which will form the sentence as `Command failed. Error: **{{.Error}}**`.

server/plugin.go
```diff
@@ -65,7 +69,13 @@ func (p *Plugin) OnActivate() error {
 		return errors.Wrap(err, "failed to ensure bot account")
 	}
 
-	p.conf = config.NewService(mm, p.BuildConfig, botUserID)
+	i18nBundle, err := i18n.InitBundle(p.API, filepath.Join("assets", "i18n"))
+	if err != nil {
+		return err
+	}
+	p.i18n = i18nBundle
```
hanzei, line 76: Nit: Does the `Plugin` need to store it? Is storing it in `Service` enough?

server/proxy/uninstall.go
```diff
@@ -4,20 +4,28 @@
 package proxy
 
 import (
-	"fmt"
-
+	"github.com/nicksnyder/go-i18n/v2/i18n"
```
hanzei, line 7: 1/5 that it's cleaner to use only use `github.com/mattermost/mattermost-plugin-api/i18n` and drop the dependency on `github.com/nicksnyder/go-i18n/v2/i18n`.

server/proxy/uninstall.go
```diff
@@ -4,20 +4,28 @@
...
-		return "", errors.Wrapf(err, "failed to get app. appID: %s", appID)
+		return "", errors.Wrap(err, p.conf.I18N().LocalizeWithConfig(loc, &i18n.LocalizeConfig{
+			DefaultMessage: &i18n.Message{
+				ID:    "apps.uninstall.error.getApp",
+				Other: "failed to get app. appID: {{.AppID}}",
+			},
+			TemplateData: map[string]string{
+				"AppID": string(appID),
+			},
+		}))
```
hanzei, line 28: Using `errors.Wrap` to return a /localized/ error might cause issues as the whole stack needs to know the users local in order to localize it's part. It's also worth noting that `GetUserLocalizer` costs one RPC call.

I was wondering if it would make sense to return a `i18n.LocalizeConfig` instead of an `error` and localize it on the top of the stack. Maybe even embedded into some kind of `UserError` struct.

Maybe my concerns aren't blocking for this PR, but I would like to double check that we are starting with the right pattern.

server/proxy/uninstall.go
```diff
@@ -35,43 +43,99 @@ func (p *Proxy) UninstallApp(client mmclient.Client, sessionID string, cc *apps.
...
 
 	if message == "" {
-		message = fmt.Sprintf("Uninstalled %s", app.DisplayName)
+		message = p.conf.I18N().LocalizeWithConfig(loc, &i18n.LocalizeConfig{
+			DefaultMessage: &i18n.Message{
+				ID:    "apps.uninstall.default",
+				Other: "Uninstalled {{.AppID}}",
+			},
+			TemplateData: map[string]string{
+				"AppID": app.DisplayName,
```
hanzei, line 52: Should this use "DisplayName"?

hanzei, line 52: Maybe out of scope of this PR: Should `DisplayName` or `AppID` be used in the command outputs? The code base it not consistent on that.

server/proxy/uninstall.go
```diff
@@ -35,43 +43,99 @@ func (p *Proxy) UninstallApp(client mmclient.Client, sessionID string, cc *apps.
...
+		}))
 	}
 
 	// delete app
 	if err = p.store.App.Delete(app.AppID); err != nil {
-		return "", errors.Wrapf(err, "can't delete app - %s", app.AppID)
+		return "", errors.Wrap(err, p.conf.I18N().LocalizeWithConfig(loc, &i18n.LocalizeConfig{
+			DefaultMessage: &i18n.Message{
+				ID:    "apps.uninstall.error.delete",
+				Other: "can't delete app - {{.AppID}}",
```
hanzei, line 105:
```suggestion
				Other: "can't delete app {{.AppID}}",
```

server/proxy/uninstall.go
```diff
@@ -35,43 +43,99 @@ func (p *Proxy) UninstallApp(client mmclient.Client, sessionID string, cc *apps.
...
 	}
 
 	// in on-prem mode the manifest need to be deleted as every install add a manifest anyway
 	if !conf.MattermostCloudMode {
 		if err = p.store.Manifest.DeleteLocal(app.AppID); err != nil {
-			return "", errors.Wrapf(err, "can't delete manifest for uninstalled app - %s", app.AppID)
+			return "", errors.Wrap(err, p.conf.I18N().LocalizeWithConfig(loc, &i18n.LocalizeConfig{
+				DefaultMessage: &i18n.Message{
+					ID:    "apps.uninstall.error.deleteManifest",
+					Other: "can't delete manifest for uninstalled app - {{.AppID}}",
```
hanzei, line 119:
```suggestion
					Other: "can't delete manifest for uninstalled app {{.AppID}}",
```

server/proxy/uninstall.go
```diff
@@ -35,43 +43,99 @@ func (p *Proxy) UninstallApp(client mmclient.Client, sessionID string, cc *apps.
...
 		}
 	}
 
 	// remove data
 	if err = p.store.AppKV.DeleteAll(app.BotUserID); err != nil {
-		return "", errors.Wrapf(err, "can't delete app data - %s", app.AppID)
+		return "", errors.Wrap(err, p.conf.I18N().LocalizeWithConfig(loc, &i18n.LocalizeConfig{
+			DefaultMessage: &i18n.Message{
+				ID:    "apps.uninstall.error.deleteData",
+				Other: "can't delete app data - {{.AppID}}",
```
hanzei, line 133:
```suggestion
				Other: "can't delete app data {{.AppID}}",
```

hanzei (commented):
A few comments below.

hanzei:
Great work here @larkox :+1: Looking forward to seeing more languages getting added.

server/proxy/uninstall.go
larkox: In order for the automation to work, we need to use the `i18n.LocalizeConfig` structure from the `nicksnyder` package, so as far as I know, we cannot get rid of this dependency.

server/proxy/uninstall.go
larkox: We probably should have a deeper thought about this.
If were to return a special error that implements the error interface, we would be in the same situation, since this error interface would need the localizer, so the `.Error()` function can return the localized string.

If we were to return all around the LocalizeConfig, concatenating errors will be also challenging. We could return a list of localize config, and assume that it is a list of errors that have to be connected.

Another option is to fill out the localizer at the entry of the stack, and pass it along to the functions.

Thoughts?

server/proxy/uninstall.go
larkox: I feel app id is more unique, and therefore should be used for most of these. Not sure though. I don't mind changing these in this PR if we get to a decision.

server/proxy/uninstall.go
hanzei: The plugin api package also exports these: https://pkg.go.dev/github.com/mattermost/mattermost-plugin-api@v0.0.19/i18n#LocalizeConfig

server/proxy/uninstall.go
hanzei: To be more clear: The `UserError` is had in mind won't implement `Error` for the reason you mentioned.

Concatenating errors is a good point. But maybe we shouldn't do that at all as it's more for us developers then for end users. Passing the localizer seems like a good option if we want to continue with your approach.

I like that returning something like `UserError` makes it clear that this output is for a user while that is not clear for an generic `Error`.

server/proxy/uninstall.go
hanzei: @levb Do you have thoughts on this? I'm 2/5 that `AppID` is a technical necessity, which we shouldn't show to the user if possible.

server/proxy/uninstall.go
larkox: Not sure why, but using those make the extraction tool not to find the strings. My guess is that the reflection mechanism thinks that they are different types.

larkox:
@hanzei @catalintomai Improved a bit on this after our last talk. This is how it looks now for commands:

Screenshot from 2021-08-20 13-20-54
Screenshot from 2021-08-20 13-21-33

catalintomai (approved):

utils/errors.go
```diff
@@ -32,3 +35,27 @@ func NewForbiddenError(args ...interface{}) error     { return NewError(ErrForbi
 func NewInvalidError(args ...interface{}) error       { return NewError(ErrInvalid, args...) }
 func NewNotFoundError(args ...interface{}) error      { return NewError(ErrNotFound, args...) }
 func NewUnauthorizedError(args ...interface{}) error  { return NewError(ErrUnauthorized, args...) }
+
+type LocError []*i18n.LocalizeConfig
```
hanzei, line 39: It would be great to have a godoc comment here explaining what this does, how to use it and why it doesn't implement the `Error` interface

server/command/service.go
```diff
@@ -261,60 +261,65 @@ func (s *service) ExecuteCommand(pluginContext *plugin.Context, commandArgs *mod
...
-		return s.errorOut(params, errors.New(s.conf.I18N().LocalizeDefaultMessage(loc, &nsi18n.Message{
-			ID:    "apps.command.error.noSubCommand",
-			Other: "no subcommand specified, nothing to do",
-		})))
+		return s.errorOut(params, utils.NewLocError(&nsi18n.LocalizeConfig{
+			DefaultMessage: &nsi18n.Message{
+				ID:    "apps.command.error.noSubCommand",
+				Other: "no subcommand specified, nothing to do",
+			},
+		}), errors.New("no subcommand specified, nothing to do"))
```
hanzei, line 308: Should this even return a `error`  given that the output is specifically targeting the user? I.e. should errors that "overlap" be reported as two different one or should the generic be an extension of the localized?

hanzei (approved):
Great work :+1: 

I've leaved some questions on the design below, but they are not blocking. The direction the PR is pushing for is good :+1:

server/command/service.go
larkox: We have the option in the ErrorOut for the command. And the error is not used either way. The only reason I saw to keep it here was more semantic. If we were to pass nil in this error, we may consider there is no error. If we consider an error happening every time either locError or error is != nil, we would need to update all the checks (which may be none in this particular case).

hanzei: I couldn't find a place where the locError wrapping is used so I don't think any checks need to be changed right now. `ErrorOut` already checks for both errors when outputting them. 1/5 that not overlapping the errors would also make it simpler for developers to think about them and the code would be cleaner as the information is only in one place.

For example 

```go
	m, err := apps.ManifestFromJSON(data)
	if err != nil {
		return errorOut(params, errors.Wrap(err, "unable to decode "+manifestURL))
		return s.errorOut(params, utils.NewLocError(&i18n.LocalizeConfig{
			DefaultMessage: &i18n.Message{
				ID:    "apps.command.install.http.error.decodeManifest",
				Other: "unable to decode {{.ManifestURL}}",
			},
			TemplateData: map[string]interface{}{
				"ManifestURL": manifestURL,
			},
		}), errors.Wrap(err, "unable to decode "+manifestURL))
	}
```

would be

```go
	m, err := apps.ManifestFromJSON(data)
	if err != nil {
		return errorOut(params, errors.Wrap(err, "unable to decode "+manifestURL))
		return s.errorOut(params, utils.NewLocError(&i18n.LocalizeConfig{
			DefaultMessage: &i18n.Message{
				ID:    "apps.command.install.http.error.decodeManifest",
				Other: "unable to decode {{.ManifestURL}}",
			},
			TemplateData: map[string]interface{}{
				"ManifestURL": manifestURL,
			},
		}), err
	}
```

as the locError contains all the error that was generated here.

larkox: I agree with you 90%. The only caveat that comes to mind is one of the concerns you were having about the localization of errors, and that is the "searchability" of those. I will remove all the redundant information, but please take another look in case you want to put back anything into the "not localized string".

larkox:
I added another minor change that is removing the error return. It was not being used, and was prone to future mistakes thinking it will be used. Now it is more clear that you must return a command response, whether it is an error or not.

hanzei (approved):
Let's start using this infrastructure!

catalintomai (approved):

hanzei:
@larkox Heads up that I merged `master` into your PR. Hope you are fine with that.

larkox:
Blocking this PR since https://github.com/mattermost/mattermost-plugin-apps/pull/251 will remove the whole command package. We may use this to dogfood our own localization system for apps.

levb:
@larkox Oops... we need to seriously rebase this

larkox:
@hanzei Regarding the install modal, I can take a look. My objective was to keep things to the commands, but seems a quick win localizing that too.

Regarding errors, that would mean go down to other packages, and get back to the discussion on whether we should show the english error, the localized error, both, or what. That discussion was more or less sorted on the previous version of this ticket, but I could foresee that discussion being reopened. Because of that, I prefer to keep it so far like this, and keep on translating on the next iteration (maybe with help of the community?).

PR #220, opened 2021-07-26
[MM-26812] Add bot-mention and bot leave/join subscriptions

mickmister:
#### Summary

This PR adds subscriptions for:

- `bot_mentioned`
- `bot_joined_channel`
- `bot_left_channel`
- `bot_joined_team`
- `bot_left_team`

For the `bot_mention` subscription, bot channel membership is checked before sending the notification.

Also added tests for the subscriptions in general

#### Ticket Link

Fixes https://mattermost.atlassian.net/browse/MM-36812
Fixes https://mattermost.atlassian.net/browse/MM-36813


server/proxy/notify.go
```diff
@@ -0,0 +1,193 @@
...
+func (p *Proxy) Notify(cc *apps.Context, subj apps.Subject) error {
+	subs, err := p.store.Subscription.Get(subj, cc.TeamID, cc.ChannelID)
+	if err != nil {
+		return err
+	}
+
+	return p.notify(cc, subs)
+}
+
+func (p *Proxy) NotifyRemoteWebhook(app *apps.App, data []byte, webhookPath string) error {
```
mickmister, line 24: This function is unchanged, and is only part of this diff because it moved files.

server/proxy/proxy.go
```diff
@@ -190,6 +195,101 @@ func (p *Proxy) NotifyRemoteWebhook(app *apps.App, data []byte, webhookPath stri
 	return upstream.Notify(up, creq)
 }
 
+func (p *Proxy) NotifyMessageHasBeenPosted(post *model.Post, cc *apps.Context) error {
```
larkox, line 198: I would say this function name is misleading. We are not notifying whether a message has been posted. We are notifying a bot mention.

server/proxy/proxy.go
```diff
@@ -116,6 +117,10 @@ func (p *Proxy) Notify(cc *apps.Context, subj apps.Subject) error {
 		return err
 	}
 
+	return p.notify(cc, subs)
+}
+
+func (p *Proxy) notify(cc *apps.Context, subs []*apps.Subscription) error {
 	expander := p.newExpander(cc, p.mm, p.conf, p.store, "")
 
 	notify := func(sub *apps.Subscription) error {
```
larkox, line 126: This may be considered as shadowing the function. Does it make sense to use a different name?

larkox (approved):
LGTM, just two small nits.

server/proxy/proxy.go
levb: I think the intention is for this function to handle the other notification as well, eventually?

server/proxy/proxy.go
```diff
@@ -190,6 +195,101 @@ func (p *Proxy) NotifyRemoteWebhook(app *apps.App, data []byte, webhookPath stri
...
+
+		if app.BotUserID == cc.UserID {
+			subs = append(subs, sub)
+		}
+	}
+
+	return p.notify(cc, subs)
+}
+
+func (p *Proxy) canReadChannel(userID, channelID string) bool {
```
levb, line 289: 0/5 this function is useless, `p.mm.User.HasPermissionToChannel(userID, channelID, model.PERMISSION_READ_CHANNEL)` is perfectly intention-revealing as is

levb (approved):

server/proxy/proxy.go
mickmister: @larkox @levb The function actually handles both the standard `created_post` subscription and `bot_mention` subscription. The name is meant to be a one-to-one match with the plugin hook, but prefixed with `Notify` to show that it is not the hook itself.

mickmister:
I've added a `GET subscriptions` route to be able to examine the saved subscriptions. Need to follow up with https://mattermost.atlassian.net/browse/MM-37447 `Apps subscriptions should only be createable using the App's bot token`

levb:
I removed myself in favor of @hanzei

apps/subscription.go
```diff
@@ -24,13 +24,25 @@ const (
 	SubjectUserJoinedChannel Subject = "user_joined_channel"
 	SubjectUserLeftChannel   Subject = "user_left_channel"
 
+	// SubjectBotJoinedChannel and SubjectBotLeftChannel subscribes to
+	// SubjectUserJoinedChannel and SubjectUserLeftChanne plugin events,
+	// specifically when the App's bot is added or removed from the team
```
jfrerich, line 29:
```suggestion
	// specifically when the App's bot is added or removed from the channel
```

apps/subscription.go
```diff
@@ -24,13 +24,25 @@ const (
 	SubjectUserJoinedChannel Subject = "user_joined_channel"
 	SubjectUserLeftChannel   Subject = "user_left_channel"
 
+	// SubjectBotJoinedChannel and SubjectBotLeftChannel subscribes to
+	// SubjectUserJoinedChannel and SubjectUserLeftChanne plugin events,
```
jfrerich, line 28:
```suggestion
	// SubjectUserJoinedChannel and SubjectUserLeftChannel plugin events,
```

jfrerich (approved):
Nice work, @mickmister.  I did a quick scan sense you already have 2 passing dev reviews and only found some non-code related nitpicks.

mickmister:
@DHaussermann Please see testing notes below and leave feedback. Thank you 🙂 

Scope

Bot mention notification is sent to app when:
- the app is subscribed to the `bot_mention` subscription
- the bot is mentioned in a channel the bot is a member of

-----

Testing notes

- When app has subscribed to the bot mention event:
  - app will receive call when:
    - bot must be a member of channel
    - bot is mentioned in post
    - app only receives one call when the bot is mentioned more than once
    - when the bot and another app's bot is mentioned

  - app will not receive call when:
    - the bot is not mentioned in post
    - when another bot is mentioned, but this bot is not mentioned
- When app has not subscribed to the bot mention event:
  - app should never receive a call in any scenarios above

-----

Follow-up concerns

- Authentication for subscriptions API is currently just that the token being used must be from a sysadmin. This will change once we have OAuth scopes for the bots.
- Bot must be a member of the channel to receive the bot mention. This may not be obvious to users.

-----

Screenshots


image

image

mickmister:
`/test subscriptions bot_mention subscribe` and `/test subscriptions bot_mention unsubscribe` commands added to the test app here: https://github.com/larkox/mattermost-app-test/pull/7

mickmister:
@DHaussermann I've added some commands to test the join/leave functionality here https://github.com/larkox/mattermost-app-test/pull/8

DHaussermann (approved):

DHaussermann:
Tested and passed
- All subscription events can be delivered
- All subscription events can be successfully enabled and disabled
- No issues found

LGTM!

hanzei:
@mickmister Is there a docs PR for this change?

server/httpin/restapi/restapi.go
```diff
@@ -36,6 +36,7 @@ func Init(router *mux.Router, conf config.Service, proxy proxy.Service, appServi
 		httputils.CheckAuthorized(mm, a.handleCall)).Methods("POST")
 
 	subrouter.HandleFunc(mmclient.PathSubscribe, a.handleSubscribe).Methods("POST")
+	subrouter.HandleFunc(mmclient.PathSubscribe, a.handleGetSubscriptions).Methods("GET")
```
hanzei, line 39: Is there a corresponding method in `ClientPP`?

server/proxy/proxy.go
```diff
@@ -193,6 +200,100 @@ func (p *Proxy) NotifyRemoteWebhook(app *apps.App, data []byte, webhookPath stri
...
+		appsMap := p.store.App.AsMap()
+		mentionSubs, err := p.store.Subscription.Get(apps.SubjectBotMentioned, cc.TeamID, cc.ChannelID)
+		if err != nil && err != utils.ErrNotFound {
+			return errors.Wrap(err, "failed to get bot_mentioned subscriptions")
+		}
+
+		for _, sub := range mentionSubs {
+			app := appsMap[sub.AppID]
+			if app == nil {
+				continue
```
hanzei, line 224: Should we log an error here? I was expecting subscriptions to be cleaned when an app is uninstalled and hence this case should not happen, right?

mickmister: Indeed the subscriptions are not being cleaned up when an app is uninstalled. Created a ticket for this here https://mattermost.atlassian.net/browse/MM-37930

https://github.com/mattermost/mattermost-plugin-apps/blob/dfcace445d39fa39a06bd11d8f959cacadb589f0/server/proxy/uninstall.go#L15

server/httpin/restapi/restapi.go
mickmister: @hanzei This was mainly for debugging, but it should be put there you're right. I'll follow up with this change.

FYI we will likely be reworking how subscribing works, see https://community-daily.mattermost.com/core/pl/wu83zyfgdidgxbsdc9cftpyuto and https://mattermost.atlassian.net/browse/MM-37929

mickmister:
> @mickmister Is there a docs PR for this change?

@hanzei I'm hesitant to make docs for it because the way the App subscribes may change, but we don't have any docs on subscriptions so I'll make a docs page. Existing ticket for this here https://mattermost.atlassian.net/browse/MM-36970

hanzei:
@mickmister Got it. Just wanted to check that we have a ticket.

PR #227, opened 2021-08-03
Add localisation for Apps

larkox:
#### Summary
Send the locale of the acting user with all calls so the App can perform localisations.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-32335


utils/utils.go
```diff
@@ -106,3 +106,19 @@ func LastN(s string, n int) string {
 	}
 	return string(out)
 }
+
+func GetLocale(mm *pluginapi.Client, userID string) string {
+	if u, err := mm.User.Get(userID); err == nil {
+		return u.Locale
+	}
+
+	if locale := mm.Configuration.GetConfig().LocalizationSettings.DefaultClientLocale; locale != nil {
```
levb, line 115: we already have a cached Mattermost config in `conf`, is this necessary?

levb (changes requested):

levb (approved):

larkox:
@hanzei Friendly reminder for review.

server/proxy/proxy.go
```diff
@@ -286,6 +288,7 @@ func (p *Proxy) NotifyRemoteWebhook(app *apps.App, data []byte, webhookPath stri
 		},
 		Context: p.conf.Get().SetContextDefaultsForApp(app.AppID, &apps.Context{
 			ActingUserID: app.BotUserID,
+			Locale:       utils.GetLocale(p.conf.MattermostAPI(), p.conf.MattermostConfig().Config(), app.BotUserID),
```
hanzei, line 291: What is the use case for a bot specific locale?

server/proxy/bindings.go
```diff
@@ -72,6 +73,7 @@ func (p *Proxy) GetBindingsForApp(sessionID, actingUserID string, cc *apps.Conte
 	appCC := *cc
 	appCC.AppID = appID
 	appCC.BotAccessToken = app.BotAccessToken
+	appCC.Locale = utils.GetLocale(p.conf.MattermostAPI(), p.conf.MattermostConfig().Config(), actingUserID)
```
hanzei, line 76: That is a costly call. Doing that on every bindings call might be a big burden. I'm wondering, if we could deffer getting the locale to the `ExpandedContext` which will contain the user anyway (if the app requested it)

hanzei (changes requested):
I'me exited for the i18n effort :tada: 

I'm not sure this change is worth the performance impact.

server/proxy/proxy.go
larkox: You are right, we should probably use the server locale instead of the bot.

server/proxy/bindings.go
larkox: If we don't expand the user, then we cannot localise the bindings? Or do you mean taking the local from the user if expanded, and if not, do it this way?

hanzei: I mean relying on the app to request an expanded binding, if it needs it. Most apps won't localize (for now) and hence won't slow down the bindings requests.

hanzei (approved):
LGTM

levb (approved):

larkox:
Docs PR: https://github.com/mattermost/mattermost-developer-documentation/pull/922

larkox:
@DHaussermann Heads up that we are merging this without QA Review.

PR #230, opened 2021-08-04
[MM-37324] Add telemetry

larkox:
#### Summary
Add telemetry to the Apps Framework. Right now only the proxy package has access to the telemetry, but it can be easily added to any other package that needs it. We are tracking:

- Every Install, with the AppID and AppType of the App installed
- Every Uninstall, also with the AppID and AppType
- Every call performed, with the AppID and the location (this will help us to know which bindings are more popular)

Feel free to suggest any other event worth tracking.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-37324


aaronrothschild:
Adding @enelson720 so this is on his radar. 

@larkox can we log when each user has authenticated with each app?  (or some indication of how many different users are using an app on a particular server - ie: "On this server 12% of users are using App X" and "Looking across all Apps, on average Y% of users are using apps". and "Looking across all servers, App X,Y,Z are most popular amongst users based on authenticated calls"

larkox:
@aaronrothschild It is possible to log each time a user finish the OAuth steps for App X if the App X is using the Mattermost OAuth abstraction (mainly useful for lambda).

It is also possible for the calls register the acting user ID, and therefore see how many people per server is using the bindings. Nevertheless, we would have to do some filtering (either on code or interpreting the analytics) to filter out the "get bindings" call, since that one will be done by all users.

aaronrothschild:
@larkox I'd like to track those actions so we can track installation vs authentication vs ongoing usage if possible.

server/telemetry/telemetry.go
```diff
@@ -0,0 +1,52 @@
...
+		"appType": appType,
+	})
+}
+
+func (t *Telemetry) TrackCall(appID string, location string) {
+	if t == nil {
+		return
+	}
+
+	_ = t.tracker.TrackEvent("uninstall", map[string]interface{}{
```
catalintomai, line 48: "uninstall"- > "call' ?

server/telemetry/telemetry.go
```diff
@@ -0,0 +1,52 @@
...
+		"appID":   appID,
+		"appType": appType,
+	})
+}
+
+func (t *Telemetry) TrackCall(appID string, location string) {
```
catalintomai, line 43: could this be too noisy - tracking every call?

larkox:
I have added a new event for OAuth complete, and instead of tracking every single call (including bindings calls) we will just be registering "submit calls".

levb (approved):

levb:
What's up with cypress tests?

server/proxy/oauth2.go
```diff
@@ -78,6 +78,7 @@ func (p *Proxy) CompleteRemoteOAuth2(sessionID, actingUserID string, appID apps.
 		return errors.Errorf("oauth2: unexpected response type from the app: %q", cresp.Type)
 	}
 
+	p.conf.Telemetry().TrackOAuthComplete(string(appID), actingUserID)
```
hanzei, line 81: Nit
```suggestion
	p.conf.Telemetry().TrackOAuthComplete(string(appID), actingUserID)
	
```

server/telemetry/telemetry.go
```diff
@@ -0,0 +1,62 @@
...
+	return &Telemetry{
+		tracker: tracker,
+	}
+}
+
+func (t *Telemetry) UpdateTracker(tracker mmtelemetry.Tracker) {
+	t.tracker = tracker
+}
+
+func (t *Telemetry) TrackInstall(appID string, appType string) {
```
hanzei, line 21: Would the signature be simpler with 
```suggestion
func (t *Telemetry) TrackInstall(appID, appType string) {
```
?

server/telemetry/telemetry.go
```diff
@@ -0,0 +1,62 @@
...
+		return
+	}
+
+	_ = t.tracker.TrackEvent("uninstall", map[string]interface{}{
+		"appID":   appID,
+		"appType": appType,
+	})
+}
+
+func (t *Telemetry) TrackCall(appID string, location string, actingUserID string) {
```
hanzei, line 43: Should the call type also get tracked? If we add additional types later, that might help.

hanzei (approved):
Good work :+1: Looking forward to seeing the metrics

catalintomai (approved):

hanzei (approved):

larkox:
@DHaussermann Gentle reminder to review this. Not sure if you want to test it, since it is just adding telemetry.

larkox:
@DHaussermann Friendly reminder on this.

hanzei:
@larkox Also here

larkox:
Added the environment variable to work with telemetry.

hanzei:
Discussed offline. QA testing will happen on Community. @larkox Please merge.

aaronrothschild:
FYI @enelson720

larkox:
@aaronrothschild Yes, it makes sense. Not 100% sure if we have the mechanism to make the distinction (I guess so, but not 100% sure). Feel free to create a ticket to track it.

PR #231, opened 2021-08-04
refactored common service dependencies into config

levb:
https://github.com/mattermost/mattermost-plugin-apps/compare/lev-servicecontext?expand=1#diff-5f44c54f32509964dfd9c962c81d4c62e93ba15bcfcf7eac8c481a84120bf17aL20 is the core change - moved common dependencies like `mm`, `log` into `config.Service` so we just pass `conf` around.


larkox (approved):
Love it. The only caveat is that the "conf" starts to feel not so much as a "configuration package" but much more. But not sure if it is worth looking for a different name.

levb:
I'm contemplating one more "fix" (before merging it, but likely a separate PR) making `config.Service` a *struct rather than an interface. The reasoning is simple: we use interfaces for mock-ability, but `Config` has a `TestService` that is arguably better suited for testing than full-on API-level mocks. 

RE:  _not so much as a "configuration package" but much more._  - see the branch name :)

mickmister (approved):

levb:
@DHaussermann not quite sure if there's anything specific to test for here, this pure refactoring

DHaussermann (approved):
Tested and passed - briefly deployed and regression tested.
- No issue deploy or enabling
- Tested install / uninstall flow
- Bindings are shown as expected
No issues found 

LGTM!

PR #232, opened 2021-08-06
Define the command trigger at manifest

larkox:
#### Summary
This defines the command trigger at manifest. There is also a breaking change, since we simplify the shape of commands. This also works as a stepping stone towards allowing aliases for commands. To illustrate the breaking change, this is how we used to create a command:
```
    pingCommand = Binding()
    pingCommand.location = "ping"
    pingCommand.label = "ping"
    pingCommand.call = Call()
    pingCommand.call.path = "/command/ping"
    pingCommand.form = Form()
    pingCommand.icon = "icon.png"
    
    baseCommandBinding = Binding()
    baseCommandBinding.label = "blogapp"
    baseCommandBinding.description = "Commands for Blog App"
    baseCommandBinding.location = "blogapp"
    baseCommandBinding.bindings = []
    baseCommandBinding.bindings.append(pingCommand)

    commandBinding = Binding()
    commandBinding.location = "/command"
    commandBinding.bindings = []
    commandBinding.bindings.append(baseCommandBinding)
```
Now it would be something like this:
```
    pingCommand = Binding()
    pingCommand.location = "ping"
    pingCommand.label = "ping"
    pingCommand.call = Call()
    pingCommand.call.path = "/command/ping"
    pingCommand.form = Form()
    pingCommand.icon = "icon.png"
    
    commandBinding = Binding()
    commandBinding.description = "Commands for Blog App"
    commandBinding.location = "/command"
    commandBinding.bindings = []
    commandBinding.bindings.append(pingCommand)
```

Old Apps will then have two levels at the base command. Instead of `/blogapp ping` they will have `/appID blogapp ping`.

I feel this change make things much more intuitive. Nevertheless, if you feel it is not worth it, we can revert that part of the code and keep the trigger related code as non-breaking.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-35530


larkox:
@levb @hanzei Friendly reminder to review this.

hanzei (commented):
Talked about this PR with @larkox offline.

While this PR is first step in allowing for command aliases it adds complexity and makes the developers experience more confusing as the root command is now modified before being send to the client.

While I not objected to the change, I would like to deffer it until there is a full spec that talk about command aliases as it's not clear to me if/when we need this feature.

levb:
@larkox @hanzei I feel similarly - defer, we are not ready to implement "shortcuts" yet.

larkox:
Closing until we decide to move in this direction. We can reopen it later.

PR #233, opened 2021-08-10
Add webapp plugin to trigger plugin enable/disable events

mickmister:
#### Summary

This PR adds a dummy webapp plugin to hook into the plugin status system hooked up with the MM webapp already. When the plugin is enabled or disabled, the frontend is aware and can avoid pointless requests accordingly.

#### Ticket Link

Fixes https://mattermost.atlassian.net/browse/MM-34233
Fixes https://mattermost.atlassian.net/browse/MM-34710

#### Related Pull Requests

Webapp PR that makes use of this https://github.com/mattermost/mattermost-webapp/pull/8545

Also needs to be implemented for mobile, which will require more code than the webapp since it's currently not using the plugin enable sync logic.


larkox (approved):
LGTM

mickmister:
I wonder if in theory we could get rid of most of the dependencies added for this. This could really be done in one file, so no need for es6/babel syntax. "If it's not broken don't fix it", but it would be cool to just ship one file from source.

levb (approved):

levb:
@mickmister what's up with CI?

hanzei (commented):
This seems like a hack to me. The server already sends a `WebsocketEventPluginEnabled` event to all clients when ever a plugin gets disabled. Can we make use of that and disable bindings when ever it's received?

hanzei:
Please note that my suggestion requires server changes as https://github.com/mattermost/mattermost-server/blob/757dc96461ffb27b0c8a5a1e30cd5fbabc871eef/app/plugin.go#L128-L133 is only triggered for webapp plugins.

mickmister:
@hanzei This is indeed a hack, though agreed upon in a meeting I think you may not have attended. There's potentially a problem with notifying all users (not just admins) about all plugins installed on the server. If there are any server-side plugins that the admin doesn't necessarily want non-admins to know they are installed, now all users can be aware of these plugins.

Thoughts on this issue @aaronrothschild @catalintomai @levb? For context we are adding a dummy webapp portion to the Apps plugin to hook into the existing system of notifying the client when a webapp plugin (in this case the Apps framework as a whole) has been enabled or disabled, as well as allowing the client to query active plugins and analyze the state of the Apps plugin.

hanzei:
@mickmister and I talked about this PR offline. The mm server already send websocket event for enabled plugin. Hence, clients already know the list of installed plugins. For disabled plugins the event is only send when there is a webapp part. @mickmister is going work on a server PR that changes that to send a event for all plugins.

mickmister:
Closing in favor of https://github.com/mattermost/mattermost-plugin-apps/pull/253

PR #235, opened 2021-08-13
[MM-36622] Remove http-hello and update e2e tests

mickmister:
#### Summary

This PR removes the helloapp from the repo, including references to the helloapp from the Apps plugin.

It also updates the e2e tests to use the `examples/go/hello-world` app instead. I've updated `.circleci/config.yml` to include a new step to start the `hello-world` app.

#### Ticket Link

Fixes https://mattermost.atlassian.net/browse/MM-36622

-----

Wondering if we should also close these Jira issues:

https://mattermost.atlassian.net/browse/MM-34386
MM Apps - http-hello threw a react error warning of memory leak 
^ Seems to be an issue with the `EmbeddedBindings` component on the frontend

~~https://mattermost.atlassian.net/browse/MM-34287~~
~~MM Apps - http-hello sends survey post in triple quotes~~
^ Closed since this was an implementation detail of the http-hello app

~~https://mattermost.atlassian.net/browse/MM-33996~~
~~Subscribe commands on http-hello return a 404 error~~
^ Closed since this was an implementation detail of the http-hello app


levb (approved):

.circleci/config.yml
```diff
@@ -152,6 +152,10 @@ jobs:
       - run:
           name: Install Apps plugin
           command: make deploy
+      - run:
+          name: Start hello-world App
+          command: make hello_world
+          background: true
```
hanzei, line 158: Why is `background` used here? What happens if the cypress tests are run before the `hello_world` app started?

hanzei (approved):
Nice cleanup :+1: 

One non blocking question below

.circleci/config.yml
mickmister: @hanzei If `background` is not used here, the step for starting the `hello_world` app never finishes, because it just listens and never exits. Using `background` may not be the right solution but it works. Any thoughts on how to properly accomplish this?

> What happens if the cypress tests are run before the `hello_world` app started?

The app install fails, and subsequently the e2e test fails

hanzei: Does `make hello_world&` do the trick?

mickmister: @hanzei I haven't tried yet but I think that might make the CI step exit since the foreground will run to completion. Or are you saying that starting the `hello_world` app should be within the step that runs the tests?

PR #236, opened 2021-08-16
Update ci to run hello app within cypress step instead of its own step

mickmister:
#### Summary

This PR is a followup to https://github.com/mattermost/mattermost-plugin-apps/pull/235#discussion_r689776274, which notes that the hello app does not need to be run as a separate step than the tests running.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -152,10 +152,6 @@ jobs:
       - run:
           name: Install Apps plugin
           command: make deploy
-      - run:
-          name: Start hello-world App
-          command: make hello_world
-          background: true
       - *save_go_cache
       - *restore_cypress_cache
       - run:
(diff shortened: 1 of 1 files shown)


mickmister:
@hanzei I'm not sure if this is better than using `background: true`, but I thought I'd throw this up here since you mentioned this in the other PR https://github.com/mattermost/mattermost-plugin-apps/pull/235#discussion_r689776274

hanzei (changes requested):
1/5 that the idiomatic way is to have a separate job for each dependency. The previous way looks fine to me.

mickmister:
> 1/5 that the idiomatic way is to have a separate job for each dependency. The previous way looks fine to me.

@hanzei So this PR should be closed?

hanzei:
@mickmister Yes, given that the second commit failed, I think we can close this

PR #240, opened 2021-08-19
Remove usage of non-existent c.SourePluginId

mickmister:
#### Summary

On master of mattermost-server, the `SourcePluginId` has been **removed**. Plugins should instead use the `Mattermost-Plugin-ID` header, as noted in the original deprecation comment:

```go
type Context struct {
	SessionId      string
	RequestId      string
	IpAddress      string
	AcceptLanguage string
	UserAgent      string
	SourcePluginId string // Deprecated: Use the "Mattermost-Plugin-ID" HTTP header instead
}
```

This PR changes the proxy plugin's `ServeHTTP` function to avoid using this field, since it's always blank. Plugins can successfully send requests to the proxy plugin with this change.


DHaussermann:
@mickmister could you please see about merging latest `master` up when you have a chance? 

I believe they'll be many changes and the testing would be more valuable if you're able to pull in Lev's latest changes that are now in master.

mickmister:
@DHaussermann I've updated the PR with master, but I'm now seeing this error when I run `make deploy`:

```
./build/bin/pluginctl deploy com.mattermost.apps dist/com.mattermost.apps-0.7.0.tar.gz
2021/08/19 15:29:24 Connecting using local mode over /var/tmp/mattermost_local.socket
2021/08/19 15:29:24 Uploading plugin via API.
Failed: failed to upload plugin bundle: AppErrorFromJson: model.utils.decode_json.app_error, body: http: request body too large,
```

See conversation here https://community-daily.mattermost.com/core/pl/f7riwciyzfn8jcdpo1xx5fn8jy

hanzei (approved):
Good catch. :+1: I've filled https://mattermost.atlassian.net/browse/MM-38047 to update the dependencies, which will catch other possible issues.

mickmister:
@DHaussermann FYI that the issue I've pointed out above is solved when you increase the max file upload size. We're discussing this issue in triage this week https://mattermost.atlassian.net/browse/MM-34215. With that in mind, this PR is ready for QA review

levb:
The `plugin` type support is still evolving, I am not sure we need a full QA pass for it yet. Will fix if we find issues before v0.8.0, this will get test coverage.

PR #241, opened 2021-08-23
[MM-38047] Use mattermost-server/v6

hanzei:
#### Summary
This PR updates the dependency on `github.com/mattermost/mattermost-server` from `v5` to `v6`.

Most changes are straight forward. On noticeable is the removal of `model.AppError` from `model.Client4`. Hence I've also removed it from `ClientPP`. Now a generic `error` is returned.

`model.PossibleAtMentions` not longer exists, so I copied the code over.

Please note that the plugin still works against v5 servers as the E2E test ensures.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-38047


server/proxy/proxy.go
```diff
@@ -225,7 +228,28 @@ func (p *Proxy) NotifyMessageHasBeenPosted(post *model.Post, cc *apps.Context) e
 
 	subs := []*apps.Subscription{}
 	subs = append(subs, postSubs...)
-	mentions := model.PossibleAtMentions(post.Message)
+
+	// This code is copied over from mattermost-server/app.possibleAtMentions
+	possibleAtMentions := func(message string) []string {
```
mickmister, line 233: Can we define this function somewhere else? I want to keep these `Notify*` functions as simple as possible

mickmister (changes requested):
Nice work @hanzei :+1:  I have one request to extract the `possibleAtMentions` function into its own flat function to simplify the function it's currently defined in

mickmister:
Any idea why this PR has such a smaller diff in `go.sum` than https://github.com/mattermost/mattermost-plugin-starter-template/pull/157/files?

server/proxy/proxy.go
hanzei: Sure

hanzei:
@mickmister I suspect it's because the version bump in the starter template is much higher (5.32 -> v6.0) then in this plugin.

hanzei:
Heads up that I've sneaked in a fix for https://mattermost.atlassian.net/browse/MM-38116 into this PR. 92887c7 contains it. The return type of `KVSet` is not `interface{}` but only a single `bool` that indicates if the value changed. We can return a proper `struct` in the future, if we want. The changes are in 92887c7e3a1e88165da7e2776cd62006c076038b.

mickmister (approved):
LGTM

hanzei:
@DHaussermann I don't think there is any function testing applicable here. I've smoke tested it myself. Is there anything you want to look into? Otherwise I would just merge it.

hanzei:
@DHaussermann Heads up that I'm merging this PR without further QA testing.

PR #245, opened 2021-08-25
[MM-38099] Limit reads in API and command layer

hanzei:
#### Summary
When reading a http response body the reader should be limited to avoid running out of memory when parsing a very large request. This PR limits all the reads happening at the API layer

Most notable the KV put methods is limited by the max length of the KV store values. Please note that the apps plugin simply cuts of the spare bytes. The clients need to handle this case.

The other notable restriction is on the manifest size. I now can be up to 1  MB, which is much bigger then e.g. the MM server default config file.

Given that this is a security feature, I've also added unit tests for the changes REST API endpoints.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-38099


levb (approved):

hanzei:
I'm going to merge this PR as it's covered by tests and I've smoke tested the hello app.

PR #246, opened 2021-08-25
[MM-38153] Allow test-e2e to run all tests

hanzei:
#### Summary
This PR allows running all test using `make test-e2e` by removing the build tags for the non e2e tests. It also allows for easier developing as a developer can now set the local build tags to `e2e` and compile all files.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-38153

### DIFF
--- a/apps/call_test.go
+++ b/apps/call_test.go
@@ -1,5 +1,3 @@
-// +build !e2e
-
 package apps
 
 import (
--- a/server/httpin/restapi/kv_test.go
+++ b/server/httpin/restapi/kv_test.go
@@ -1,5 +1,3 @@
-// +build !e2e
-
 package restapi
 
 import (
--- a/server/proxy/bindings_test.go
+++ b/server/proxy/bindings_test.go
@@ -1,5 +1,3 @@
-// +build !e2e
-
 package proxy
 
 import (
--- a/server/store/service_test.go
+++ b/server/store/service_test.go
@@ -1,5 +1,3 @@
-// +build !e2e
-
 package store
 
 import (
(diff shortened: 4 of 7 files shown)


catalintomai:
@hanzei - wrt to "running all test using make test-e2e" - shouldn't we change the naming since not all test (including unit, etc. ?) are considered e2e now?

hanzei:
@catalintomai The target `make test` is currently an alias to `make test-unit`, which I think it a good things as most of the time a developers wants to run the non-e2e test as the e2e ones take a long time.
While renaming `make test-e2e` makes sense, I'm struggling to find a better name as `make test` is already taken. Maybe `make test-all`. Do you have suggestions?

catalintomai:
@hanzei, wrt to "as most of the time a developers wants to run the non-e2e test as the e2e ones take a long time" - it looks like we expect a differentiation between e2e and non-e2e, so putting all tests under the e2e umbrella seems to go contrary to that.  On the naming, as a developer I'd be a bit confused on the difference between `make test` and `make test-all` - I'd rather that we don't have `make test` to begin with - then we can have only smth like `make test-all`. I'd say that we follow the taxonomy that we use in other repos/parts of the product. Do they also encompass all tests under the `make test-e2e` moniker?

hanzei:
> As I mentioned before, imo we should be consistent to what we do in other places/repos -

In the mattermost server repo `make test` runs _some_ of the tests, while leaving specific and long running tests to specific commands. Please note that similar to how it's done here the `make test` calls sub commands like `make test-server`. Are there other repos that you are referring to? 


FWIW this PR doesn't change the semantic. Currently `make test-e2` runs _some_ of the unit tests, but without consistency which are run and which not. With this change it's runs all of them and is consistent. Maybe we can defer the make target discussion to another ticket?

catalintomai (approved):

PR #247, opened 2021-08-25
[MM-33770] Use permssion specific client to expands calls

hanzei:
#### Summary
Before this PR apps could expand call to access resources they would not have access otherwise. This PR prevents that by using a permission specific client to expand the calls.

For apps that have the `PermissionActAsAdmin` permission the RPC client is used. If `PermissionActAsUser` is granted, an HTTP client with the users session is used. For `PermissionActAsBot` an HTTP client with the bot session is used.

I've opted for this approach because the apps plugin doesn't need to re-implement the permission checks and just relies on the server to do the right checks.

I tried adding an (go) e2e test for this, but that turned out to be more complicated that I thought. If I get it to work, I will add it later.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-33770

### DIFF
--- a/server/command/service.go
+++ b/server/command/service.go
@@ -358,7 +358,7 @@ func (s *service) newCommandContext(commandArgs *model.CommandArgs) *apps.Contex
 }
 
 func (s *service) newMMClient(commandArgs *model.CommandArgs) (mmclient.Client, error) {
-	return mmclient.NewHTTPClient(s.conf, commandArgs.Session.Id, commandArgs.UserId)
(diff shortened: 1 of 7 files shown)


levb (approved):

hanzei:
@DHaussermann I'm not sure how you can test this ticket without writing an app to test it. Both production apps are not effected by this.

DHaussermann:
There is no functional testing currently applicable for this. Please merge and I can do functional testing later when there is an app that leverages this PR.

PR #248, opened 2021-08-26
[MM-38155] Add recover handler for commands and HTTP requests

hanzei:
#### Summary
The plugin has two main entry points for external request: commands and HTTP requests. A panic in either the command or HTTP handler does shut down the whole plugin. While the plugin health check helps with mitigating the impact, a malicious attacker could continue to crash the plugin until the health check doesn't kick in any longer.

This PR fixes the attack vector by introducing a recovery handler for both command and HTTP requests. Debug information are logged. When the server is running in developer mode it also outputs the debug information in the command response and the HTTP response respectively. That should help developers with debugging without having to switch context.

#### Screenshots
Screenshot from 2021-08-26 13-38-05
Screenshot from 2021-08-26 13-38-40

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-38155

### DIFF
--- a/server/command/service.go
+++ b/server/command/service.go
@@ -2,6 +2,7 @@ package command
 
 import (
 	"fmt"
+	"runtime/debug"
 	"sort"
 	"strings"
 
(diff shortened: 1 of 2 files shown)


levb (approved):

catalintomai (approved):

DHaussermann (approved):
There is no functional testing applicable here. 

@hanzei Please merge.

PR #249, opened 2021-08-27
[MM-38190] [MM-38015] Don't try to set the bot icon in install process if request to it failed

hanzei:
#### Summary
This PR fixes two bugs:
- In 73c2f8e1b7b1436ce57e4e717dd7148c55620d3e a check is added that confirms that the bot icon resource returned a 200. Otherwise if e.g. the file is not found the install process fails much later as the body is blindly passed over to the MM server.
- In d3e373c886b122439a4b89b48890f5c34ff4120f fetching the app icon is moved at the start of the install process allowing it to fail early if the icon can't be fetched.

#### Ticket Link
- https://mattermost.atlassian.net/browse/MM-38190
- https://mattermost.atlassian.net/browse/MM-38015

### DIFF
--- a/server/proxy/install.go
+++ b/server/proxy/install.go
@@ -5,6 +5,7 @@ package proxy
 
 import (
 	"fmt"
+	"io"
 	"net/http"
 	"strings"
 
@@ -53,7 +54,17 @@ func (p *Proxy) InstallApp(client mmclient.Client, sessionID string, cc *apps.Co
 		app.WebhookSecret = model.NewId()
 	}
 
-	err = p.ensureBot(client, app)
+	var icon io.ReadCloser
+	if app.Manifest.Icon == "" {
+		icon, err = p.getAppIcon(app)
+		if err != nil {
+			return nil, "", errors.Wrap(err, "failed get bot icon")
+		}
+
+		defer icon.Close()
+	}
+
+	err = p.ensureBot(client, app, icon)
 	if err != nil {
 		return nil, "", err
 	}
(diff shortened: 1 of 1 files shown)


hanzei:
@mickmister Thanks for the review :+1:. I've restructured the make it more readable and also fix the two bugs that you found. PTAL.

larkox (approved):

levb (approved):

mickmister (approved):
LGTM :+1:

DHaussermann:
@hanzei  I don't believe this is working as expected. The installation still fails when the request to the icon returns a 404. 
This may be an improvement as **I think** previously this would fail silently which was far worse. But, should the install not succeed? 

Steps Followed:
- Compile the apps plugin from this PR and deploy to local server
- Get latest `master` of Zendesk
- Modify the manifest to make the icon unreachable `"icon": "zendesk-profile-not-icon.png",`
- Start the app as a local http and start the watch  
- Expose the app via ngrok
- Install the app as `http` by pasing in the URL to the modified manifest
**Observed**:
Breaking the icon worked
image
The installation still fails (No DM from the bot or bindings)
image

hanzei:
@DHaussermann The code works as I intentioned it. Previously, there was a crytic error and the installation would fails with an new bot account which doesn't have an icon. Now the installation still fails, but with a clear error message and no broken bot account.

Does that UX could good to you?

DHaussermann (approved):
Tested  and passed

Thanks @hanzei  I just wasn't sure if the goal was still to complete this install. This is much better as there is now user facing feedback confirming the install has failed and why.
Yes the UX on this is fine and the message is specific enough.

LGTM!

PR #251, opened 2021-08-30
OpenFaaS part 2: convert command and dialog to builtin

levb:
- Mostly parity with previous command/dialog
- `/apps install` now has 3 contextual subcommands: `marketplace`, `s3`, `url`
- added S3 List methods to support autocomplete for `/apps install s3`


larkox:
Waiting for #250 to merge before reviewing.

levb:
https://github.com/levb/mattermost-plugin-apps/pull/41 has the preview of the incremental diffs

mickmister:
> [levb#41](https://github.com/levb/mattermost-plugin-apps/pull/41) has the preview of the incremental diffs

Any reason not to just point this PR to the other PR's branch?

hanzei (changes requested):
See https://github.com/levb/mattermost-plugin-apps/pull/41#pullrequestreview-744890566

levb:
@hanzei @larkox @mickmister this is ready for review, I incorporated (almost) all feedback from https://github.com/levb/mattermost-plugin-apps/pull/41

server/builtin/debug.go
```diff
@@ -0,0 +1,53 @@
...
+					Path: pDebugBindings,
+				},
+				Form: &noParameters,
+			},
+		},
+	}
+}
+
+func (a *builtinApp) debugClean(creq apps.CallRequest) apps.CallResponse {
+	_ = a.conf.MattermostAPI().KV.DeleteAll()
```
larkox, line 42: Maybe I am wrong, but seems like the only "sysadmin" check is when providing the bindings. That means that I could just curl the endpoint with any user token and clean the database. Am I right? Should we add some more checks here? (or in a middleware in the handling level)

larkox: Similar concerns with other endpoints.

larkox (changes requested):
Thank you for the changes. There is still one pending concern in terms of security, but I am not sure if it is already solved by something I may have overlooked.

server/builtin/debug.go
levb: Argh, I think you're right. I will fix it, of course, but really... it bothers me. I wanted a shortcut - using direct `go` interfaces from `builtin` but I question it now. Perhaps, the simplest/best course of action is to have `builtin` be a true app in the sense of needing to expand the admin token, and then using `appsclient` to invoke the necessary APIs.

levb: Done, somewhat idiomatically for apps, with the help of Expand.

server/builtin/app.go
```diff
@@ -0,0 +1,273 @@
...
+}
+
+const SysadminOnly = true
+
+func (a *builtinApp) handle(
+	path string,
+	sysadminOnly bool,
+	submitf func(apps.CallRequest) apps.CallResponse,
+) {
+	a.router[submitPath(path)] = requireSysadmin(sysadminOnly, submitf)
```
larkox, line 215: 0/5 seems hard to read that you pass a function "requireSysadmin" to all endpoints, instead of only to the functions that require it. I see that the current code works correctly, but seems confusing at this point.

server/builtin/app.go
```diff
@@ -0,0 +1,273 @@
...
+			return formResponse(*form)
+		})
+}
+
+func requireSysadmin(require bool, handler func(apps.CallRequest) apps.CallResponse) func(apps.CallRequest) apps.CallResponse {
+	if !require {
+		return handler
+	}
+	return func(creq apps.CallRequest) apps.CallResponse {
+		if creq.Context.ExpandedContext.AdminAccessToken == "" {
```
larkox, line 268: 0/5. Do you feel it is better to expand the token instead of expanding the user and checking its role? The token seems more "dangerous" than the user object.

larkox (approved):
In general, LGTM. I added two comments to discuss, but I am ok with the PR being merged as it is.

server/builtin/app.go
levb: 0/5 the token's presence is a single bit, it can be explicitly expressed with the call definition... expanding the user is "fatter" (more data), and a little opaque (all Calls are passed to the builtin, rather than having the unauthorized rejected by the proxy itself, based on the `Expand` declaration... And we'd still need an `Expand`... so it's not more compact.

server/builtin/app.go
levb: @larkox this resulted in a bigger change than expected, but should be cleaner now.

mickmister (approved):
LGTM  🚀  Just a few questions

server/builtin/install_marketplace.go
```diff
@@ -0,0 +1,48 @@
...
+				Location:    "marketplace",
+				Hint:        "[app ID]",
+				Description: "Installs an App from the Marketplace",
+				Call:        &installMarketplaceCall,
+				Form:        appIDForm(installMarketplaceCall),
+			}
+		},
+
+		lookupf: func(creq apps.CallRequest) ([]apps.SelectOption, error) {
+			return a.lookupAppID(creq, func(app apps.ListedApp) bool {
```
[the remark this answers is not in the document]
levb: Out of scope for the PR, I'm ok with it where it is. cc @hanzei. 0/5 I'd wait until `apps` further matures.

server/builtin/install_consent.go
```diff
@@ -0,0 +1,123 @@
...
+	// Consent
+	consent := ""
+	if len(m.RequestedLocations) > 0 {
+		consent += "\n- Add the following elements to the **Mattermost User Interface**:\n"
+		for _, l := range m.RequestedLocations {
+			consent += fmt.Sprintf("  - %s\n", l.Markdown())
+		}
+	}
+	if len(m.RequestedPermissions) > 0 {
+		consent += "- Access **Mattermost API** with the following permissions:\n"
```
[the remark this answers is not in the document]
levb: It's a nit, fixed in the next PR

hanzei (changes requested):
There seems to be something fishy with the autocomplete options. The three (dynamic) autocomplete calls (enable, disable, uninstall) always return output, that no suitable app was found. Could you please look into this?

Secondly: I'm not sure if this a known issue: The image in the install dialog is broken. I know we talked about the difficulties on fetching the image from the correct upstream. Do you plan to fix the broken image? Or should it just be removed for now?

levb:
> There seems to be something fishy with the autocomplete options. The three (dynamic) autocomplete calls (enable, disable, uninstall) always return output, that no suitable app was found. Could you please look into this?

@hanzei Done, was fixed in part3, was not using the right data type.
 
> Secondly: I'm not sure if this a known issue: The image in the install dialog is broken. I know we talked about the difficulties on fetching the image from the correct upstream. Do you plan to fix the broken image? Or should it just be removed for now?

It is removed in part3/4, 1/5 we just leave as is here.

hanzei (approved):
LGTM :+1:

DHaussermann:
I have done some testing to ensure correct installation methods are exposed based on Cloud Vs On-Prem mode

1. I see an issue when trying to install hello-lambda after it has been provisioned. 
image

2. The Cloud mode true/false seems to have been dropped form what is shown when running `/apps info`

Testing of this is ongoing.

DHaussermann (approved):
Tested and passed
- Tested http and S3 installation methods for on-Prem
- Tested Markertplace install for cloud mode
- All apps tested where function
- Authentication was working successfully
- Uninstall and deactivate are working
- `/apps list` data is accurate 
- Mode value now visible in the UI again

LGTM!

PR #253, opened 2021-08-31
Broadcast websocket message when plugin is enabled/disabled

mickmister:
#### Summary

This PR makes it so the Apps plugin sends a websocket message to all users when it is activated and deactivated. The webapp and mobile app will receive this message and adjust its local state to keep track of the Apps plugin's presence.

Discussion on this approach - https://community-daily.mattermost.com/core/pl/wswoayjrxbbgznsj6ewxip8jie

It may be a good idea to send the version in the message, and have an API endpoint to fetch the version, so clients can fetch the version in their reconnect handlers.

#### Ticket Link

Fixes https://mattermost.atlassian.net/browse/MM-34233
Fixes https://mattermost.atlassian.net/browse/MM-34710

Related webapp PR that needs to be updated to use this websocket message https://github.com/mattermost/mattermost-webapp/pull/8545

#### Testing Notes

The `custom_com.mattermost.apps_plugin_enabled` and `custom_com.mattermost.apps_plugin_disabled` messages should show in the browser dev tools when the plugin is enabled or disabled:

image


larkox (approved):

hanzei (approved):
Thanks Michael for seeking this great solution :+1:

mickmister:
@hanzei What do you think about the idea of supplying the frontend with the current version of the Apps plugin? Then we can potentially make conditional logic on the frontends pertaining to the specific version. The sooner we add the version the better, so it is available when we need it. We would send the version in the message, and have an API endpoint to fetch the version, so clients can fetch the version in their reconnect handlers.

hanzei:
Sounds good to me @mickmister :+1:

hanzei:
@mickmister  Heads up that I merged `master` into your PR. Hope you are fine with that.

server/proxy/bindings.go
```diff
@@ -94,7 +94,7 @@ func (p *Proxy) getBindingsForApp(in Incoming, cc apps.Context, app apps.App) []
 	b, _ := json.Marshal(resp.Data)
 	err := json.Unmarshal(b, &bindings)
 	if err != nil {
-		log.Debugf("Bindings are not of the right type.")
+		log.WithError(err).Debugf("Bindings are not of the right type.")
```
hanzei, line 97: Good catch

PR #255, opened 2021-09-07
Don't require act_as_admin permssion to use RemoteOAuth2

hanzei:
#### Summary
Before this change, requesting the `remote_oauth2` permission would require also requesting `act_as_admin`. This unnecessary as most of the OAuth2 REST API endpoints as scoped by the user. (`/oauth2/user`)

The only app specific endpoint is `/oauth2/app`, which requires admin permission. The admin token can already be accessed by the app through `act_as_user` permission. The app only has to ensure that the `configure` bindings is only show to admins for a better UX. For non-admin it will just fail.

Please let me know if you have suggestion on the example.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-38243


larkox (approved):
Code LGTM.

Nevertheless, there are a few conceptual details that seem weird.

First of all, if we are acting as the user to do an admin task, we are conceptually acting as the admin (and therefore should inform the system).

Also, for the long term, the idea is to do this with an "app admin token" (which seems that will be a special bot token), so we will remove both act as user and act as admin from that part of the equation.

Also, the "app admin token" allows the app to handle its own permissions, like allowing users from a group to perform "app admin" stuff.

All that being said, LGTM for a short term solution, but we will change this again when we figure out the "app admin token".

hanzei:
> First of all, if we are acting as the user to do an admin task, we are conceptually acting as the admin (and therefore should inform the system).

Well, the app does what every the user credentials allow. That is part of the act_as_user permission.

> Also, the "app admin token" allows the app to handle its own permissions, like allowing users from a group to perform "app admin" stuff.

What makes the "app admin token" special in comparison to the other tokens? Why can't the other tokens be used to perform app wide actions?

larkox:
> Well, the app does what every the user credentials allow. That is part of the act_as_user permission.

In the current permission system, yes, you are right. In the desired permission/scope system, this is tricky. Acting as a user, implicitly lets you act as an admin. But acting as an admin is something that should be stated explicitly. Therefore my concerns around this change.

> What makes the "app admin token" special in comparison to the other tokens? Why can't the other tokens be used to perform app wide actions?

What if the app doesn't need to act as an user, or as an admin, just as itself. There might be some management task that will need a token, like the KV store (which will be handled by the bot token). I feel it is important that we separate what the app is doing in behalf of a different entity (user) or what it is doing for itself.

levb (approved):

levb (changes requested):
Asa matter of fact, I too am confused by this PR.

It seems that `configure` is something that only a syadmin should be able to perform, and I believe that the action of `StoreOAuth2Config` requires one. An admin token is not expanded to the app unless the App has act_as_admin permission. What am I missing?

A real OAuth2 App token that we will issue some day will not usually have sysadmin privs for a sysadmin used...

hanzei:
@levb 

> An admin token is not expanded to the app unless the App has act_as_admin permission

If an app has the `act_as_user` permission every call made from an admin will include a user access token for that admin account.

Surely, once proper OAuth scopes and session are implemented, this will not longer be the case.

levb (approved):

levb:
I am still not comfortable with the approach, but we'll need to revisit as we implement the scopes et all

hanzei:
@DHaussermann https://github.com/mattermost/mattermost-app-servicenow/pull/34 can be used for testing

levb:
@hanzei should this be merged to include in the `0.8.0` release testing?

hanzei:
I'm going to merge this PR as it should go into `v0.8.0`. If it broke something, it broke it loud.

hanzei:
cc @DHaussermann :point_up:

PR #256, opened 2021-09-07
[MM-38242] Directly use local http port to access Mattermost server from Apps plugin

hanzei:
#### Summary
This change greatly improves the performance as the apps plugin doesn't have to go through the reverse proxy any longer.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-38242


server/config/config.go
```diff
@@ -107,10 +109,25 @@ func (conf *Config) Reconfigure(stored StoredConfig, mmconf *model.Config, licen
...
+	}
+	host, port, err := net.SplitHostPort(*listenAddress)
+	if err != nil {
+		return err
+	}
+
+	if host == "" {
+		host = "127.0.0.1"
+	}
+	listenURL := "http://" + host + ":" + port
```
mickmister, line 124: Will we run into any certificate complications when the server is configured to forward `80 -> 443`? For example, the server's SSL cert may not match the `host` we are using here. Though I could be way wrong here.

levb: @mickmister What certificate? (`http`, not `https`)

@hanzei nit 3/5 s/listenURL/loopbackURL/, listen(Address-not-URL) should be ":8065", on the server.

server/mmclient/http.go
```diff
@@ -13,7 +13,7 @@ type httpClient struct {
 
 func NewHTTPClientFromSessionID(config config.Service, sessionID, actingUserID string) (Client, error) {
 	conf, mm, _ := config.Basic()
-	client, err := utils.ClientFromSession(mm, conf.MattermostSiteURL, sessionID, actingUserID)
+	client, err := utils.ClientFromSession(mm, conf.MattermostListenURL, sessionID, actingUserID)
```
levb, line 16: ditto s/Listen/Local[host]/g

levb (changes requested):
Just a naming nit, otherwise LGTM

server/config/config.go
hanzei: @mickmister Good question. The `Forward80To443` setting isn't an issue as it only adds an additional listener to port 80. The caveat is in `ConnectionSecurity`. If enabled, the MM server takes care of terminating the TLS connections. In that case there using the SiteURL is preferred as indeed the host name could mismatch.

server/mmclient/http.go
hanzei: Good point, updated

server/config/config.go
```diff
@@ -107,10 +109,33 @@ func (conf *Config) Reconfigure(stored StoredConfig, mmconf *model.Config, licen
...
+		host, port, err := net.SplitHostPort(*listenAddress)
+		if err != nil {
+			return err
+		}
+
+		if host == "" {
+			host = "127.0.0.1"
+		}
+
+		localURL = "http://" + host + ":" + port
```
mickmister, line 131: Is it possible that `port` is zero?

mickmister (approved):

catalintomai (approved):

levb (approved):

server/config/config.go
hanzei: Yes, that is possible, but I would count that as a bug of the server: https://github.com/mattermost/mattermost-server/blob/92837fa1ee69c4420ea0033a35b5058ca8afc76f/model/config.go#L3692-L3702

DHaussermann (approved):
Tested and passed
- Tested locally by installing test app and using various commands
- Tested on an HA server by installing Service Now and ensuing the the oAutrh config functionality is working which uses an expand
No issues found

LGTM!

PR #268, opened 2021-10-07
MM-39101: Support HTTP.UseJWT

levb:
#### Summary
- Added support for `"use_jwt"` in the `"http"` section of manifest.json.
- Added `hello-jwt` to document its use.
- Improved `InstallApp`'s check for whether the app is reachable, message if not. Note, unless the app provides a `/ping` end-point the check is not reliable.
- 

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-39101
https://mattermost.atlassian.net/browse/MM-39178

This is to match the docs PR for 0.8.0


levb:
@hanzei there was a scan error, 
image

I dismissed it since I did not really change the code, and presumably it had been dismissed before? Can you please confirm?

server/proxy/install.go
```diff
@@ -68,6 +68,20 @@ func (p *Proxy) InstallApp(in Incoming, cc apps.Context, appID apps.AppID, deplo
 		defer icon.Close()
 	}
 
+	// See if the app is inaaccessible. Call its ping path with nothing
+	// expanded, ignore 404 errors coming back and consider everything else a
+	// "success".
+	//
+	// Note that this check os often ineffective, but "the best we can do"
```
mickmister, line 75:
```suggestion
	// Note that this check is often ineffective, but "the best we can do"
```

server/proxy/install.go
```diff
@@ -93,30 +107,31 @@ func (p *Proxy) InstallApp(in Incoming, cc apps.Context, appID apps.AppID, deplo
...
-		// TODO fail on all errors except 404
-		if resp.Type == apps.CallResponseTypeError {
-			log.WithError(err).Warnf("OnInstall failed, installing app anyway.")
-		} else {
-			message = resp.Markdown
+		cresp := p.call(in, *app, *app.OnInstall, &cc)
+		if cresp.Type == apps.CallResponseTypeError {
+			// TODO: should fail and roll back.
+			log.WithError(cresp).Warnf("Installed %s, despite on_install failure.", app.AppID)
+			message = fmt.Sprintf("Installed %s despite on_install failure: %s", app.AppID, cresp.Error())
```
mickmister, line 116: To be consistent with the message above

```suggestion
			message = fmt.Sprintf("Installed %s, despite on_install failure: %s", app.AppID, cresp.Error())
```

server/proxy/install.go
```diff
@@ -93,30 +107,31 @@ func (p *Proxy) InstallApp(in Incoming, cc apps.Context, appID apps.AppID, deplo
...
-			Context: cc,
-		})
-		// TODO fail on all errors except 404
-		if resp.Type == apps.CallResponseTypeError {
-			log.WithError(err).Warnf("OnInstall failed, installing app anyway.")
-		} else {
-			message = resp.Markdown
+		cresp := p.call(in, *app, *app.OnInstall, &cc)
+		if cresp.Type == apps.CallResponseTypeError {
+			// TODO: should fail and roll back.
```
mickmister, line 114: Should this be ticketed or prioritized?

mickmister (approved):
LGTM :+1:

server/builtin/install_consent.go
```diff
@@ -132,7 +132,7 @@ func (a *builtinApp) newInstallConsentForm(m apps.Manifest, creq apps.CallReques
 	fields = append(fields, deployTypeField)
 
 	// JWT secret
-	if deployType == apps.DeployHTTP {
+	if deployType == apps.DeployHTTP && m.SupportsDeploy(apps.DeployHTTP) && m.HTTP.UseJWT {
```
larkox, line 135: If we have a AWS app that also support HTTP... will this show the JWT field even if we are trying to install it as a AWS app?

larkox (approved):
LGTM, just one small detail.

server/builtin/install_consent.go
levb: AWS Apps should not be checking JWTs, they should use the API Gateway for that if deployed as HTTP.

hanzei:
> @hanzei there was a scan error, image
> 
> I dismissed it since I did not really change the code, and presumably it had been dismissed before? Can you please confirm?

Yes, this one can be dismissed. It keeps popping up every time something in that files gets changed.

examples/go/hello-jwt/hello.go
```diff
@@ -0,0 +1,134 @@
...
+	return func(w http.ResponseWriter, req *http.Request) {
+		if _, err := checkJWT(req); err != nil {
+			http.Error(w, err.Error(), http.StatusUnauthorized)
+			return
+		}
+		f(w, req)
+	}
+}
+
+func checkJWT(req *http.Request) (*apps.JWTClaims, error) {
```
hanzei, line 116: Should the example demonstrate what claims are passed to the app and how to check them?

server/proxy/proxy.go
```diff
@@ -61,43 +61,77 @@ func (p *Proxy) Call(in Incoming, creq apps.CallRequest) CallResponse {
...
-	cresp := p.callApp(in, *app, creq)
+	cresp, _ := p.callApp(in, *app, creq)
 	return NewProxyCallResponse(cresp).WithMetadata(AppMetadataForClient{
 		BotUserID:   app.BotUserID,
 		BotUsername: app.BotUsername,
 	})
 }
 
-func (p *Proxy) callApp(in Incoming, app apps.App, creq apps.CallRequest) apps.CallResponse {
+func (p *Proxy) call(in Incoming, app apps.App, call apps.Call, cc *apps.Context, valuePairs ...interface{}) apps.CallResponse {
```
hanzei, line 71: 0/5 maybe it's time to move this to `server/proxy/call.go`

server/proxy/oauth2.go
```diff
@@ -59,11 +55,11 @@ func (p *Proxy) CompleteRemoteOAuth2(in Incoming, appID apps.AppID, urlValues ma
 		return err
 	}
 
-	cresp := p.callApp(in, *app,
-		apps.CallRequest{
-			Call:   app.OnOAuth2Complete.WithDefault(apps.DefaultOnOAuth2Complete),
-			Values: urlValues,
-		})
+	cresp, _ := p.callApp(in, *app, apps.CallRequest{
```
hanzei, line 58: Why doesn't this one use `p.call`?

upstream/upkubeless/upstream.go
```diff
@@ -106,15 +106,25 @@ func (u *Upstream) invoke(clientset kubernetes.Interface, url, method string, da
 	// So we need to manually build the URL
 	req = req.AbsPath(url)
 
-	received, err := req.Do().Raw()
-	if err != nil {
+	statusCode := 0
```
hanzei, line 109: Nit pick:
```suggestion
	var statusCode int
```

server/proxy/install.go
```diff
@@ -68,6 +68,20 @@ func (p *Proxy) InstallApp(in Incoming, cc apps.Context, appID apps.AppID, deplo
...
 	}
 
+	// See if the app is inaaccessible. Call its ping path with nothing
+	// expanded, ignore 404 errors coming back and consider everything else a
+	// "success".
+	//
+	// Note that this check is often ineffective, but "the best we can do"
+	// before we start the diffcult-to-revert install process.
+	_, err = p.callApp(in, *app, apps.CallRequest{
+		Call:    apps.DefaultPing,
```
hanzei, line 78: I'm wondering if the additional concept of a "ping path" is needed here. There are other known paths like `/bindings` that can be used. Any errors returned from there can be ignored. But maybe developers won't expect the path to be used in that way.

hanzei (approved):
LGTM :+1: A few non blocking comments

server/proxy/install.go
levb: I tried... the problem is, binding may have an expand and expect the data; the OAuth example app would crash if the `ActingUser` is not there... It does not seem proper to invoke /bindings without a full context, and the full context can not be properly expanded until the app is installed (sad!) so it's a chicken/egg situation.

server/proxy/proxy.go
levb: 👍  another PR.

PR #276, opened 2021-10-26
Separate calls

larkox:
#### Summary
Add separate calls for each function on Apps (submit, form source, form submit and lookup). Also updated the filtering with the new logic.

This PR also piggybacks the renaming of markdown and error to text, and filtering out "white space labels" (to avoid no text labels on post options and channel header).

Notes for testing: This is a breaking change. All Apps will need to be updated to work properly. The test app will be updated to fit the new approach.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-39092
https://mattermost.atlassian.net/browse/MM-39096
https://mattermost.atlassian.net/browse/MM-31121
https://mattermost.atlassian.net/browse/MM-33532

#### Related Pull Requests
Webapp: https://github.com/mattermost/mattermost-webapp/pull/9263
Mobile: TBD
Test App: TBD

### Breaking Changes
-  @larkox: TODO

The following apps need to get updated:
- [ ] Examples
- [ ] Test app
- [ ] ServiceNow
- [ ] Zendesk


apps/call.go
```diff
@@ -47,211 +26,101 @@ type Call struct {
...
+	if clone.Expand != nil {
+		cloneExpand := *clone.Expand
+		clone.Expand = &cloneExpand
 	}
 
-	opt, ok := c.Values[name].(map[string]interface{})
-	if ok {
-		if v, ok2 := opt["value"].(bool); ok2 {
-			return v
+	// Only know how to clone map values for State.
```
catalintomai, line 117: @larkox - this leaves open the possibility of a partial clone - should we call the method it PartialClone()then?

apps/field.go
```diff
@@ -87,9 +87,12 @@ type Field struct {
 	// autocomplete.
 	SelectRefresh bool `json:"refresh,omitempty"`
 
-	// SelectStaticOptions is the list of options to display in a static select
-	// field.
-	SelectStaticOptions []SelectOption `json:"options,omitempty"`
+	// SelectOptions is the list of options to display in a static select field.
+	SelectOptions []SelectOption `json:"options,omitempty"`
```
catalintomai, line 91: @larkox - why the renaming? 1/5 but I think it is easier to follow if we have `SelectStatisOptions` and `SelectLookupOptions` instead of `SelectOptions` and `SelectLookup`.

apps/form.go
```diff
@@ -15,17 +19,15 @@ package apps
...
-// When a select field with "refresh" set changes value, it forces reloading of
-// the form. A call request type form is made to fetch it, with the partial
-// values provided. Expected response is a "type":"form" response.
+// A form can be dynamically fetched if it specifies its Source. Source may
+// include Expand and State, allowing to create custom-fit forms for the
+// context.
 type Form struct {
+	// Source is the call to make when the form's definition is required (i.e.
+	// it has no fields, or needs to be refreshed from the app). A simple call
+	// can be specified as a path (string). It will contain no expand/state.
```
catalintomai, line 28: "It will contain no expand/state." - we mention above that "Source may include Expand and State" ?

apps/call_request.go
```diff
@@ -0,0 +1,124 @@
...
+	}
+
+	s, ok := creq.Values[name].(string)
+	if ok && s != "" {
+		return s
+	}
+
+	opt, ok := creq.Values[name].(map[string]interface{})
+	if ok {
+		if v, ok2 := opt["value"].(string); ok2 {
```
mickmister, line 99: If we have `GetValue` and `GetBool` functions, should we have one for every data type? I suppose we could make a `GetValues(name string) []string` to extract chosen values.

apps/form.go
```diff
@@ -53,3 +56,56 @@ type Form struct {
...
+	structValue := struct {
+		Source        *Call   `json:"source,omitempty"`
+		Title         string  `json:"title,omitempty"`
+		Header        string  `json:"header,omitempty"`
+		Footer        string  `json:"footer,omitempty"`
+		Icon          string  `json:"icon,omitempty"`
+		Submit        *Call   `json:"submit,omitempty"`
+		SubmitButtons string  `json:"submit_buttons,omitempty"`
+		Fields        []Field `json:"fields,omitempty"`
+	}{}
```
mickmister, line 82: Can we make a linting rule or something similar to make sure this stays in sync?

mickmister (approved):
Awesome job @larkox, this definitely improves the developer experience a lot.

From what I understand, the default is/should be that the developer puts a form into a binding. The less common case is when the form needs to be dynamically generated from heavy computations or needs a form refresh, then we have to use the `Source` call.

I have left some non-blocking questions, code LGTM :+1:

apps/call_response.go
```diff
@@ -0,0 +1,112 @@
...
+//
+// Lookup requests expect ok or error.
+//
+// In case of an error, the returned response type is "error", ErrorText
+// contains the overall error text. Data contains optional, field-level errors.
+type CallResponse struct {
+	Type CallResponseType `json:"type"`
+
+	// Text is used for OK and Error response, and will show the text in the
+	// proper output.
```
hanzei, line 54: I'm actually wondering if we should keep ` Error` as a field as all other call response types use there own field, why not OK and Error?

apps/call.go
```diff
@@ -47,211 +26,101 @@ type Call struct {
...
+	return nil
 }
 
-func NewCall(path string) Call {
+func NewCall(path string) *Call {
 	c := Call{
 		Path: path,
 	}
-	return c
+	return &c
```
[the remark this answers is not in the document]
larkox: NewCall is broadly used to assign directly to fields like SelectLookup. In that case, we cannot do `&NewCall("something")`. In order to simplify the code, we change this back to pointer.

apps/call_request.go
larkox: We can add them as we need them. I would say it is out of the scope of this PR.

apps/call_response.go
larkox: The main idea was to simplify the structures. @lev thoughts?

apps/form.go
larkox: That would be awesome, but not sure how to achieve it, and a bit low on mana to look into it. Are you OK if we create a ticket for it?

mickmister: @larkox Yeah this may be a good HW ticket too

mickmister: Maybe for now we put a comment near the `Form` struct definition saying to make sure to update this struct.

levb: done.

apps/call_response.go
levb: Well, `Error` was a `string` before and as such does not need a special field. If we decide to add data to it, we can do it as `Data`, as per convention.

levb:
Replaced by https://github.com/mattermost/mattermost-plugin-apps/pull/296, with all feedback done/resolved.

PR #281, opened 2021-11-01
MM-39776 moved Build variables to config

levb:
#### Summary
The build hashes were no longer displayed in `/apps info`. 2 causes were identified and fixed.
- package path seems to be different when building as a test and deploy, the latter does not work with `.../server`. Moved the build variables to `.../server/config` which is a FQ, accurate package name and it works in both cases
- https://stackoverflow.com/questions/44148449/how-to-check-whether-golang-binary-is-compiled-with-ldflags-s-w/44168916 - `custom.mk` was adding a separate `-ldflags`, overriding the prior values. Fixed by changing `LDFLAGS` instead of `GO_BUILD_FLAGS`

#### Ticket:
https://mattermost.atlassian.net/browse/MM-39776

### DIFF
--- a/server/builtin/app.go
+++ b/server/builtin/app.go
@@ -95,7 +95,7 @@ func NewBuiltinApp(conf config.Service, proxy proxy.Service, httpOut httpout.Ser
 func Manifest(conf config.Config) apps.Manifest {
 	return apps.Manifest{
 		AppID:       AppID,
-		Version:     apps.AppVersion(conf.BuildConfig.BuildHashShort),
+		Version:     apps.AppVersion(conf.PluginManifest.Version + "-" + conf.BuildHashShort),
 		DisplayName: AppDisplayName,
 		Description: AppDescription,
 		Bindings: &apps.Call{
--- a/server/builtin/app.go
+++ b/server/builtin/app.go
@@ -95,7 +95,7 @@ func NewBuiltinApp(conf config.Service, proxy proxy.Service, httpOut httpout.Ser
 func Manifest(conf config.Config) apps.Manifest {
 	return apps.Manifest{
 		AppID:       AppID,
-		Version:     apps.AppVersion(conf.PluginManifest.Version + "-" + conf.BuildHashShort),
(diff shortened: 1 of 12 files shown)


server/builtin/app.go
```diff
@@ -95,7 +95,7 @@ func NewBuiltinApp(conf config.Service, proxy proxy.Service, httpOut httpout.Ser
 func Manifest(conf config.Config) apps.Manifest {
 	return apps.Manifest{
 		AppID:       AppID,
-		Version:     apps.AppVersion(conf.BuildConfig.BuildHashShort),
+		Version:     apps.AppVersion(conf.PluginManifest.Version + "-" + conf.BuildHashShort),
```
levb, line 98: This may not be a valid version strings in some contexts, but appears to work. I can take this change out if controversial.

hanzei: I'm hesitant to add that as it might break at some point and that might be hard to debug.

Also: What is the value of building the version like this?

hanzei (approved):

server/builtin/app.go
```diff
@@ -95,7 +95,7 @@ func NewBuiltinApp(conf config.Service, proxy proxy.Service, httpOut httpout.Ser
 func Manifest(conf config.Config) apps.Manifest {
 	return apps.Manifest{
 		AppID:       AppID,
-		Version:     apps.AppVersion(conf.BuildConfig.BuildHashShort),
+		Version:     apps.AppVersion(conf.PluginManifest.Version),
```
levb, line 98: Still a change, but more appropriate now

levb:
@DHaussermann this is a noncontroversial change, so can you please either prioritize or ok merging before you test. The main change to test is actually that the Telemetry ID still works in "official" CI builds. cc @larkox

DHaussermann (approved):
@levb I have tested to ensure the build hash is visible as expected and version working. 

There is no practical way to test the telemetry ID pre-merge. This can be tested post-merge. 

LGTM!

PR #282, opened 2021-11-01
Simplified use of local strings

levb:
#### Summary
Simplified the use of localized strings in the app
- refactored functionality to conf.Local[WithTemplate]
- added loading the `en` map on activate, using it for default strings, and logging keys that do not match.


larkox (changes requested):
This simplifies very much the code, but it breaks most of the automatization of the i18n.

I am not sure that now, when we run i18n-extract we get the new ids. Also, we don't get the defaults. And since we don't get the defaults, we don't get changes notifications in the other languages.

Any plan to solve those issues?

hanzei (changes requested):
To add another point: I really like having the default text directly in the code. That makes searching for it _much_ easier.

PR #283, opened 2021-11-02
[MM-39731] Don't request bindings from app that doen't have granted locations

hanzei:
#### Summary
If an app doesn't have any granted locations, it not longer get's a bindings call. This PR also adds a test for that.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-39731


levb:
Merging - a minor fix, covered with tests, does not impact any current apps.

PR #285, opened 2021-11-03
Apply i18n-extract-server and i18n-extract-merge and translate missin…

larkox:
#### Summary
Apply i18n-extract-server and i18n-extract-merge and translate missing strings

#### Ticket Link
None

PR #286, opened 2021-11-03
Add check for i18n files to CI

hanzei:
#### Summary
Fixated the version of `go-i18n` to `v2.1.2` to avoid compatibility issues in the future.

#### Ticket Link
https://github.com/mattermost/mattermost-plugin-apps/pull/285#issuecomment-958752420

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -33,6 +33,17 @@ jobs:
           command: |
             make mock
             git --no-pager diff --exit-code server/mocks* || (echo "Please run \"make mock\" and commit the changes in the generated files." && exit 1)
+  check-i18n:
+    docker:
+      - image: circleci/golang:1.16.0
+    steps:
+      - checkout
+      - run:
+          name: Checking if extracted i18n files are up to date
+          command: |
+            go install github.com/nicksnyder/go-i18n/v2/goi18n@v2.1.2
+            make i18n-extract-server
+            git --no-pager diff --exit-code assets/i18n/* || (echo "Please run \"make i18n-extract-server\" and commit the changes in the generated files." && exit 1)
 
   test-e2e-postgres11:
     docker:
@@ -288,6 +299,7 @@ workflows:
                 - master
     jobs:
       - check-mocks
+      - check-i18n
       - plugin-ci/lint
       - plugin-ci/test
       - e2e-cypress-tests-pinned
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -33,6 +33,17 @@ jobs:
...
+  check-i18n:
+    docker:
+      - image: circleci/golang:1.16.0
+    steps:
+      - checkout
+      - run:
+          name: Checking if extracted i18n files are up to date
+          command: |
+            go install github.com/nicksnyder/go-i18n/v2/goi18n@v2.1.2
+            make i18n-extract-server
```
larkox, line 45: We also need to run i18n-merge-server

larkox (changes requested):

.circleci/config.yml
levb: Just for my understanding, I thought `merge` was to merge back provided there were changes to the translation files, but if we just ran `extract`, and `merge` follows immediately, is it not a `no-op`?

hanzei: I have the same understanding as @levb. `merge` used to merge back `translate.*` files, which never should be committed. In fact, they are `gitignored`.

hanzei: @larkox Do I misunderstand the flowß

larkox: If I understood correctly, extract creates the translate.* files, and merge merge the translate.* files into the active.* files. I may be wrong, I haven't looked enough into the code on how this is done, but the log messages seem to say that:

> Please update your assets/i18n/translate.*.json files and execute "make i18n-merge-server"
If you don't want to change any locale file, simply remove the assets/i18n/translate.??.json file before calling "make i18n-merge-server"

hanzei: `make i18n-merge-server` would generate a diff if not all the translation are up to date. To me failing is that case seems not correct, as translation can get updated later in the release process. 

It's worth pointing out that removing the `translate.*` files before running `make i18n-merge-server` errors with:
```
$ make i18n-merge-server 
open assets/i18n/translate.*.json: no such file or directory
make: *** [build/custom.mk:21: i18n-merge-server] Error 1
```

larkox (approved):

hanzei:
@levb PTAL

levb (approved):

PR #288, opened 2021-11-05
MM-39798: expand channel and team membership

levb:
#### Summary
Added support for Expand.(Channel|Team)Member and test

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-39798

No testing required since it's not used yet


levb:
@hanzei @catalintomai 1/5 we can file a few HW tickets to add more `Expand` tests, following the pattern here. Should we do it? If so, in GitHub alone, or start in Jira?

catalintomai:
@levb , I'd say start in Jira, mainly for tracking purposes.

apps/expand.go
```diff
@@ -48,6 +48,10 @@ type Expand struct {
 	// DisplayName, Name
 	Channel ExpandLevel `json:"channel,omitempty"`
 
+	// ChannelMember: expand model.ChannelMember if ChannelID and
+	// ActingUserID are set.
```
hanzei, line 52: Is there a difference between `all` and `summary`?

apps/expand.go
```diff
@@ -56,10 +60,13 @@ type Expand struct {
 	Post     ExpandLevel `json:"post,omitempty"`
 	RootPost ExpandLevel `json:"root_post,omitempty"`
 
-	// Team: all for model.team, summary for Id, DisplayName, Name, Description,
+	// Team: all for model.Team, summary for Id, DisplayName, Name, Description,
 	// Email, Type.
 	Team ExpandLevel `json:"team,omitempty"`
 
+	// TeamMember: expand model.TeamMember if TeamID and ActingUserID are set.
+	TeamMember ExpandLevel `json:"team_member,omitempty"`
```
hanzei, line 68: Could we ever made a case for a `TeamMember` expand for `User`? Should the team and channel expand for  `ActingUser` make it more clear that that it's about `ActingUser`, by e.g. being named `ActingTeamMember`?

server/proxy/expand.go
```diff
@@ -261,7 +291,12 @@ func stripApp(app apps.App, level apps.ExpandLevel) *apps.App {
 	return nil
 }
 
-func getExpandClient(app apps.App, conf config.Config, mm *pluginapi.Client, in Incoming) (mmclient.Client, error) {
+func (p *Proxy) getExpandClient(app apps.App, in Incoming) (mmclient.Client, error) {
+	if p.expandClientOverride != nil {
+		return p.expandClientOverride, nil
+	}
```
hanzei, line 297: If a `pluginapi.Client` is passed to `getExpandClient` this workaround is not longer needed. Why does the `Proxy` need a `expandClientOverride`?

server/proxy/expand_test.go
```diff
@@ -0,0 +1,192 @@
...
+		{
+			name: "channel_member",
+			tcs: map[string]TC{
+				"happy": {
+					base: apps.Context{
+						ActingUserID:     userID,
+						UserAgentContext: apps.UserAgentContext{ChannelID: channelID},
+					},
+					expectClientCalls: func(client *mock_mmclient.MockClient) {
+						client.EXPECT().GetChannelMember(channelID, userID).Times(4).Return(channelMember, nil)
```
hanzei, line 64: Why 4 times? I was expecting 2 times.

hanzei (approved):
Nice work on the tests :+1: 

I'm wondering they already should be split up in multiple test functions are debugging such a large and nested test can be hard.

apps/expand.go
levb: no, it's a small-ish struct

server/proxy/expand_test.go
levb: Good point. Presently expandContext() loads entities for `expand.XXX != ""` and then strips them. This is wrong. I'll file a ticket and cut another PR to fix them all? https://mattermost.atlassian.net/browse/MM-39935

apps/expand.go
levb: Going to merge as is, @hanzei if you feel strongly that "ActingUser" prefix is meaningful, I'll do another PR for that.

PR #294, opened 2021-11-16
[MM-40134] Reduce appID length to 26 characters

hanzei:
#### Summary
To avoid issues like the one in https://mattermost.atlassian.net/browse/MM-34968 and to unify the length with other IDs length in the MM ecosystem this PR reduces the appID length to 26 characters.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-40134

### DIFF
--- a/apps/manifest.go
+++ b/apps/manifest.go
@@ -216,7 +216,7 @@ type AppID string
 
 const (
 	MinAppIDLength = 3
-	MaxAppIDLength = 32
+	MaxAppIDLength = 26
 )
 
 func (id AppID) Validate() error {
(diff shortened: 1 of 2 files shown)


levb (commented):
@hanzei I am confused as to the reason for the limit. What is the benefit of limiting it at 26? Is it because we are adding the AppID to the database now as an "ID"?

Otherwise I don't quite understand why we should be imposing the limit, especially given that the typical MM IDs are _exactly_ 26 characters, and the AppID would vary.

larkox (approved):

hanzei:
>  I am confused as to the reason for the limit. What is the benefit of limiting it at 26? Is it because we are adding the AppID to the database now as an "ID"?

That is one reason. Given that it's used in the DB, we need a limit. Alternately, we could hash the appID and use that, as the list of AppID's is known. The other reason is that the bot user name is the appID. Long user names are shown badly in the UI and hence should be short.

hanzei:
Closing as there is no need for a stricter limit right now.

PR #295, opened 2021-11-18
Fix subpath for local requests

mickmister:
#### Summary

When we create a `Client4` struct for local HTTP requests, we construct the URL manually using the `ListenAddress`, but we don't take into account the server's configured subpath on the site URL. This PR uses the `SiteURL` to calculate the subpath for the constructed URL.

#### Ticket Link

Fixes https://mattermost.atlassian.net/browse/MM-40195

### DIFF
--- a/server/config/config.go
+++ b/server/config/config.go
@@ -112,7 +112,7 @@ func (conf *Config) Update(stored StoredConfig, mmconf *model.Config, license *m
 			host = "127.0.0.1"
 		}
 
-		localURL = "http://" + host + ":" + port
+		localURL = "http://" + host + ":" + port + mattermostURL.Path
 	}
 
 	conf.StoredConfig = stored
--- a/server/httpin/restapi/restapitestlib.go
+++ b/server/httpin/restapi/restapitestlib.go
@@ -11,6 +11,7 @@ import (
 	"net"
 	"net/http"
 	"net/http/httptest"
+	"net/url"
 	"os"
 	"testing"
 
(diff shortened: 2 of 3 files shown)


levb (approved):

catalintomai (approved):

hanzei (commented):
https://github.com/mattermost/mattermost-plugin-apps/blob/93953878af0d3cd97ce02a528f9c04fa90b50e8d/server/httpin/restapi/restapitestlib.go#L80 should also get fixed. Should we also add an (go) e2e test for a server that runs in a subpath?

Nit pick: https://github.com/mattermost/mattermost-plugin-apps/blob/8fe1e6fe86e3960bb147cdee87d9f57450d0d455/server/plugin_test.go#L41 might also be a good one to update.

hanzei:
@mickmister Does this PR fix https://mattermost.atlassian.net/browse/MM-40195?

mickmister:
@hanzei I'm not sure if the second suggestion above needs to be applied. Should we _always_ test with a subpath in the tests?

hanzei (approved):

hanzei:
> @hanzei I'm not sure if the second suggestion above needs to be applied. Should we _always_ test with a subpath in the tests?


I like that suggestion. Filled https://mattermost.atlassian.net/browse/MM-40287

DHaussermann (approved):
Tested and passed
- Confirmed install and uninstall are now working on the subpath server
- No issues fetching icons for command line or post menu
- Briefly regression tested a few test-app commands

No issues found. 

LGTM!

PR #305, opened 2022-03-15
force /cloud test serversand community-xxx to developer mode

levb:
#### Summary
Force developer mode on the `/cloud` test servers and community[-daily, etc]

#### Ticket Link
Fixes https://mattermost.atlassian.net/browse/MM-42459


server/config/config.go
```diff
@@ -58,12 +59,11 @@ type Config struct {
 	DeveloperMode       bool
 	MattermostCloudMode bool
 
-	BotUserID              string
-	MattermostSiteHostname string
```
levb, line 62: `MattermostSiteHostname` does not appear to be used

catalintomai (approved):

server/builtin/install.go
```diff
@@ -11,7 +11,8 @@ import (
 )
 
 func (a *builtinApp) installCommandBinding(loc *i18n.Localizer) apps.Binding {
-	if a.conf.Get().MattermostCloudMode {
+	conf := a.conf.Get()
+	if conf.MattermostCloudMode && !conf.DeveloperMode {
```
hanzei, line 15: Why should this command _not_ work in dev mode?

server/config/config.go
```diff
@@ -112,23 +120,25 @@ func (conf *Config) Update(stored StoredConfig, mmconf *model.Config, license *m
...
 
 	conf.MaxWebhookSize = 75 * 1024 * 1024 // 75Mb
 	if mmconf.FileSettings.MaxFileSize != nil {
 		conf.MaxWebhookSize = *mmconf.FileSettings.MaxFileSize
 	}
 
 	conf.DeveloperMode = pluginapi.IsConfiguredForDevelopment(mmconf)
+	if devHostRegexp.MatchString(u.Hostname()) {
+		conf.DeveloperMode = true
+	}
```
hanzei, line 141: `DeveloperMode` is also the _unsafe_ mode. I'm concerned about using that mode in somewhat production instance. At least community is one.

Is there another approach that we can take? Like enabling `http` Apps on Cloud servers?

hanzei (dismissed):

server/config/config.go
mickmister: +1 for just enabling http instead of dev mode

server/builtin/install.go
levb: This is the "cloud-only" mode of the command, no sub-commands.

server/config/config.go
levb: I will introduce a separate flag for "allow install HTTP".

levb:
@DHaussermann now includes #307

server/builtin/install.go
```diff
@@ -35,6 +37,8 @@ func (a *builtinApp) installCommandBinding(loc *i18n.Localizer) apps.Binding {
 			},
 		}
 	}
+
+	// install http|listed
```
hanzei, line 41: Why `listed`? Isn't that what the previous case is about?

server/config/config.go
```diff
@@ -83,20 +85,32 @@ func (conf Config) StaticURL(appID apps.AppID, name string) string {
 	return conf.AppURL(appID) + "/" + path.Join(appspath.StaticFolder, name)
 }
 
+// developerModeDomains is the list of domains for which DevelopmentMode will be
+// forced on. Empty for the time being.
+var developerModeDomains = regexp.MustCompile("^" + strings.Join([]string{}, "|") + "$")
```
hanzei, line 90: 1/5 that DevelopmentMode should never be enabled by in an instance on the public internet. What use case did you had in mind?

server/config/config.go
```diff
@@ -112,23 +126,30 @@ func (conf *Config) Update(stored StoredConfig, mmconf *model.Config, license *m
...
 	if mmconf.FileSettings.MaxFileSize != nil {
 		conf.MaxWebhookSize = *mmconf.FileSettings.MaxFileSize
 	}
 
 	conf.DeveloperMode = pluginapi.IsConfiguredForDevelopment(mmconf)
+	if developerModeDomains.MatchString(u.Hostname()) {
+		conf.DeveloperMode = true
+	}
+
+	conf.AllowHTTPApps = !conf.MattermostCloudMode || conf.DeveloperMode
```
hanzei, line 149: Why should customers on a cloud server not be able to install HTTP apps?

server/builtin/install.go
levb: the first use case (cloud) does not need a subcommand since it only exposes the listed apps, thus `/apps install appid`. The second use-case differentiates between `/apps install listed appid`, identical handler to the first one, and `/apps install http url`. The latter allows to install any (allowed) app deployment mode, from an HTTP manifest.

4/5 that this all needs a refactoring into a few simple tables, but not ATM?

server/config/config.go
levb: My dev instance is on ngrok, therefore on the public internet. Having said that, I do not have a specific use case, leftover code that I was hesitant to remove, no problem removing altogether.

server/config/config.go
levb: This is an @aaronrothschild question, not a technical one, correct?

mickmister: @hanzei @levb I believe the blocker for this is the ongoing effort for oauth scopes. I'm 0/5 on if we should make that block this though

mickmister (approved):

levb:
@DHaussermann ping to review

DHaussermann (approved):
Tested and passed
Confirmed that the commands to debug and install http app are available in cloud mode when the domain matches the regex.

@levb there's no further functional testing I can do at this time. Please merge. I will revisit this and open a new issues if any changes are needed.

PR #311, opened 2022-03-25
bundle ServiceNow 0.3.2 on Mattermost Cloud

levb:
@DHaussermann there is no real way yet to test this short of deploying to production.


server/builtin/info.go
```diff
@@ -40,5 +43,10 @@ func (a *builtinApp) info(_ *incoming.Request, creq apps.CallRequest) apps.CallR
 		},
 		TemplateData: conf.InfoTemplateData(),
 	}) + "\n"
+
+	if conf.DeveloperMode && conf.AWSAccessKey != "" {
+		out += fmt.Sprintf("AWS config: region: `%s`, bucket: `%s`, access: `%s`, secret: `%s`",
```
hanzei, line 48: Should this be localized?

levb: I'd say, not, but 0/5. this is debugging only.

hanzei: Given that the info command is available even if Debug mode is disabled and the rest of the command is localized, I'm 2/5 that this addition should also be localized.

go.mod
```diff
@@ -12,7 +12,7 @@ require (
 	github.com/hashicorp/go-multierror v1.1.1
 	github.com/mattermost/mattermost-plugin-api v0.0.22-0.20211210183909-beb4761e4bd3
 	github.com/mattermost/mattermost-server/v6 v6.0.0-20220315170027-7710a2fe3741
-	github.com/nicksnyder/go-i18n/v2 v2.1.2
+	github.com/nicksnyder/go-i18n/v2 v2.2.0
```
hanzei, line 15: Would you please also update https://github.com/mattermost/mattermost-plugin-apps/blob/c14f3d4a76de3b8801e456cd83be56b59ce78746/.circleci/config.yml#L95?

hanzei (dismissed):

DHaussermann (approved):
@levb There is no functional testing applicable on this.   Please merge.

PR #291, opened 2021-11-09
[MM-39912] App specific session

hanzei:
#### Summary
This PR implements [the app specific session proposal](https://docs.google.com/document/d/1tSs39Qkp9lGorGdPU8Wid9jWOyxXZXyiWMWcAg_Z3-M/edit).

Beside that the other big change is the introduction of `request.Context`, which is used pass context of an incoming request though the stack. It's in a separate package to ensure only the API is used to access it. It also contains a `context.Context`, which is used for cancellation of outgoing requests.

Kubless support is getting dropped as discussed offline.

~~CI will fail until https://github.com/mattermost/circleci-orbs/pull/27 is merged and released.~~

~~Please note that the API layer method still need to get updated to take the app session into account.~~

~~It needs to be run against the `ben_tmp-sessions` branch in the server repo.~~

### ToDo
- [x] Add more tests
- [x] Fix e2e-tests
- [x] Fix tests
- [x] Add app checks to all app service endpoints
  - [x] Remote OAuth
  - [x] Subscriptions
  - [x] KV

### Future Considerations
- I would like to pass `request.Context` through the whole stack i.e. also to the store and the appservices. That seemed out of scope for this PR to me. Yes
- Should the bot access token also be an OAuth2 one? Yes
- How should access to the KV store be handled? Should every user have access to it or only the bot account? It should be per user: https://mattermost.atlassian.net/browse/MM-40059

#### Ticket Link
Fixes https://github.com/mattermost/mattermost-plugin-apps/issues/270
https://mattermost.atlassian.net/browse/MM-31117
https://mattermost.atlassian.net/browse/MM-34377
https://mattermost.atlassian.net/browse/MM-39825
https://mattermost.atlassian.net/browse/MM-39912
https://mattermost.atlassian.net/browse/MM-39917
https://mattermost.atlassian.net/browse/MM-39920
https://mattermost.atlassian.net/browse/MM-40011
https://mattermost.atlassian.net/browse/MM-40058
https://mattermost.atlassian.net/browse/MM-40059

### Breaking Changes
-  Updates some of the `appsclient` methods

The following apps need to get updated:
- [x] Examples
- [x] Test app: https://github.com/mattermost/mattermost-app-test/pull/13
- [x] ServiceNow: https://github.com/mattermost/mattermost-app-servicenow/pull/36
- [x] Zendesk: https://github.com/mattermost/mattermost-app-zendesk/pull/90
- [ ] gCal

PR #314, opened 2022-03-31
Fix bindings refresh and OAuth2 user

levb:
Fixes https://github.com/mattermost/mattermost-app-servicenow/issues/43

### DIFF
--- a/server/appservices/oauth2.go
+++ b/server/appservices/oauth2.go
@@ -14,7 +14,13 @@ import (
 	"github.com/mattermost/mattermost-plugin-apps/utils"
 )
 
-func (a *AppServices) StoreOAuth2App(r *incoming.Request, appID apps.AppID, actingUserID string, oapp apps.OAuth2App) error {
+func (a *AppServices) StoreOAuth2App(r *incoming.Request, appID apps.AppID, actingUserID string, data []byte) error {
+	var oapp apps.OAuth2App
+	err := json.Unmarshal(data, &oapp)
+	if err != nil {
+		return err
+	}
+
 	app, err := a.store.App.Get(r, appID)
 	if err != nil {
 		return err
@@ -23,11 +29,18 @@ func (a *AppServices) StoreOAuth2App(r *incoming.Request, appID apps.AppID, acti
 		return utils.NewUnauthorizedError("%s is not authorized to use remote OAuth2", app.AppID)
 	}
 
+	oldData, _ := json.Marshal(app.RemoteOAuth2)
+	if bytes.Equal(oldData, data) {
+		return nil
+	}
+
 	app.RemoteOAuth2 = oapp
 	err = a.store.App.Save(r, *app)
 	if err != nil {
 		return err
 	}
+
+	a.conf.MattermostAPI().Frontend.PublishWebSocketEvent(config.WebSocketEventRefreshBindings, map[string]interface{}{}, &model.WebsocketBroadcast{})
 
 	return nil
 }
@@ -53,13 +66,17 @@ func (a *AppServices) StoreOAuth2User(r *incoming.Request, appID apps.AppID, act
 	if err != nil {
 		return err
 	}
-
-	// Trigger a bindings refresh if the OAuth2 user was updated
-	if !bytes.Equal(data, oldData) {
-		a.conf.MattermostAPI().Frontend.PublishWebSocketEvent(config.WebSocketEventRefreshBindings, map[string]interface{}{}, &model.WebsocketBroadcast{UserId: actingUserID})
+	if bytes.Equal(data, oldData) {
+		return nil
 	}
 
-	return a.store.OAuth2.SaveUser(r, appID, actingUserID, data)
+	err = a.store.OAuth2.SaveUser(r, appID, actingUserID, data)
+	if err != nil {
+		return err
+	}
+
+	a.conf.MattermostAPI().Frontend.PublishWebSocketEvent(config.WebSocketEventRefreshBindings, map[string]interface{}{}, &model.WebsocketBroadcast{UserId: actingUserID})
+	return nil
 }
 
 // GetOAuth2User returns the stored OAuth2 user data for a given user and app.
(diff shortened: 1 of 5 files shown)


catalintomai (approved):

server/appservices/oauth2.go
```diff
@@ -53,13 +66,17 @@ func (a *AppServices) StoreOAuth2User(r *incoming.Request, appID apps.AppID, act
 	if err != nil {
 		return err
 	}
+	if bytes.Equal(data, oldData) {
+		return nil
+	}
```
mickmister, line 71: So the main change here is that we don't call `a.store.OAuth2.SaveUser` if the `OAuth2User` has not changed. Is that the main fix in the PR?

DHaussermann:
@levb  Not sure what's happening here but when I try and get Service Now  installed to see if it seems resolved I get an an error that blocks me.

image

I've seen this in the past but not on a clean state. Not sure if this may be a regression. @mickmister let me know if you have a chance to try this as well. 

I will investigate this more to see if it's intermotent.

mickmister (approved):

DHaussermann:
@levb probably nothing but -  I also wonder if the PR branch was forked from current `master` the version stamp is `v0.7.0` which I found a bit curious.

server/appservices/oauth2.go
```diff
@@ -82,10 +99,5 @@ func (a *AppServices) GetOAuth2User(r *incoming.Request, appID apps.AppID, actin
 		return nil, err
 	}
 
-	if len(data) == 0 {
-		// Ensure valid json is returned even if no data is set yet
-		data = []byte(string("{}"))
-	}
```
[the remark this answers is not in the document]
levb: This was added in https://github.com/mattermost/mattermost-plugin-apps/pull/291/files#diff-c28599bd884406d0a27efaef130bc5f762e22bd7da83921df5ae07d8fc535c12R85, but I don't understand why. It makes the OAuth.User to be always non-nil, and makes it more difficult to check if one is set up or not, in the app.

@hanzei what was the motivation?

server/appservices/oauth2.go
levb: No, the main fix is adding the refresh bindsings websocket message to SaveOAuth2App

levb:
> @levb probably nothing but - I was also wonder if the PR branch was forked from current `master` the version stamp is `v0.7.0` which I found a bit curious.

I did on auto-pilot, but we usually cut plugin releases from `master`, right? So what's the difference? I could re-target the 1.0 "bump" branch if it makes it easier for you to test, but it's the same code, just different version stamps.

mickmister:
> I could re-target the 1.0 "bump" branch if it makes it easier for you to test, but it's the same code, just different version stamps.

1/5 I think it's best to keep the process simple, and have the version bump be the "final" change in the process. Seeing `0.7` as the version in the UI can be confusing, but you can look at the commit hash if you're concerned about what's running @DHaussermann

DHaussermann:
@levb I still see this issue occurring where then install is failing with bot user ID. 

- Used `make nuke` on my local
- Confirmed I'm running Apps plugin with build hash from this last PR commit
- Tried install Service now over http
```
{"timestamp":"2022-03-31 14:29:47.950 -04:00","level":"debug","msg":"Received HTTP request","caller":"web/handlers.go:156","method":"POST","url":"/api/v4/users/status/ids","request_id":"yn66hreuhtff9c7bq9szkhasjo","status_code":"200"}
{"timestamp":"2022-03-31 14:29:49.959 -04:00","level":"debug","msg":"App install flow: Created Bot Account ","caller":"app/plugin_api.go:934","plugin_id":"com.mattermost.apps","user_id":"di84uyri5bdpmxqrf5h9df989r","request_id":"ao84a18o4b8uxydmz1phu73pkw","app_id":"servicenow","username":"servicenow"}
{"timestamp":"2022-03-31 14:29:49.960 -04:00","level":"debug","msg":"Received call response","caller":"app/plugin_api.go:934","plugin_id":"com.mattermost.apps","app_id":"apps","request_id":"tirpqqbpw7ff9ei9ohif41sj4y","path":"/api/v1/call","error":"failed to install App: bot user id: f78ipm3topyipdq3hptcotg46o","type":"error","call_path":"/install-consent","user_id":"di84uyri5bdpmxqrf5h9df989r"}
```

Any suggestions on next steps?

DHaussermann:
@levb sorry the issue above was due to building with local changes that should have been discarded - My mistake.

The installation is now working and so far I can't reproduce the oauth storage issue. Testing is ongoing.

server/appservices/oauth2.go
```diff
@@ -23,12 +29,19 @@ func (a *AppServices) StoreOAuth2App(r *incoming.Request, appID apps.AppID, acti
...
+		return nil
+	}
+
 	app.RemoteOAuth2 = oapp
 	err = a.store.App.Save(r, *app)
 	if err != nil {
 		return err
 	}
 
+	a.conf.MattermostAPI().Frontend.PublishWebSocketEvent(config.WebSocketEventRefreshBindings, map[string]interface{}{}, &model.WebsocketBroadcast{})
```
hanzei, line 43: This seems quite dangerous. Forcing a bindings refresh on _every_ client could generate a big load. The bindings cache will also miss given that it will depend often on the oauth app.

server/appservices/oauth2.go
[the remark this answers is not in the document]
hanzei: This was needed because the Api endpoint to get the oauth user must return a valid json payload and not nil in the 404 case.

server/appservices/oauth2.go
levb: We have no way of refreshing bindings for a specific app :(

For all other installed apps, the bindings would've been cached, but for SN all clients need to refresh

hanzei: Why do we need to refresh the bindings immediately? Client will fresh the bindings soon anyway (channel switch, reconnect event, ...)

PR #316, opened 2022-04-05
[MM-42915] Check if oauth apps are enabled before running any /apps commands

mickmister:
#### Summary

This check was removed in apps plugin `1.0`

#### Ticket Link

Fixes https://mattermost.atlassian.net/browse/MM-42915


catalintomai:
> This check was removed in apps plugin `1.0`

@mickmister - would help to point to the change/commit that removed it to provide full context here(e.g. - why was it removed?)

mickmister:
@catalintomai Looks like it was a drive-by change in this refactor PR https://github.com/mattermost/mattermost-plugin-apps/pull/251

catalintomai (approved):

levb (approved):

server/builtin/app.go
```diff
@@ -256,3 +264,12 @@ func (a *builtinApp) newLocalizer(creq apps.CallRequest) *i18n.Localizer {
 
 	return a.conf.I18N().GetUserLocalizer(creq.Context.ActingUser.Id)
 }
+
+func (a *builtinApp) checkConfigValid() error {
+	oauthEnabled := a.conf.MattermostConfig().Config().ServiceSettings.EnableOAuthServiceProvider
+	if oauthEnabled == nil || !*oauthEnabled {
+		return errors.Errorf(oauthConfigErrorMessage, a.conf.Get().MattermostSiteURL)
+	}
+
```
hanzei, line 273: Should we also check if bot account creation is enabled as in https://github.com/mattermost/mattermost-plugin-apps/blob/52ed70d8eeb68b3c73b41549143736e2065b62db/server/command/service.go#L258-L261 ?

mickmister: @hanzei The apps plugin is now using the plugin API to create bots, which bypasses this config value. Since the config value is now irrelevant in regards to the Apps framework, I've chosen to leave the check out of here.

hanzei: Good point :+1:

server/builtin/app.go
```diff
@@ -30,6 +31,8 @@ const (
 	AppID          = "apps"
 	AppDisplayName = "Mattermost Apps plugin"
 	AppDescription = "Install and manage Mattermost Apps"
+
+	oauthConfigErrorMessage = "The system setting `Enable OAuth 2.0 Service Provider` needs to be enabled in order for the Apps plugin to work. Please go to %s/admin_console/integrations/integration_management and enable it."
```
hanzei, line 35: Should this be localized similar to the other output to slash command?

hanzei (approved):

mickmister:
@DHaussermann Maybe this is something @dipak-demansol can test?

DHaussermann:
@mickmister I see an issue on this PR that maybe you could try to reproduce. 
When `Enable OAuth 2.0 Service Provider` is set to `false` the /apps binding no longer appears. 
The biding response shows only `null`
image


This is different from cases where the binding is occasionally unavailable. The binding remains unavailable after various refreshes and channel changes. 

This does not occur when oAuth Apps are enabled.
This does not occur on the marketplace build

DHaussermann:
@mickmister the issue mentioned above ☝️  https://github.com/mattermost/mattermost-plugin-apps/pull/316#issuecomment-1101563210 does not reproduce on the production release. Does it seem likely this PR introduces it? 

I can re-test `master` if needed.

mickmister:
@levb Can you please take a look at the remaining issues for this PR. No worries if you think it can wait

hanzei (approved):

server/builtin/app.go
hanzei: Blocking on this

hanzei (changes requested):

hanzei (approved):

javaguirre (approved):
Screenshot 2022-12-05 at 11 51 12

I tested a couple commands and this seems to work fine @levb  👍

hanzei:
Thanks for testing the PR @javaguirre :+1:

PR #322, opened 2022-04-10
MM-43004: better logging

levb:
#### Summary
Better logging

#### Ticket Link
Ref: https://mattermost.atlassian.net/browse/MM-43004


catalintomai (approved):

server/config/service.go
```diff
@@ -114,45 +114,53 @@ func (s *service) Reconfigure(stored StoredConfig, services ...Configurable) err
...
 		if license == nil {
-			s.log.Infof("Failed to fetch license twice. May incorrectly default to on-prem mode.")
+			s.log.Debugf("Failed to fetch license twice. May incorrectly default to on-prem mode")
 		}
 	}
 
-	err := newConfig.Update(stored, mmconf, license, s.log)
+	err := newConfig.update(stored, mmconf, license, s.log)
 	if err != nil {
+		s.log.WithError(err).Infof("failed to update config")
```
hanzei, line 133: Log _and_ return seems off. Why can't the caller log?

server/plugin.go
```diff
@@ -163,17 +163,18 @@ func (p *Plugin) OnDeactivate() error {
 }
 
 func (p *Plugin) OnConfigurationChange() (err error) {
+	if p.conf == nil {
+		// pre-activate, nothing to do.
+		return nil
+	}
+	p.log.Debugf("OnConfigurationChange")
```
hanzei, line 170: Why is this needed? It seems spamy.

server/proxy/service.go
```diff
@@ -212,15 +212,15 @@ func (p *Proxy) initUpstream(typ apps.DeployType, newConfig config.Config, log u
 		up, err := makef()
 		switch {
 		case errors.Cause(err) == utils.ErrNotFound:
-			log.WithError(err).Debugf("Skipped %q upstream: not configured.", typ)
+			log.WithError(err).Debugf("Skipped upstream: %s: not configured.", typ)
```
hanzei, line 215: `%q` is not longer the way to go?

server/config/service.go
mickmister, line 133: +1

```suggestion
		return errors.Wrap(err, "failed to update config")
```

mickmister (approved):

server/config/service.go
levb: @hanzei 0/5 I'd argue that _in general_ errors should be logged where they happen, and the logger used should be the most contextualized.

But I also agree with your statement, and the redundancy of the message bothers me. I tried to untangle this to fix in its entirety, there are other instances of the "log and return" (anti?-)pattern. 1/5 a separate ticket to rework the logging, add consistency/best practices?

server/proxy/service.go
levb: It was escaping them in JSAON output, annoying

server/httpin/service.go
```diff
@@ -16,18 +16,19 @@ import (
 )
 
 type Service interface {
-	ServeHTTP(c *plugin.Context, w http.ResponseWriter, r *http.Request)
+	ServeHTTP(c *plugin.Context, w http.ResponseWriter, req *http.Request)
 }
 
+// TODO <>/<>: 1/5: Can Handler be combined into Service?
```
levb, line 22: @mickmister @hanzei do you know if there were special considerations for why we split those 2? Seem easily combinable, just 1 struct?

server/incoming/request.go
```diff
@@ -68,7 +68,9 @@ func NewRequest(mm *pluginapi.Client, config config.Service, log utils.Logger, s
 		config:         config,
 		Log:            log,
 		sessionService: session,
-		requestID:      model.NewId(),
+
+		// TODO <>/<>: is the incoming Mattermost request ID available, and should it be used?
```
levb, line 72: @hanzei?

server/appservices/service.go
```diff
@@ -21,22 +21,22 @@ var ErrIsABot = errors.New("is a bot")
 type Service interface {
 	// Subscriptions
 
-	Subscribe(r *incoming.Request, sub apps.Subscription) error
-	GetSubscriptions(r *incoming.Request, appID apps.AppID, actingUserID string) ([]apps.Subscription, error)
-	Unsubscribe(r *incoming.Request, sub apps.Subscription) error
+	Subscribe(*incoming.Request, apps.Subscription) error
+	GetSubscriptions(_ *incoming.Request, _ apps.AppID, actingUserID string) ([]apps.Subscription, error)
+	Unsubscribe(*incoming.Request, apps.Subscription) error
```
catalintomai, line 26: @levb , what is the rule for passing `_` (instead of `r`) in the signature and when do we skip it?

catalintomai (approved):

server/appservices/service.go
levb: Generally in the interface declaration the argument names are not required; but for the non-intention-revealing type names like `string` they are used effectively as comments. However if one arg is named, all of them must be named. Conveniently, `_` allows skipping it.

Generally, I tried to reduce the use of names in the interface definitions, where the type itself is self-explanatory. (Since names get changed over time, stylistic ballast)

mickmister (approved):

mickmister (approved):

PR #324, opened 2022-04-14
Avoid stripping context values if call is for built-in app

mickmister:
#### Summary

When the install form is submitted, the Apps framework will send the form submission (including the info about the current channel/team) to the "built-in app" used to orchestrate the state of installed apps. When the call is prepared for the built-in app, the team id and channel id are stripped from the context during the context expansion. The built-in app then tries to use its received context when expanding for the `OnInstall` call, but it doesn't have access to the stripped values.

This PR makes it so the values are only stripped when the call is meant for an app other than the built-in app.

#### Ticket Link

Fixes https://mattermost.atlassian.net/browse/MM-43436

### DIFF
--- a/server/proxy/expand.go
+++ b/server/proxy/expand.go
@@ -151,13 +151,15 @@ func (p *Proxy) expandContext(r *incoming.Request, app apps.App, base *apps.Cont
 		}
 	}
 
-	// Cleanup fields for app
-	cc.UserAgentContext.ChannelID = ""
-	cc.UserAgentContext.TeamID = ""
-	cc.UserAgentContext.RootPostID = ""
-	cc.UserAgentContext.PostID = ""
-	cc.UserID = ""
-	cc.ActingUserID = ""
+	if app.DeployType != apps.DeployBuiltin {
+		// Cleanup fields for app
+		cc.UserAgentContext.ChannelID = ""
+		cc.UserAgentContext.TeamID = ""
+		cc.UserAgentContext.RootPostID = ""
+		cc.UserAgentContext.PostID = ""
+		cc.UserID = ""
+		cc.ActingUserID = ""
+	}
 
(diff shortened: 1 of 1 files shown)


DHaussermann:
As per discussion with @mickmister we should add a unit test around this functionality around this.

levb:
I am a little fuzzy on the intend of this PR. If the built in app wants to receive these IDs, it should add the Expand clause and use `[ExpandedContext].Team.Id` and `.Channel.Id`; why the need for special tricks?

mickmister:
@levb The built-in app receives the `install-consent` call, which unconditionally has the expand below:

```json
{
    "path": "/install-consent",
    "expand": {
        "acting_user": "summary",
        "locale": "all"
    },
    "state": "(id of App)",
    "context": {
        "app_id": "apps",
        "location": "/command/apps/install/http",
        "root_id": "",
        "channel_id": "ue9xwww1gjdsbrnmtph33c874o",
        "team_id": "ef1mommohbnk9jnqkhihoz73sy",
        "track_as_submit": true,
        "user_agent": "webapp"
    },
    "values": {
        "consent": true,
        "deploy_type": {
            "label": "HTTP",
            "value": "http"
        }
    }
}
```

The `on_install` call registered by the App we're installing is not taken into account during this HTTP request from the browser to the plugin. Therefore, there are no instructions to expand the `Team` when the call is processed by the built-in app. When we drop the `team_id` etc. during this time, we lose the context to expand when we do eventually have access to the `expand` clause of the `on_install` call.

mickmister:
Since `OnInstall` is not meant for this sort of thing, the "issue" this PR is solving is not an actual issue. In the future, if the built-in app is to make a call request to another app, it should itself fill in the `UserAgentContext` before sending the call.

PR #325, opened 2022-04-25
Display HTTP URL in /apps list

levb:
Just as the title says, instead of `http`, display `https:/x.y.z` for installed HTTP apps


catalintomai (approved):

mickmister (approved):
Yes!!

mickmister:
Any security issues that may arise from this? Should end-users be able to know where the http server resides?

levb:
Sorry just pushed another update, please review again

levb:
> Any security issues that may arise from this? Should end-users be able to know where the http server resides?

@mickmister no - it's a sysadmin-only command.

catalintomai (approved):

server/proxy/list.go
```diff
@@ -21,19 +21,42 @@ func (p *Proxy) GetInstalledApp(r *incoming.Request, appID apps.AppID) (*apps.Ap
...
-	out := []apps.App{}
-	for _, app := range installed {
-		out = append(out, app)
+func (p *Proxy) GetInstalledApps(r *incoming.Request, ping bool) (installed []apps.App, reachable map[apps.AppID]bool) {
+	all := p.store.App.AsMap(r)
+
+	// all ping requests must respond, unreachable respond with "".
+	reachableCh := make(chan apps.AppID)
+	for _, app := range all {
+		rr, cancel := p.timeoutRequest(r, pingAppTimeout)
```
mickmister, line 30: Can we utilize this timeout logic when fetching bindings from all apps?

mickmister (approved):

server/proxy/list.go
levb: Sure, when the time for it comes? I am not sure we want to be timing out more aggressively than the default, yet, since the binding fetch should be in theory async to the user, right?

server/proxy/call.go
```diff
@@ -178,3 +180,19 @@ func (p *Proxy) getStatic(r *incoming.Request, app apps.App, path string) (io.Re
...
+
+	return err == nil || errors.Cause(err) == utils.ErrNotFound
+}
+
+func (p *Proxy) timeoutRequest(r *incoming.Request, timeout time.Duration) (*incoming.Request, context.CancelFunc) {
+	r = r.Clone()
+	ctx, cancel := context.WithTimeout(r.Ctx(), timeout)
+	incoming.WithCtx(ctx)(r)
+	return r, cancel
+}
```
hanzei, line 198: Should this be a method of `incoming.Request`? What does it have to do with `Proxy`?

levb: See https://github.com/mattermost/mattermost-plugin-apps/pull/331

PR #328, opened 2022-04-27
Fixed the E2E test

levb:
Fixes the E2E test, still with `hello-world`.

Thx, @mickmister!


test/e2e/package.json
```diff
@@ -50,5 +52,8 @@
     "fix": "eslint --ext .ts --ignore-pattern *d.ts . --quiet --fix --cache",
     "cypress:open": "cypress open",
     "test": "cypress run"
+  },
+  "dependencies": {
```
levb, line 56: @mickmister Can you explain this one to me? (I had to add it, but not sure what's pulling it in, and why does it need to be listed)

levb:
image

test/e2e/cypress.json
```diff
@@ -5,7 +5,7 @@
     "taskTimeout": 20000,
     "pluginID": "com.mattermost.apps",
     "pluginFile": "com.mattermost.apps-0.1.0.tar.gz",
```
catalintomai, line 7: @levb - is this the right `pluginFile`?

build/custom.mk
```diff
@@ -70,7 +70,13 @@ i18n-merge-server:
...
+## Run the hello-world app
+.PHONY: run-example-hello-4000
+run-example-hello-4000:
+	cd examples/go/hello-world && go run .
+
 ## Run the test app
-.PHONY: run-test-app
-run-test-app:
+.PHONY: run-test-app-8081
+run-test-app-8081:
```
catalintomai, line 80: do we still need `run-test-app-8081`?

levb: We will, soon; plus it sort of exposes the test app at the `make` level, kind of nice.

test/e2e/cypress.json
levb: I don't think it's used...

catalintomai (approved):

test/e2e/package.json
mickmister: Compared to `devDependencies`, you can sort of view `dependencies` as "public exports" for other packages to have access to if you import "this" project. Since mm-webapp has all of its e2e test deps in its `devDependencies`, we don't get them "for free" when importing mm-webapp into our test code. This is because the webapp's e2e test setup is not necessarily meant to be imported. So when we try to run the bootstrapping code from mm-webapp and we come across importing one of these deps, cypress fails to load the dep.

As a result, we end up having to duplicate these dependencies in order to import the relevant boilerplate test boostrapping code from the webapp. 

2/5 we should have a follow up ticket to remove the dependency on mm-webapp. There are generally two reasons we import:

- General test bootstrapping
- Shared functions such as `getLastPostId`, but those can easily be copied over as needed

mickmister: In a conventional sense, the main difference between `devDependencies` and `dependencies`, is `devDependencies` is meant for tooling, and `dependencies` is meant for packages used at runtime in your application. When a "package" that is not meant to be imported is used like this, we run into issues with missing dependencies.

PR #331, opened 2022-05-02
#325 feedback: moved WithTimeout to incoming.Request methods

levb:
See https://github.com/mattermost/mattermost-plugin-apps/pull/325#discussion_r862727381

### DIFF
--- a/server/proxy/list.go
+++ b/server/proxy/list.go
@@ -4,14 +4,18 @@
 package proxy
 
 import (
+	"context"
 	"sort"
 	"strings"
+	"time"
 
 	"github.com/mattermost/mattermost-server/v6/model"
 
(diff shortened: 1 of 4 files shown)


server/proxy/list.go
```diff
@@ -4,15 +4,19 @@
...
 	"strings"
+	"time"
 
 	"github.com/mattermost/mattermost-server/v6/model"
 
 	"github.com/mattermost/mattermost-plugin-apps/apps"
 	"github.com/mattermost/mattermost-plugin-apps/server/incoming"
 )
 
+const pingAppTimeout = 1 * time.Second
```
mickmister, line 18: I think should be documented that your App respond within 1 second for this request, whether the App implements that endpoint or not

mickmister (approved):

PR #339, opened 2022-05-09
Fixed subscription validation

levb:
This looks like it was a bug/omission but I didn't really have a chance to investigate. This fixed my problem.

@jfrerich I think you are the original author, can you review, please?

### DIFF
--- a/server/appservices/subscriptions.go
+++ b/server/appservices/subscriptions.go
@@ -10,7 +10,6 @@ import (
 
 	"github.com/mattermost/mattermost-plugin-apps/apps"
 	"github.com/mattermost/mattermost-plugin-apps/server/incoming"
-	"github.com/mattermost/mattermost-plugin-apps/utils"
 )
 
 type PermissionChecker interface {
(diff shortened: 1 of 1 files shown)


catalintomai (approved):

server/appservices/subscriptions.go
```diff
@@ -68,10 +67,10 @@ func CheckSubscriptionPermission(checker PermissionChecker, sub apps.Subscriptio
 
 func (a *AppServices) Subscribe(_ *incoming.Request, sub apps.Subscription) error {
 	if err := sub.Validate(); err != nil {
-		return utils.NewInvalidError("invalid subscription")
+		return err
 	}
 
-	if err := CheckSubscriptionPermission(&a.conf.MattermostAPI().User, sub, "", ""); err != nil {
+	if err := CheckSubscriptionPermission(&a.conf.MattermostAPI().User, sub, sub.ChannelID, sub.TeamID); err != nil {
```
hanzei, line 73: Why is this change needed? `CheckSubscriptionPermission` already checks against if the user has permission to the resource associated to `sub`. `channelID, teamID` are used for context checking, e.g. to check if the user has access to a channel for `SubjectBotJoinedChannel`.

hanzei (changes requested):

server/appservices/subscriptions.go
levb: Basically, this entire model seems obsolete, and I am debugging it now. The check was put in place before we used authenticated APIs in expand. Now, the notifications are expanded with the bot-level permissions, so most checks will happen then, inevitably.

A notable exception would be the Join/Leave[e.g. Team] notifications where the Expand.TeamMember=id does not involve an API call, but arguably should still fail if the app's bot does not have access to the team member. There may be more like it.

2/5 I'll probably close this PR once I fully debug/test https://github.com/mattermost/mattermost-plugin-apps/pull/336, you're right this change is suspicious, but so is the current underlying logic.

PR #342, opened 2022-05-19
disabled MessageHasBeenPosted processing

levb:
#### Summary
Disabled MessageHasBeenPosted and the related subscription subjects until we figure out performance issues. This disables both `post_created` and `bot_mentioned` notifications.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-44388

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -34,7 +34,6 @@ linters:
     - misspell
     - nakedret
     - staticcheck
-    - structcheck
     - stylecheck
     - typecheck
     - unconvert
--- a/apps/subscription.go
+++ b/apps/subscription.go
@@ -50,15 +50,18 @@ const (
 	// used to expand other entities.
 	SubjectChannelCreated Subject = "channel_created"
 
+	// TODO: re-enable post_created and bot_mentioned once perf issues are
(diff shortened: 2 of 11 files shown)


.golangci.yml
```diff
@@ -34,7 +34,6 @@ linters:
     - misspell
     - nakedret
     - staticcheck
-    - structcheck
```
levb, line 37: was flagging a false positive for a test struct field, seems generally useless.

aaronrothschild (approved):
These operations are important for Apps to be notified of, is the plan to re-enable them?

levb:
@aaronrothschild We do want to re-enable them for sure, but we need a subscriptions cache to avoid traveling to the KV store on every post. Will file a ticket.

levb:
@aaronrothschild https://mattermost.atlassian.net/browse/MM-44421

hanzei (commented):
Can you please elaborate what causes the performance issue? How many API requests are happening here? Is the issue specifically with the KV calls?

For reference: Autolink does up to three API requests per post (But no KV one)

mickmister (approved):

PR #345, opened 2022-05-19
Release v1.1.0

levb:


catalintomai (approved):

DHaussermann:
I have done some smoke testing on this: 
- Deployed the test app and smoke tested various commands
- Deployed Service now via HTTP and rand through install, configure, authenticate and usage

I did hit a couple snags - (possibly tester error) 
1. I am unable to see and install an app that should be listed and deployed on AWS. Some investigation is required. 

Regarding new developer mode changes:
2A. For the debug changes - it's unclear to me when developer mode **should** be true. I see it as false even when set to true on the MM server
2B. When developer is false I'm still seeing the debug command even though the code change for this should now be in the branch.

@levb Can we discuss these points ☝️ before approving the release in case further changes may be warranted?

DHaussermann:
Issue 1. above was a knowledge gap about needing to re-deploy the app after a `make nuke` has occurred as the deployment process use the URL and token to write data to the server. This data does not survive the `make nuke` 

Issue 2. Was simple stale deployment. The latest pull from `master` was missing in the compiled version   

Tested and passed. 
- Confirmed AWS deployment process is functional
- Confirmed AWS provisioned apps are installable and usable 
- Confirmed Debug mode can be correctly enabled and disabled
- Confirmed `DEBUG` binding no longer shown when developer mode is false. This change will hide debug commands on Cloud servers.

LGTM!

DHaussermann (approved):

hanzei (commented):
Should https://github.com/mattermost/mattermost-plugin-apps/blob/e1f8678e87808ad1e14a685d86c5e3062afdfe1a/dev/docker-compose.yml#L23 get updated?

PR #346, opened 2022-05-21
/apps debug is now dev mode only, other feedback from DH

levb:
#### Summary
- `/apps debug` command is now only visible in dev mode
- @DHaussermann reported seeing HTTP enabled on a production host, so added debugging to see why the regex matches falsely
- Added a small optimization to GetInstalledApps - don't even make a channel if there are no apps installed

#### Ticket Link
None


catalintomai (approved):

hanzei (approved):

levb:
@DHaussermann I am merging this now, you can test in the 1.1 release cycle?

PR #296, opened 2021-11-27
API changes - separate calls etc. Replaces #276.

levb:
**REPLACES:** https://github.com/mattermost/mattermost-plugin-apps/pull/276

#### Summary
Add separate calls for each function on Apps (submit, form source, form submit and lookup). Also updated the filtering with the new logic.

This PR also piggybacks the renaming of markdown and error to text, and filtering out "white space labels" (to avoid no text labels on post options and channel header).

Notes for testing: This is a breaking change. All Apps will need to be updated to work properly. The test app will be updated to fit the new approach.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-39092
https://mattermost.atlassian.net/browse/MM-39096
https://mattermost.atlassian.net/browse/MM-31121
https://mattermost.atlassian.net/browse/MM-33532

#### Related Pull Requests
Webapp: https://github.com/mattermost/mattermost-webapp/pull/9263
Mobile: https://github.com/mattermost/mattermost-mobile/pull/5877
Test App: https://github.com/mattermost/mattermost-app-test/pull/14
ServiceNow: https://github.com/mattermost/mattermost-app-servicenow/pull/37

### Breaking Changes
Call API, routing

#### TODO
The following apps need to get updated:
- [x] Examples
- [x] Test app
- [x] ServiceNow
- [ ] Zendesk
- [x] gCal

PR #350, opened 2022-06-10
Fix and centralize the Node.js app tutorial

azigler:
There are currently two different versions of the same JS tutorial example in this repository, and both are broken. This PR fixes the broken bindings from #296 in the `dev/node_app` example and changes `examples/js` to point to this new example, to reduce confusion and allow for this resource to be more easily maintained in the future. This PR also makes some modifications to the Docker setup in `dev` to ensure networking works correctly in the current version of Docker.

In sum:
- remove duplicate example in examples/js and point to the dev development setup
- fix broken bindings in the Node.js example 
- improve dev Docker setup

See accompanying [mattermost-developer-documentation PR #1075](https://github.com/mattermost/mattermost-developer-documentation/pull/1075) for the tutorial updates that match these new changes.


azigler:
Hey @levb and @mickmister, can I get your review and merge on this JS tutorial fix? Please let me know if you have any questions and feel free to directly edit my updates. 🚀

dev/node_app/package.json
```diff
@@ -11,7 +11,7 @@
   "license": "ISC",
   "dependencies": {
     "express": "4.17.1",
-    "mattermost-redux": "github:mattermost/mattermost-redux#5fa256e1f5ae6ba237a276e35ec23a4c1f73af7b",
+    "mattermost-redux": "5.33.1",
```
mickmister, line 14: FYI we're moving away from `mattermost-redux` for node apps, as all wanted things should be available in new packages `@mattermost/types` and `@mattermost/client` now. It takes way less time to download!!

mickmister:
I'm doing a more thorough review, though I wanted to point :point_up: out

dev/node_app/package.json
azigler: Thanks @mickmister! I tried updating my dependencies like so:

```
"dependencies": {
    "express": "4.17.1",
    "@mattermost/types": "6.7.0-0",
    "@mattermost/client": "6.7.0-0",
    "node-fetch": "2.6.1"
  },
```

Then in `app.ts` I changed the imports to the following:

```
import {AppBinding, AppCallRequest, AppCallResponse, AppForm, AppManifest} from '@mattermost/types/apps';
import {Post} from '@mattermost/types/posts';
import {Channel} from '@mattermost/types/channels';

import {Client4} from '@mattermost/client';
```

However, when trying to run the app, I receive this error:

```
/Users/andrew/techne/__REPOS/mattermost-plugin-apps/dev/node_app/node_modules/@mattermost/client/lib/index.js:3
export { default as Client4 } from './client4';
^^^^^^

SyntaxError: Unexpected token 'export'
```

I believe, based on [this part of the README](https://github.com/mattermost/mattermost-webapp/tree/master/packages#adding-a-new-subpackage) that I need to modify `tsconfig.json` as well, but I'm not sure how to do so for this purpose. Can you help me resolve this?

mickmister: What if you try

```ts
import Client4 from '@mattermost/client/client4';
```

azigler: Just tried that, my error is:

```
Error: Cannot find module '@mattermost/client/client4'
```

Perhaps that is an alias that needs to be set in `tsconfig.json`?

azigler:
> @mickmister Thanks for the initial review. Do you have the bandwidth to help us get this merged before Wednesday? We were hoping to have these fixes implemented for the 7.0 release because we're going to promote these resources to encourage adoption of the App framework.

I'm going to do another pass on these tutorials just to make them tighter, set up the Docker network ideas we're discussing, and move most of the code examples into the repo and out of the tutorial to remove redundancy for future maintenance. Since that's the case, we can also tackle that dropping `mattermost-redux` at that time and make sure we have the best process laid out for the developer.

dev/docker-compose.yml
```diff
@@ -1,33 +1,29 @@
 version: '3'
 services:
     mattermost:
+        container_name: mattermost
         # This can be updated with docker tags from https://hub.docker.com/r/mattermost/mattermost-enterprise-edition/tags
         image: mattermost/mattermost-enterprise-edition:6.6
-
         ports:
-            - "8066:8066"
+            - 8065:8065
```
mickmister, line 8: This was originally `8066` so you could also run the typical Go server on port `8065`. 0/5 on if it should be kept that way

dev/node_app/package.json
mickmister: I think the full path would actually be `@mattermost/client/lib/client4`, but I'm not sure if that will work either.

dev/node_app/src/app.ts
```diff
@@ -86,20 +86,20 @@ const commandBindings = {
...
             label: 'node-example',
-            description: 'Example App written with Node.js',
+            description: 'Example app written with Node.js',
             hint: '[send]',
             bindings: [
                 {
                     location: 'send',
                     label: 'send',
-                    form,
+                    form
```
mickmister, line 95: It's better convention to have the dangling comma here

```suggestion
                    form,
```

dev/node_app/src/app.ts
```diff
@@ -73,7 +73,7 @@ const channelHeaderBindings = {
             location: 'send-button',
             icon: 'icon.png',
             label: 'send hello message',
-            call: {
+            submit: {
                 path: '/send',
```
mickmister, line 77:
```suggestion
                path: '/send/submit',
```

dev/node_app/src/app.ts
```diff
@@ -122,6 +122,26 @@ type FormValues = {
...
+                    name: 'message',
+                    label: 'message',
+                },
+            ],
+            submit: {
+                path: '/send/submit',
+            },
+        },
+    });
+});
```
mickmister, line 143: I would prefer not to duplicate this form, to drive home the fact that the same form can be reused across commands and modals. One of my suggestions above makes it so this will actually no longer be called anyway

mickmister (approved):
Nothing blocking, just same suggestions for improvement

mickmister:
@azigler I see you're developing off the master branch of your fork. I recommend using a different branch, as this could otherwise cause issues with sync'ing between the forks.

> Since that's the case, we can also tackle dropping `mattermost-redux` at that time and make sure we have the best process laid out for the developer.

Makes sense to me :+1: As a side note, the `@mattermost/types` package actually needs an update for new types from App Framework 1.0

azigler:
> @azigler I see you're developing off the master branch of your fork. I recommend using a different branch, as this could otherwise cause issues with sync'ing between the forks.
> 
> > Since that's the case, we can also tackle dropping `mattermost-redux` at that time and make sure we have the best process laid out for the developer.
> 
> Makes sense to me 👍 As a side note, the `@mattermost/types` package actually needs an update for new types from App Framework 1.0

Is there a different branch you recommend I use? Or should I make a new branch in the future for these updates?

dev/docker-compose.yml
azigler: Was that the reason why? I found it odd that this was sitting on `8066` since we use `8065` everywhere.

dev/node_app/src/app.ts
azigler: Can you help me understand how this one would work? The channel header button needs to ask the user for input because they're merely pressing a button instead of passing in parameterized arguments via a slash command. If the button sends them right to `send/submit` then we're skipping the form collection stage. But if we put the form in the `/send/submit` path, then it doesn't make sense for the slash command because they've already passed in the parameter they want, no? Is there an idea way to cater to both that you recommend? Is there some kind of arg checking we could do to determine if they've already passed in parameter(s) via a slash command?

azigler:
Thanks @mickmister for the review! @levb, let me know if you have any questions ahead of merging this. Would love to get this merged soon to coincide with the `7.0` launch, but I'm open to any feedback or concerns you might have. Thanks!

cc @BenLloydPearson

azigler:
@levb @mickmister Hi y'all! Just wanted to follow up to see if we could get a merge on this one? Let me know if either of you have any questions. 

cc @BenLloydPearson

levb (changes requested):

dev/node_app/package.json
azigler: @mickmister If I try `@mattermost/client/lib/client4` I get `Cannot use import statement outside a module`.

azigler:
@levb @mickmister Okay, I understand the confusion now. In a version of this I was working on, the `channel_header` and `command` bindings didn't both work if they used the same form. I thought this was by design but it's likely my local example just wasn't working. Based on y'alls feedback, I (think I) understand now how this should be structured. I did some testing to make sure this works. Everything is more clear to me now. I committed some changes to reflect the new bindings, which removes the redundant form. I also opened [a new PR on `mattermost-developer-documentation`](https://github.com/mattermost/mattermost-developer-documentation/pull/1084) and made relevant updates there.

When this PR is closed, I will move off `master` to my own branch as Michael suggested. Based on the other things we discussed here, a future PR will address:

- using a Docker network with an external repo instead of encouraging development in `dev/`
- moving off `mattermost-redux`

Please let me know if there are still outstanding issues to address here, and thank you for your patience!

azigler:
Closing this PR because `/examples` and `/dev` are moving to https://github.com/mattermost/mattermost-app-examples.

Issue #351, opened 2022-07-13
Request to community: Please submit examples of your Apps :rocket:


mickmister:
Please see this issue for more info https://github.com/mattermost/mattermost-plugin-apps/issues/351

mickmister:
Great thank you @dsharma522!!

mickmister:
Merged https://github.com/mattermost/mattermost-plugin-apps/pull/352

mickmister:
Reopening to use as a sticky issue

javaguirre:
I don't know if there is a shortcut to pin/unpin that I pressed unintendedly, but I didn't want to unpin this xD

mickmister:
Moved this to https://github.com/mattermost/mattermost-plugin-apps/issues/351, but I'll keep this open for visibility

PR #352, opened 2022-07-22
Add example Apps in additional languages - Python

dsharma522:
#### Ticket Link
Github Issue - https://github.com/mattermost/mattermost-plugin-apps/issues/351

This PR adds basic example of Hello-world app in Python language. It also includes the required requirements.txt file which will be needed.


examples/python/hello-world.py
```diff
@@ -0,0 +1,138 @@
...
+import os
+
+import requests
+from flask import Flask, request
+
+logging.basicConfig(level=logging.DEBUG)
+
+APP_PORT = 8080
+APP_HOST = 'localhost'
+ROOT_URL = 'http://localhost:8080'
```
mickmister, line 11: Can this be made configurable using environment variables? The port and root URL specifically should be configurable. Please verify it works with an ngrok URL.

mickmister (changes requested):
Thanks for this contribution @dsharma522 :tada: LGTM, though I have a request to make the `PORT` and `ROOT_URL` configurable via environment variables.

examples/python/hello-world.py
dsharma522: Noted! will make it configurable and push commit.

examples/python/README.md
```diff
@@ -0,0 +1,7 @@
+### Pre-requisite
+- have python installed , preferably `>=3.0`
+- install the requirements mentioned in the `requirement.txt`
+- make required changes in the `settings.json` to run the app on custom url & port
```
mickmister, line 4:
```suggestion
- Have python installed, preferably `>=3.0`
- Install the requirements mentioned in the `requirement.txt`
- Make required changes in the `settings.json` to run the app on custom url & port
```

examples/python/settings.json
```diff
@@ -0,0 +1,5 @@
+{
+    "port": "8080",
+    "root_url": "https://6fe6-103-157-160-168.in.ngrok.io",
+    "host": "0.0.0.0"
+}
```
mickmister, line 5: I think it's confusing how there are 3 sources of configuration (hardcode, settings file, and env vars). I'm thinking we should get rid of this file, as it's not something established among the examples already

examples/python/src/hello-world.py
```diff
@@ -0,0 +1,212 @@
...
+        logging.debug(f'subscribed to team_join event for {site_url}')
+
+
+if __name__ == '__main__':
+    # load settings
+    root_dir = os.path.dirname(os.path.dirname(__file__))
+    with open(f'{root_dir}/settings.json') as config_file:
+        settings = json.load(config_file)
+        for env_name, env_value in settings.items():
+            os.environ[env_name] = env_value
```
mickmister, line 205: If we do use the `settings.json` file, the environment variables should take precedence over the `settings.json` file, rather than the `settings.json` file editing the environment here. At the moment, the environment variables will not do anything unless we delete the `settings.json` file.

examples/python/src/hello-world.py
```diff
@@ -0,0 +1,212 @@
...
+    #    'team_id': 'team_id'   # get this team_id when bot_joined_team event occurs
+    # }
+    return {'type': 'ok', 'data': []}
+
+
+# Subscribing to events. For example, Subscribe to 'bot_joined_team' event
+def _subscribe_team_join(context: dict) -> None:
+    site_url = context['mattermost_site_url']
+    bot_access_token = context['bot_access_token']
+    url = site_url + 'plugins/com.mattermost.apps/api/v1/subscribe'
```
mickmister, line 179: Should we do `path.join` here instead? Does the site URL always end in a slash?

examples/python/README.md
```diff
@@ -0,0 +1,7 @@
+### Pre-requisite
+- have python installed , preferably `>=3.0`
+- install the requirements mentioned in the `requirement.txt`
```
mickmister, line 3: `pip3 install -r requirements.txt` right? We should explain that here

examples/python/src/hello-world.py
```diff
@@ -0,0 +1,212 @@
...
+                            'path': '/first_command',
+                            # expand block is optional. This is more of metadata like which channel, team this command
+                            # was called from
+                            'expand': {
+                                'app': 'all',
+                                # if you want to expand team & channel, ensure that bot is added to the team & channel
+                                # else command will fail to expand the context
+                                # 'team': 'all',
+                                # 'channel': 'all',
+                            },
```
mickmister, line 90: > if you want to expand team & channel, ensure that bot is added to the team & channel
> else command will fail to expand the context

The bot must be a member of the team and channel, otherwise the expand fails and the request is never sent to the App? It seems that we shouldn't show the command to the user in this case. cc @levb

examples/python/src/hello-world.py
```diff
@@ -0,0 +1,212 @@
...
+            {
+                # binding for a command, command with embedded form
+                'location': '/command',
+                'bindings': [
+                    {
+                        'description': 'test command',
+                        'hint': '[This is testing command]',
+                        # this will be the command displayed to user as /second-command
+                        'label': 'second-command',
+                        'icon': 'icon.png',
```
mickmister, line 104: This should instead be in the same array that begins on line 73, rather than creating the new object on line 95

examples/python/src/hello-world.py
```diff
@@ -0,0 +1,212 @@
...
+@app.route('/bot_joined_team', methods=['GET', 'POST'])
+def on_bot_joined_team() -> dict:
+    context = request.json['context']
+    logging.info(
+        f'bot_joined_team event received for site:{context["mattermost_site_url"]}, '
+        f'team:{context["team"]["id"]} name:{context["team"]["name"]} '
+        f'{request.args} {request.data}'
+    )
+    # Here one can subscribe to channel_joined/left events as these required team_id now to be subscribed,
+    # hence use the team_id received in the event and make a call for subscribing to channel_joined/left events.
```
mickmister, line 159: The framework should make sure these subscriptions are removed when the bot is removed from this team cc @levb for discussion

mickmister (changes requested):

examples/python/src/hello-world.py
dsharma522: Use of `path.join` could result in different result based on the platform running this code. 
For example, in Windows that would result into `\` while running on Mac system, that would become `/`. 

We need to use some url specific libs for this.

examples/python/src/hello-world.py
dsharma522: @levb @mickmister , 

I am able to get the `app` in the context without even adding the bot to the team/channel. And as expect expansion of `team` , `channel` failed as it is not the part of any team/channel currently. 

--- 
Steps to reproduce
1. install the app via /apps install http <url> (use the same app binding as in the current PR)
2. run the command `/first-command` and press enter to submit. The command will succeed and won't show any error. 

Sharing raw response from `ngrok` for ref, if needed.

```json
{
    "path": "/first_command",
    "expand": {
        "app": "all"
    },
    "context": {
        "app_id": "hello-world",
        "location": "/command/first-command",
        "user_agent": "webapp",
        "track_as_submit": true,
        "mattermost_site_url": "http://localhost:8066",
        "developer_mode": true,
        "app_path": "/plugins/com.mattermost.apps/apps/hello-world",
        "bot_user_id": "3xxj6snbwby19npr1tca3ys9zo",
        "bot_access_token": "3mf1popcwjnzxyeps9kqsqeuoc",
        "app": {
            "app_id": "hello-world",
            "webhook_secret": "zrp5e37iejgtixpfqtjt43ibua",
            "bot_user_id": "3xxj6snbwby19npr1tca3ys9zo",
            "bot_username": "hello-world",
            "remote_oauth2": {}
        },
        "acting_user": {
            "id": "kredsjwqwbrp9r41c86b69akno",
            "delete_at": 0,
            "username": "",
            "auth_service": "",
            "email": "",
            "nickname": "",
            "first_name": "",
            "last_name": "",
            "position": "",
            "roles": "",
            "locale": "",
            "timezone": null,
            "disable_welcome_email": false
        },
        "oauth2": {}
    },
    "raw_command": "/first-command "
}
```

examples/python/src/hello-world.py
dsharma522: fixed.

examples/python/README.md
dsharma522: Done.

examples/python/src/hello-world.py
dsharma522: done.

examples/python/src/hello-world.py
dsharma522: Removed `settings.json` file and will use the configured environment variables. The app will use default ones in case not configured.

examples/python/README.md
dsharma522: Done

examples/python/README.md
```diff
@@ -1,7 +1,18 @@
...
-- have python installed , preferably `>=3.0`
-- install the requirements mentioned in the `requirement.txt`
-- make required changes in the `settings.json` to run the app on custom url & port
+- Have python installed, preferably `>=3.0`
+- Install the requirements mentioned in the `requirement.txt` with `pip3 install -r requirements.txt`
+- Configure the following environment variables to run the app on custom port/url
+  ```
+  export APP_PORT=8080
+  export ROOT_URL=http://localhost:8080
+  export APP_HOST=localhost
```
mickmister, line 8: Some PaaS systems require you to use `PORT` as the env variable for the port to listen on. Let's stick with `PORT`

examples/python/src/hello-world.py
```diff
@@ -0,0 +1,199 @@
...
+
+import requests
+from flask import Flask, request
+
+logging.basicConfig(level=logging.DEBUG)
+
+app = Flask(__name__, static_url_path='/static', static_folder='./static')
+
+DEFAULT_APP_PORT = 8080
+DEFAULT_APP_HOST = 'localhost'
```
mickmister, line 13: For ease of configurability, maybe we make this `0.0.0.0`? Also let's make the env var be `HOST`

examples/python/src/hello-world.py
```diff
@@ -0,0 +1,199 @@
...
+DEFAULT_ROOT_URL = 'http://localhost:8080'
+FORM = {
```
mickmister, line 15: Not sure why this is capitalized. Even though it's only defined once, it's not really a "constant"

examples/python/src/hello-world.py
mickmister: Okay so when you say "command will fail to expand the context", it's sort of a silent error, and the app still receives the request. It's just not the full request with all values. I think this is due to new changes to expand (optional vs required), depending on which version of the App framework you're working with.

examples/python/src/hello-world.py
dsharma522: done

examples/python/src/hello-world.py
dsharma522: replaced with lower case variable names

examples/python/README.md
dsharma522: done

examples/python/src/hello-world.py
dsharma522: No actually "command will fail to expand the context" mean is that command will fail in the UI itself and not silent failure. And app won't receive the request. 

Screenshot 2022-07-25 at 10 34 52 PM

examples/python/src/hello-world.py
```diff
@@ -0,0 +1,199 @@
...
+                        'hint': '[This is testing command]',
+                        # this will be the command displayed to user as /second-command
+                        'label': 'second-command',
+                        'icon': 'icon.png',
+                        'bindings': [
+                            {
+                                'location': 'send',
+                                'label': 'send',
+                                'form': form
+                            },
```
mickmister, line 104: Can we have a comment explaining that this is how you make a subcommand `/second-command send`? Mainly because there are already plenty of comments explaining things around this area, and this is an important concept to understand imo. Also maybe a comment mentioning that we are including the form here

mickmister: Also I know I said that `form` shouldn't be capitalized, but maybe to make it stand out more, we can make it say `SHARED_FORM`

mickmister (approved):
LGTM! Just a few more suggestions

neflyte:
Would it be possible to improve the instructions so they include step-by-step guides for getting started? For example, link to python.org for install instructions, describe (or link to) how to install the dependencies in `requirements.txt` (or some other dependency manager like `pipenv`), what changes are necessary in `settings.json.

It would be best to include this information now, before merging, so it isn't missed in the future.

examples/python/src/hello-world.py
dsharma522: Done

mickmister (approved):

PR #353, opened 2022-08-15
fixed the test app

levb:
- appclient no longer requires a userID to work
- `DM[Post]` errors out if the creator's user ID is not known, with a `empty sender user_id, perhaps Call does not expand acting_user` message. Note that the Bot user id is always expanded.
- Fixed the test apps' subscription parameters

@DHaussermann I did not review every expand in the test app, but it does not use `DM` nor `DMPost` so should be OK.

Embedded does not appear to work, will debug it separately, perhaps @mickmister can help?

### DIFF
--- a/apps/appclient/mattermost_client.go
+++ b/apps/appclient/mattermost_client.go
@@ -16,21 +16,26 @@ type Client struct {
 	userID string
 }
 
-func as(id, token string, cc apps.Context) *Client {
-	return NewClient(id, token, cc.MattermostSiteURL)
+func as(token string, cc apps.Context) *Client {
+	return NewClient(token, cc.MattermostSiteURL)
 }
 
 func AsBot(cc apps.Context) *Client {
-	return as(cc.BotUserID, cc.BotAccessToken, cc)
+	client := as(cc.BotAccessToken, cc)
+	client.userID = cc.BotUserID
+	return client
 }
 
 func AsActingUser(cc apps.Context) *Client {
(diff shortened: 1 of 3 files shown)


apps/appclient/mattermost_client.go
```diff
@@ -16,21 +16,26 @@ type Client struct {
...
 func AsBot(cc apps.Context) *Client {
-	return as(cc.BotUserID, cc.BotAccessToken, cc)
+	client := as(cc.BotAccessToken, cc)
+	client.userID = cc.BotUserID
+	return client
 }
 
 func AsActingUser(cc apps.Context) *Client {
-	return as(cc.ActingUser.Id, cc.ActingUserAccessToken, cc)
+	client := as(cc.ActingUserAccessToken, cc)
```
levb, line 30: Deferring the error handling for no user ID until `DMPost`

apps/appclient/mattermost_client.go
```diff
@@ -170,8 +175,6 @@ func (c *Client) Call(creq apps.CallRequest) (*apps.CallResponse, error) {
 }
 
 func (c *Client) CreatePost(post *model.Post) (*model.Post, error) {
-	post.UserId = c.userID
```
mickmister, line 173: :+1: I figured this wasn't required because it's going through the API with token

mickmister (approved):

catalintomai (approved):

PR #354, opened 2022-08-16
Update mattermost-server to 7.2, improve circle-ci caching

levb:
- Updated the server dependency to v7.2
- improved caching and build performance in circleci


.circleci/config.yml
```diff
@@ -39,79 +39,35 @@ commands:
       - aws-s3/copy:
           from: << parameters.filename >>
           to: << parameters.bucket >>
-          arguments: '--acl public-read --cache-control no-cache'
-
-  install-golangci-lint:
```
mickmister, line 44: Why is this no longer necessary?

mickmister (approved):

.circleci/config.yml
```diff
@@ -10,23 +10,23 @@ executors:
...
-      - ./test/e2e/node_modules
-      - /home/circleci/.cache/Cypress
+  - &restore_go_cache
+    restore_cache:
+      key: go-mod-v4-{{ checksum "go.sum" }}
+  - &save_go_cache
+    save_cache:
+      key: go-mod-v4-{{ checksum "go.sum" }}
+      paths:
+        - "~/go/pkg/mod"
```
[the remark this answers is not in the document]
levb: The env variables do not appear to work ubiquitously, this seems an easy (and idiomatic?) reference in the `cimg` world.

.circleci/config.yml
levb: Because `golangci` is already installed? Or did I miss the point of that step?

catalintomai (approved):

.circleci/config.yml
```diff
@@ -10,23 +10,23 @@ executors:
...
+  - &restore_cypress_cache
+    restore_cache:
```
levb, line 22: @mickmister this is not working, see attached, 
image
appears that `package-lock.json` is changed by the build, so the cache is saved for a different sha.

We should fix this - at the very least because we are continuously generating large caches, 0/5 we generate in `master`, refer to the cache by the commited `package-lock.json` sha

PR #359, opened 2022-08-25
Ability to force refresh bindings for app after call

### LINKED ISSUE #358: Is it possible update bindings by app?
I try to create 2 opposite bindings for /channel_header.
When we call one of it some internal state changed and we need to show another one.
When user switch channel from UI - bindings updated, because app check internal state for this channel and send right data. But I want to initiate re-request after internal state changed.
Is it possible?

PSyton:
Add optional bool field to CallResponse forcing refresh bindings for app.
Refreh requested only with success calls.

#### Ticket Link

Fixes https://github.com/mattermost/mattermost-plugin-apps/issues/358


mickmister (approved):

mickmister:
@PSyton Are you able to add a unit test for this? Ideally we have an example usage of this int he `examples` folder for manual testing as well, but I think a unit test will do here.

mickmister:
@levb There was discussion about this here https://github.com/mattermost/mattermost-plugin-apps/issues/358

PSyton:
I'll try to do it.

levb (approved):
🎉 👍

azigler (approved):
> @PSyton Are you able to add a unit test for this? Ideally we have an example usage of this int he `examples` folder for manual testing as well, but I think a unit test will do here.

Agreed @mickmister this could be demonstrated with a unit test. If one is created, I can find a home for it in the examples repo. Looks like this new optional arg for `CallResponse` would get documented [on this page in the docs](https://developers.mattermost.com/integrate/apps/structure/call/#call-response), is that right? cc @neflyte @cwarnermm

mickmister:
@PSyton Please try to avoid force-pushing to the PR. Instead, please push new commits onto the PR so it's easier to see the history of review and work done on the PR. When the PR is merged, we will squash the commits into one commit, so it will be as though there was only one commit for the PR.

mickmister (approved):
Tests LGTM, thanks @PSyton!

levb:
@PSyton @mickmister `TestRESTAPI/notify/user_left_team/expand_all/user2` seems to be flaky, I will investigate separately, going to re-run the test and see if it helps.

hanzei (commented):
Can an app cause an infinite loop my setting `RefreshBindings=true` in the bindings response?

PSyton:
> Can an app cause an infinite loop my setting `RefreshBindings=true` in the bindings response?

Good question, need to check it. And if it's true, I think we can add test for this case and ignore RefreshBindings value when handle bindings response.

levb:
@hanzei I think your concern has been taken care of. I am not 100% about checking the call path, wish we had a better indicator of being in a "getBindings" call.

server/proxy/bindings_test.go
```diff
@@ -461,3 +476,152 @@ func TestCleanAppBinding(t *testing.T) {
...
+				up := mock_upstream.NewMockUpstream(ctrl)
+
+				// set up an empty OK call response
+				b, _ := json.Marshal(tc.callResponse)
+				reader := ioutil.NopCloser(bytes.NewReader(b))
+				up.EXPECT().Roundtrip(gomock.Any(), gomock.Any(), gomock.Any(), gomock.Any()).Return(reader, nil)
+
+				upstreams[app.Manifest.AppID] = up
+				appStore.EXPECT().Get(app.Manifest.AppID).Return(&app, nil)
+			}
```
hanzei, line 604: Not a blocker. I would like to see an restapi test (`test/restapitest`) instead of all of these mocks. That way the test ensures that the websocket event is actually sent to the client.

server/proxy/invoke_call.go
```diff
@@ -149,5 +150,12 @@ func (p *Proxy) callApp(r *incoming.Request, app *apps.App, creq apps.CallReques
 		}
 		cresp.Form = &clean
 	}
+
+	if cresp.Type != apps.CallResponseTypeError &&
+		creq.Call.Path != path.Bindings &&
```
hanzei, line 155: An app can define its own `path` for the bindings call (https://github.com/mattermost/mattermost-plugin-apps/blob/647743f10b5309af3f1b719c9b12ec19c92c1ea2/apps/manifest.go#L75-L78). That also needs to get checked.

hanzei (changes requested):

hanzei (approved):
LGTM :+1:

levb:
Added a draft doc PR, cc @asatkinson

PR #363, opened 2022-09-06
Add developer route to refresh bindings on demand

mickmister:
#### Summary

This PR adds an endpoint `/api/v1/refresh_bindings` that will send a `refresh_bindings` websocket message to a given user.

This is especially useful during a rapid development cycle. As soon as my App is redeployed locally after some change, the App's start-up sequence can ping the Apps plugin to refresh the client's bindings.

This criteria is required for this to occur:

- `EnableDeveloper` is enabled
- The request is coming from an App
- A user id is provided in url params

#### Ticket Link

Fixes https://mattermost.atlassian.net/browse/MM-46901

### DIFF
--- a/server/appservices/dev.go
+++ b/server/appservices/dev.go
@@ -1,0 +1,17 @@
+package appservices
+
+import (
+	"github.com/mattermost/mattermost-plugin-apps/server/config"
+	"github.com/mattermost/mattermost-plugin-apps/server/incoming"
+	"github.com/mattermost/mattermost-server/v6/model"
+)
+
+func (a *AppServices) RefreshBindings(r *incoming.Request, userID string) error {
+	err := r.RequireSourceApp()
+	if err != nil {
+		return err
+	}
+
+	r.Config().MattermostAPI().Frontend.PublishWebSocketEvent(config.WebSocketEventRefreshBindings, map[string]interface{}{}, &model.WebsocketBroadcast{UserId: userID})
+	return err
+}
--- a/apps/appclient/mattermost_client_pp.go
+++ b/apps/appclient/mattermost_client_pp.go
@@ -148,6 +148,16 @@ func (c *ClientPP) Unsubscribe(sub *apps.Subscription) (*model.Response, error) 
 	return model.BuildResponse(r), nil
 }
 
+func (c *ClientPP) RefreshBindings(userID string) (*model.Response, error) {
+	r, err := c.DoAPIPOST(c.apipath(appspath.RefreshBindings + "?user_id=" + userID), "") // nolint:bodyclose
+	if err != nil {
+		return model.BuildResponse(r), err
+	}
+	defer c.closeBody(r)
+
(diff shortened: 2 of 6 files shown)


mickmister:
This PR will allow for quicker dev cycles during App development. Please take a look when you have the chance, it should be a quick review

server/appservices/dev.go
```diff
@@ -0,0 +1,17 @@
+package appservices
+
+import (
+	"github.com/mattermost/mattermost-plugin-apps/server/config"
+	"github.com/mattermost/mattermost-plugin-apps/server/incoming"
+	"github.com/mattermost/mattermost-server/v6/model"
+)
+
+func (a *AppServices) RefreshBindings(r *incoming.Request, userID string) error {
+	err := r.RequireSourceApp()
```
hanzei, line 10: Given that a app specific session is required the app needs to persist a token during restart. How would that be accomplished in a secure and easy way?

mickmister: @hanzei That's a good point. I figured the developer would capture it themselves and save it locally for use. There should be a more straightforward way to use this though. Do you see something that can change here to make this more accessible?

levb: Anything that is "dev" can be scripted as part of the apps makefile, and use a sysadmin token + `curl`; 2/5 persisting app session tokens should be strongly discouraged.

levb: Also, I don't understand the value of refreshing a specific user's bindings on an app restart, shouldn't all users be refreshed?

mickmister: > Also, I don't understand the value of refreshing a specific user's bindings on an app restart, shouldn't all users be refreshed?

@levb This is meant to limit the scope of how many binding requests come in when this dev route is hit. In a "production" server running with dev mode on (should we even support this case?), this route could be abused to queue up a large number of requests from all users simultaneously, and repetitively. If we don't want to be concerned with this, we can instead have the endpoint have all users refresh.

hanzei: > Do you see something that can change here to make this more accessible?

Does it make sense to have an API endpoint that a user can hit to force a bindings request for themself? Then we can hit that endpoint as part of re-running the app.

mickmister: @hanzei That sounds good, but how does the App end up sending this request? Does it need a user token for this?

We could use a personal access token approach for the development side of this. The developer would create a personal access token for the user that they're using for testing, and use the token on App startup, to trigger the binding refresh for that specific user.

hanzei: Yes, a developer would use a personal access token.

To be clear: My idea was around triggering the "refresh bindings" endpoint as part of the build process, e.g. as part of the `Makefile`. I'm hesitant to put that code into the app as it's more of a "development detail".

mickmister: @hanzei This idea makes a lot of sense. Thanks for suggesting this!

PR #362, opened 2022-08-31
 Move /examples and /dev to new mattermost-app-examples repository

azigler:
_New version of #357 based on #336_ 
 
**This PR should not be merged until https://github.com/mattermost/mattermost-developer-documentation/pull/1106 is merged.**

Fixes:
- https://github.com/mattermost/mattermost-plugin-apps/issues/330
- https://github.com/mattermost/mattermost-plugin-apps/issues/329
- https://github.com/mattermost/mattermost-plugin-apps/issues/300

Updates:
- `README` points to developer resources and guides

This PR moves:
- the content from `/examples` to the new https://github.com/mattermost/mattermost-app-examples repository.
- the `/dev` environment to the new https://github.com/mattermost/mattermost-app-examples repository
- the new content from merged PR https://github.com/mattermost/mattermost-plugin-apps/pull/352 to correspond with the new Python quick start guide in https://github.com/mattermost/mattermost-developer-documentation/pull/1106.

`/examples/goapp` has been preserved in this repository as `/goapp`.

PR #370, opened 2022-10-14
Fix e2e tests for webapp master branch

mickmister:
#### Summary

This PR fixes the issue where the e2e test would get stuck on the webapp's landing page to open the desktop app.

As a follow up to https://github.com/mattermost/mattermost-plugin-apps/pull/362#issuecomment-1279257314, I tried using the test app rather than the hello app for the e2e tests, but the test app currently doesn't test things like sending bot DMs, so it's currently not available to be a drop-in replacement.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -445,7 +445,7 @@ workflows:
       - e2e-cypress-tests-master:
           filters:
             branches:
-              only: master
+              only: /^v.*/
           requires:
             - lint
       - test-rest-postgres11:
--- a/test/e2e/cypress/support/index.ts
+++ b/test/e2e/cypress/support/index.ts
@@ -6,3 +6,7 @@
 // ***********************************************************
 
 import 'mattermost-webapp/e2e/cypress/tests/support';
+
+before(() => {
+    localStorage.setItem('__landingPageSeen__', String(true));
+});


.circleci/config.yml
```diff
@@ -445,7 +445,7 @@ workflows:
       - e2e-cypress-tests-master:
           filters:
             branches:
-              only: master
+              only: /^v.*/
```
mickmister, line 448: Temporarily setting this to make this PR's tests run with webapp master. Maybe this should be set this anyway?

test/e2e/cypress/support/index.ts
```diff
@@ -6,3 +6,7 @@
 // ***********************************************************
 
 import 'mattermost-webapp/e2e/cypress/tests/support';
+
+before(() => {
+    localStorage.setItem('__landingPageSeen__', String(true));
```
mickmister, line 11: Here's the landing page fix. The webapp checks for this value and avoids showing the landing page screen if it's set.

.circleci/config.yml
hanzei: One reason that I don't want to run webapp `master` against all branches is that I don't want to break PRs accidently..

hanzei (approved):
Thanks for the fix!

mickmister:
@levb The changes in this PR are pretty small and straightforward. Please take a quick look when you get the chance :+1:

levb (approved):

.circleci/config.yml
hanzei: Maybe I'm missing something here. This change makes it so that the e2e tests are only run against `v*` branches, I.e. release branches of this repo. Why should it be this way?

PR #372, opened 2022-10-26
[MM-47950] Allow setting developer mode via /apps settings

hanzei:
#### Summary
By default, developer mode is enabled when `Enable Developer Mode` is enabled in the system console and HTTP apps are allowed when it's an on-prem server or developer mode is enabled.

Screenshot from 2022-10-26 13-36-11
Screenshot from 2022-10-26 13-36-25

#### Ticket Link
Fixes https://mattermost.atlassian.net/browse/MM-47950


server/proxy/install.go
```diff
@@ -22,7 +22,7 @@ import (
 )
 
 // InstallApp installs an App.
-//  - cc is the Context that will be passed down to the App's OnInstall callback.
+//   - cc is the Context that will be passed down to the App's OnInstall callback.
```
javaguirre, line 25: There are no changes here, but it's just the autoformat that triggered this change probably :-)

server/config/config.go
```diff
@@ -39,6 +38,24 @@ type StoredConfig struct {
...
+func unmarshalPluginConfig(pluginConfig map[string]any, mmconf *model.Config, mattermostCloudMode bool) StoredConfig {
+	sc := StoredConfig{}
+	utils.Remarshal(sc, pluginConfig)
+
+	if _, ok := pluginConfig["developer_mode"]; !ok {
+		sc.DeveloperMode = pluginapi.IsConfiguredForDevelopment(mmconf)
+	}
+
+	if _, ok := pluginConfig["allow_http_apps"]; !ok {
+		sc.AllowHTTPApps = sc.DeveloperMode || !mattermostCloudMode
```
javaguirre, line 55: Shouldn't this be and `&&`? If developermode is on and the mattermostcloud is not active?

javaguirre (commented):
Nice! Only one question, other than that LGTM!

server/proxy/install.go
hanzei: Yes, that is due to the go 1.19 doc updates (https://tip.golang.org/doc/go1.19)

server/config/config.go
hanzei: The logic works as intended. For on-prem servers we want to allow HTTP apps by default.

javaguirre (approved):

hanzei:
Heads up that CI is still failing for unknown reasons. Need to look into it.

asatkinson (approved):
This fits well with PR "/apps debug logs to stream plugin logs to a Mattermost channel #394". We should try to put these both out in December 1.2 Release if possible

Nitpick: The configuration dialog says "Enable Developers Mode". I believe it should say "Enable Developer Mode"

server/builtin/debug_clean.go
```diff
@@ -37,7 +36,10 @@ func (a *builtinApp) debugClean(r *incoming.Request, creq apps.CallRequest) apps
 		Other: "Deleted all KV records.",
 	}) + "\n"
 
-	err = a.conf.StoreConfig(config.StoredConfig{}, r.Log)
+	sc := a.conf.Get().StoredConfig
```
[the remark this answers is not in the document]
hanzei: Because that would also wipe out the configuration settings which were introduced in this PR without setting the default value. I'm not sure that setting everything to `false` is the desired behaviour here. Maybe every setting should be set to the default value?

hanzei:
@asatkinson Good catch, I've fixed that with [`638a791` (#372)](https://github.com/mattermost/mattermost-plugin-apps/pull/372/commits/638a79175390cc844cf15553995448d642ba1517).

hanzei:
@DHaussermann The PR is ready for your review.

hanzei:
@javaguirre Thanks for the review. :+1: I've simplified the confirmation message. It does not output the values any longer.
Screenshot from 2022-12-05 12-59-48

javaguirre (approved):
Tested again and everything works great, thank you for the fast response @hanzei 💯

hanzei:
Documentation PR: https://github.com/mattermost/mattermost-developer-documentation/pull/1211

PR #374, opened 2022-10-28
Update golangci-lint version and fix linter errors

hanzei:
#### Summary
This PR updates golangci-lint to the version that other plugins use and fixate that version for CI.

It also fixes the newly found linter issues. A lot of them are formatting one due to the go1.18 doc updates.

#### Ticket Link
None

### DIFF
--- a/server/httpin/admin.go
+++ b/server/httpin/admin.go
@@ -15,17 +15,22 @@ import (
 // manifest store, making the App installable. The resulting listed manifest
 // will combine the deployment information from the prior listing, and the new
 // manifests as follows:
-//   1. The "core" manifest (except Deploy) is updated to the new values.
-//   2. Deploy types from the previously listed manifest are updated from the new manifest, or preserved.
-//   3. Deploy types specified in "add_deploys" are copied from the new manifest.
-//   4. "remove"
-//   Path: /api/v1/add-listed-app
-//   Method: POST
-//   Input: JSON{
-//      Manifest...
-//      "add_deploys": []string e.g. ["aws_lambda","http"]
-//      "remove_deploys": []string e.g. ["aws_lambda","http"]
-//   Output: The updated listing manifest
+//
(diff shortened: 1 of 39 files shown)


test/hello-world/hello.go
```diff
@@ -131,9 +131,15 @@ func main() {
...
+	server := &http.Server{
+		Addr:              addr,
+		ReadHeaderTimeout: 3 * time.Second,
+		WriteTimeout:      5 * time.Second,
+	}
+
 	fmt.Println("Listening on", addr)
 	fmt.Println("Use '/apps install http http://localhost" + addr + "/manifest.json' to install the app") // matches manifest.json
-	log.Fatal(http.ListenAndServe(addr, nil))
+	panic(server.ListenAndServe())
```
hanzei, line 142: I went with `panic` as  that is was the other apps are also using. 0/5

javaguirre (approved):
LGTM! 👏

server/httpin/admin.go
```diff
@@ -15,17 +15,22 @@ import (
...
-//      "add_deploys": []string e.g. ["aws_lambda","http"]
-//      "remove_deploys": []string e.g. ["aws_lambda","http"]
-//   Output: The updated listing manifest
+//
+//  1. The "core" manifest (except Deploy) is updated to the new values.
+//  2. Deploy types from the previously listed manifest are updated from the new manifest, or preserved.
+//  3. Deploy types specified in "add_deploys" are copied from the new manifest.
+//  4. "remove"
+//
+// _
```
levb, line 24: What does this do?

levb (approved):

server/httpin/admin.go
hanzei: It separates the code block from the numbered list. Without the character the two blocks get combined, and the code block becomes part of the 4. item.

PR #375, opened 2022-10-28
[MM-48342] Stateful timers

hanzei:
#### Summary
This is a first draft of state full timer. Apps can now create timers by defining a time when the timer should be executed.

The API is:
```golang
type Timer struct {
	// At is the unix time in milliseconds when the timer should be executed.
	At int64 `json:"at"`

	// Call is the (one-way) call to make upon the timers execution.
	Call Call `json:"call"`

	// ChannelID is a channel ID that is used for expansion of the Call (optional).
	ChannelID string `json:"channel_id,omitempty"`
	// TeamID is a team ID that is used for expansion of the Call (optional).
	TeamID string `json:"team_id,omitempty"`
}
```
#### Ticket Link
https://mattermost.atlassian.net/browse/MM-48342


hanzei:
@levb @javaguirre The details are still WIP, but I would appreciate a high level review of the API.

levb:
@hanzei would you please publish your `mattermost-plugin-api` branch? I did not find it.

server/httpin/call.go
```diff
@@ -11,11 +11,29 @@ import (
 	"github.com/mattermost/mattermost-plugin-apps/utils/httputils"
 )
 
+// CallResponse contains everything the CallResponse struct contains, plus some additional
+// data for the client, such as information about the App's bot account.
+//
+// Apps will use the CallResponse struct to respond to a CallRequest, and the proxy will
+// decorate the response using the CallResponse to provide additional information.
+type CallResponse struct {
```
levb, line 19: Is it (too) difficult to isolate the `CallResponse` changes into its own PR?

server/httpin/call.go
```diff
@@ -11,11 +11,29 @@ import (
 	"github.com/mattermost/mattermost-plugin-apps/utils/httputils"
 )
 
+// CallResponse contains everything the CallResponse struct contains, plus some additional
```
levb, line 14: 👍

go.mod
```diff
@@ -201,3 +201,5 @@ require (
 	gopkg.in/natefinch/lumberjack.v2 v2.0.0 // indirect
 	gopkg.in/yaml.v2 v2.4.0 // indirect
 )
+
+replace github.com/mattermost/mattermost-plugin-api => /home/bschumacher/src/mattermost/api
```
levb, line 205: publish the branch, please and update go.mod?

levb (changes requested):

javaguirre (commented):
It would be nice if you add a code example in a comment of how we plan to use this to better understand everything. It's nice also for the future documentation. Thank you! 👍

go.mod
hanzei: Fixed in a8ce6db215e0bd5b0f454e797f7117dab3903d37

hanzei:
@javaguirre The test app contains an example usage of timers.

server/httpin/call.go
hanzei: That is reasonably doable without too much effort: https://github.com/mattermost/mattermost-plugin-apps/pull/381

levb:
@hanzei 0/5 and optional: would be nice to add a REST API test for this, at least for the happy path? If you need help figuring the tests out, I'd be happy to.

javaguirre (commented):
I've made some sweeps on the code, but I need to spend more time on it, just wondering about your idea to change how `setCaller` works.

server/appservices/service.go
```diff
@@ -33,14 +42,45 @@ type Service interface {
...
 }
 
 var _ Service = (*AppServices)(nil)
 
-func NewService(store *store.Service) *AppServices {
-	return &AppServices{
-		store: store,
+// SetCaller must be called before calling any other methods of AppsServies.
+// TODO: Remove this uggly hack.
+func (a *AppServices) SetCaller(caller Caller) {
```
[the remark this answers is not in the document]
hanzei: What would that default `Caller` look like?

server/appservices/service.go
```diff
@@ -33,14 +43,45 @@ type Service interface {
...
 
+type Caller interface {
+	InvokeCall(*incoming.Request, apps.CallRequest) (*apps.App, apps.CallResponse)
+	NewIncomingRequest() *incoming.Request
+}
+
 type AppServices struct {
-	store *store.Service
+	store     *store.Service
+	scheduler *cluster.JobOnceScheduler
```
levb, line 53: Here's my biggest issue with this PR. JobOnceScheduler relies on `KVList`. plugin-apps stores all of the app keys in its keyspace (session and ouath2 tokens), so the list is going to be huge and `KVList` slow.

2/5 consider https://github.com/mattermost/mattermost-server/pull/21653

levb (changes requested):

server/appservices/service.go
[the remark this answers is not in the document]
javaguirre: I think the problem it's not creating a default for Caller, but AppServices doing too much, why not having different services providing their own business logic? I don't think a subscription app service would need any config to create a timer, right?

I need to think more about it, but if you want we can talk so I can be more helpful.

server/appservices/service.go
hanzei: For reference: `cluster.JobOnceScheduler` calls `KVList` on `Start()` and every 5 minutes.

hanzei: Could you please elaborate on how a solution that uses https://github.com/mattermost/mattermost-server/pull/21653 would work?

I would assume only the cluster leader executes timers, correct? Once a plugin instance received the request to create a new timer, it needs to communicate that information to the leader. How would that work?

server/appservices/service.go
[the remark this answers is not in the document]
javaguirre: This is what I meant, let me know what you think and why/why not do this (I believe a decision was made before this PR, so I'm interested) :-)

```go
// Copyright (c) 2020-present Mattermost, Inc. All Rights Reserved.
// See License for license information.

package appservices

import (
	"github.com/pkg/errors"

	"github.com/mattermost/mattermost-plugin-api/cluster"

	"github.com/mattermost/mattermost-plugin-apps/server/config"
	"github.com/mattermost/mattermost-plugin-apps/server/incoming"
	"github.com/mattermost/mattermost-plugin-apps/server/store"
	"github.com/mattermost/mattermost-plugin-apps/utils"
)

type AppSubscriptionService interface {
	// Subscriptions

	Subscribe(*incoming.Request, apps.Subscription) error
	GetSubscriptions(*incoming.Request) ([]apps.Subscription, error)
	Unsubscribe(*incoming.Request, apps.Event) error
	UnsubscribeApp(*incoming.Request, apps.AppID) error
}

type AppSubscriptionService interface {
	// Timer

	CreateTimer(*incoming.Request, apps.Timer) error
}

type AppKVService interface {
    // KV

	KVSet(_ *incoming.Request, prefix, id string, data []byte) (bool, error)
	KVGet(_ *incoming.Request, prefix, id string) ([]byte, error)
	KVDelete(_ *incoming.Request, prefix, id string) error
	KVList(_ *incoming.Request, namespace string, processf func(key string) error) error
	KVDebugInfo(*incoming.Request) (*store.KVDebugInfo, error)
	KVDebugAppInfo(*incoming.Request, apps.AppID) (*store.KVDebugAppInfo, error)


}
	
type AppOAuthService interface {
    	// Remote (3rd party) OAuth2
	StoreOAuth2App(_ *incoming.Request, data []byte) error
	StoreOAuth2User(_ *incoming.Request, data []byte) error
	GetOAuth2User(_ *incoming.Request) ([]byte, error)
}

type Caller interface {
	InvokeCall(*incoming.Request, apps.CallRequest) (*apps.App, apps.CallResponse)
	NewIncomingRequest() *incoming.Request
}

type AppService struct {
	store     *store.Service


	conf config.Service
	log  utils.Logger
}

type AppTimerService struct {
    service *AppService
    
    scheduler *cluster.JobOnceScheduler
	caller    Caller
}

type AppSubscriptionService struct {
    service *AppService
}
```

hanzei:
Notes from an offline discussion with @levb:

- Notes receiving a job create event communicate to the leader the event
- Leader writen to KV store
- Also store a list of keys per interval (e.g. per hour)
- Leader keeps a list of the jobs for the current interval in memory
- The leader is the only one who executes

On leader change:
- Read the current interval from KV store

levb (approved):

server/appservices/service.go
hanzei: I've talked with @levb offline. The solutions in https://github.com/mattermost/mattermost-plugin-apps/pull/375#issuecomment-1410445150 are the path forward, but require the cache store from https://github.com/mattermost/mattermost-plugin-apps/pull/458.

hanzei:
Let's merge this and reiterate on store once https://github.com/mattermost/mattermost-plugin-apps/pull/458 is merged.

PR #376, opened 2022-10-31
Add webhook secret in context when expanding `app`

dsharma522:
#### Summary
Add webhook secret when requested for `app` expansion

In latest release [mattermost-plugin-apps-345 (1.1.0)](https://github.com/mattermost/mattermost-plugin-apps/pull/345), there was webhook_secret included in the expanded section of `app` in `context` on `bot_joined_team event`, but now it is not there

### DIFF
--- a/server/proxy/expand.go
+++ b/server/proxy/expand.go
@@ -249,7 +249,6 @@ func (e *expander) expandUser(userPtr **model.User, userID string) expandFunc {
 func (e *expander) expandApp(level apps.ExpandLevel) error {
 	e.ExpandedContext.App = e.app.Strip(level)
 
-	e.ExpandedContext.App.WebhookSecret = ""
 	if level == apps.ExpandAll && e.r.RequireSysadminOrPlugin() == nil {
 		e.ExpandedContext.App.WebhookSecret = e.app.WebhookSecret
 	}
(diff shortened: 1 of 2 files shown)


server/proxy/expand.go
```diff
@@ -249,7 +249,6 @@ func (e *expander) expandUser(userPtr **model.User, userID string) expandFunc {
 func (e *expander) expandApp(level apps.ExpandLevel) error {
 	e.ExpandedContext.App = e.app.Strip(level)
 
-	e.ExpandedContext.App.WebhookSecret = ""
 	if level == apps.ExpandAll && e.r.RequireSysadminOrPlugin() == nil {
 		e.ExpandedContext.App.WebhookSecret = e.app.WebhookSecret
```
hanzei, line 253: Can you please elaborate why this current logic is wrong? The webhook secret should only get send to system admins.

hanzei (changes requested):

server/proxy/expand.go
dsharma522: Noted, but when expanding `app:all`, shouldn't that too get webhook secret ?

dsharma522: Also it seems to working with the changes in this PR, I could not fix tests for it currently, but tested manually

hanzei: If a non system admin expands `app:all` they should not get the webhook secrets. Only admins should get it.

dsharma522: Got it. Thank you.
I will try to look into it again!

hanzei:
Closin as invalid

PR #377, opened 2022-11-01
[MM-48054] Don't disable app before uninstalling it

hanzei:
#### Summary
I don't know why the code was changed to disable the app before uninstalling it. If the app is disabled, any `OnUninstall` call will fail.

This PR reverts the changes and moves back to the existing approach of simply removing the session store.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-48054


levb (approved):

hanzei:
Note to self: look into writing an restapi test

javaguirre (approved):
Not a lot of context here, but seems fair 👍

hanzei:
@javaguirre Is there any context that I could explain for you?

javaguirre:
> @javaguirre Is there any context that I could explain for you?

It's ok! I think my context is good enough to understand why this change is relevant, I'll get more once I have time to create my own app. Thank you Ben! :-)

levb (changes requested):
So I was adding the token cleanup, and realized why I added Disable - see the messages below, "failed to delete Mattermost OAuth2 for %s, the app is left disabled".

We ought to change this so 

```go
defer func() {
  if err != nil && appID != "" {
    p.DisableApp(appID)
  }
}
```
Otherwise on error the app is left in an inconsistent state. Alternatively, just move the Disable until after OnUinstall

levb:
https://github.com/mattermost/mattermost-plugin-apps/pull/380 moves Disabling to after the app is notified.

hanzei:
@levb 2/5 that we should merge this PR to fix the most imminent bug (uninstall doesn't work) and we move the discussion about if/how an app should get disabled on uninstall to https://github.com/mattermost/mattermost-plugin-apps/pull/380.

levb (approved):

levb:
Whoops, sorry @DHaussermann skipped QA; we should have a test for this anyway, but agree with @hanzei let's do it separately.

PR #394, opened 2022-11-19
[MM-48538] `/apps debug logs` to stream plugin logs to a Mattermost channel

levb:
#### Summary
Added the ability to stream all `mattermost-plugin-apps` logs to a channel. Use `/apps debug logs [--create-channel true | --channel <channel>] [--level] [--json]` to control.

The "core" change: https://github.com/mattermost/mattermost-plugin-apps/pull/394/files#diff-9c4561950e563d1167e9e955b5ab4c6ce45184275371b672a33bc16b1ccdcfb9R96

Looks like this, 
image

Also, some cleanup, including unused mocks.

#### Ticket Link
[MM-48538](https://mattermost.atlassian.net/browse/MM-48538)


server/proxy/service.go
```diff
@@ -44,7 +44,7 @@ type Proxy struct {
 	// expandClientOverride is set by the tests to use the mock client
 	expandClientOverride mmclient.Client
 
-	log utils.Logger
+	// log utils.Logger
```
javaguirre, line 47: should we remove this?

server/proxy/service.go
```diff
@@ -121,7 +121,7 @@ func NewService(conf config.Service, store *store.Service, mutex *cluster.Mutex,
 		httpOut:          httpOut,
 		sessionService:   session,
 		appservices:      appservices,
-		log:              log,
+		// log:              conf.Get().BaseLogger,
```
javaguirre, line 124: Idem

server/proxy/service.go
javaguirre: Or do you plan to add logs here somewhere?

javaguirre (approved):
LGTM, just commented out details

asatkinson:
This is awesome.  Here is a JIRA ticket in the developer experience EPIC
https://mattermost.atlassian.net/browse/MM-48538?atlOrigin=eyJpIjoiOTBlNDFiYjJiOTk5NDI1NjgzZGI4NmE1NjY3OWQxNGQiLCJwIjoiaiJ9

asatkinson (approved):
This is awesome. I created a JIRA ticket in the developer experience EPIC.  We can move it if desired.
https://mattermost.atlassian.net/browse/MM-48538?atlOrigin=eyJpIjoiOTBlNDFiYjJiOTk5NDI1NjgzZGI4NmE1NjY3OWQxNGQiLCJwIjoiaiJ9

- This seems like a lot of changes.  Please make sure you have good review on this.  If this PR slips due to other priorities that is OK, but it would be nice to have.
- Let me know if you need me to queue up documentation for this.  We can ship this undocumented for a while if need be.

mickmister (approved):

server/config/service.go
```diff
@@ -107,37 +220,26 @@ func (s *service) reloadMattermostConfig() *model.Config {
...
-	license := s.mm.System.GetLicense()
-	if license == nil {
-		license = s.mm.System.GetLicense()
-		if license == nil {
-			log.Debugf("failed to fetch license twice. May incorrectly default to on-prem mode")
-		}
+func (s *service) Reconfigure(newStoredConfig StoredConfig, quiet bool, services ...Configurable) error {
+	var configLogger utils.Logger = utils.NilLogger{}
+	if !quiet {
+		configLogger = s.NewBaseLogger()
```
hanzei, line 226: When this line gets called, `s.mattermostConfig` has not been refreshed. This means a change in the dev mode does not get picked up. Also, the logger used in `newInitializedConfig` will not use the updated `LogConfig`.

server/builtin/debug_logs.go
```diff
@@ -0,0 +1,230 @@
...
+						ID:    "field.create.channel.label",
+						Other: "create-channel",
+					}),
+				},
+				{
+					Name: fLevel,
+					Type: apps.FieldTypeStaticSelect,
+					Description: a.conf.I18N().LocalizeDefaultMessage(loc, &i18n.Message{
+						ID:    "field.log.level.description",
+						Other: "Create a new channel for the plugin logs. Use --channel to specify an existing channel.",
```
hanzei, line 64: This seems off.

server/builtin/debug_logs.go
```diff
@@ -0,0 +1,230 @@
...
+				return apps.NewErrorResponse(errors.Wrap(err, "failed to create channel"))
+			}
+		}
+
+		_, err := a.conf.MattermostAPI().Channel.AddMember(ch.Id, creq.Context.ActingUser.Id)
+		if err != nil {
+			return apps.NewErrorResponse(errors.Wrap(err, "failed to add user to channel"))
+		}
+		channelID = ch.Id
+		channelLabel = ch.DisplayName
```
hanzei, line 164: If we use `ch.Name` here, we get an autolink.

server/httpin/service.go
```diff
@@ -36,12 +34,11 @@ var _ http.Handler = (*Service)(nil)
 
 type handlerFunc func(*incoming.Request, http.ResponseWriter, *http.Request)
 
-func NewService(proxy proxy.Service, appservices appservices.Service, conf config.Service, log utils.Logger) *Service {
+func NewService(proxy proxy.Service, appservices appservices.Service, conf config.Service) *Service {
 	rootHandler := &Service{
 		AppServices: appservices,
 		Config:      conf,
 		Proxy:       proxy,
-		baseLog:     log,
```
hanzei, line 44: Good cleanup

server/plugin.go
```diff
@@ -55,7 +53,6 @@ func NewPlugin(pluginManifest model.Manifest) *Plugin {
 
 func (p *Plugin) OnActivate() (err error) {
 	mm := pluginapi.NewClient(p.API, p.Driver)
-	p.log = utils.NewPluginLogger(mm)
```
hanzei, line 58: In order to reduce the number of different loggers used, should a `PluginLogger` be used here?
```golang
log := utils.NewPluginLogger(mm, nil)
```

Once `p.conf` is setup, `log` gets reassigned `log = p.conf.NewBaseLogger()`

server/proxy/synchronize.go
```diff
@@ -24,8 +23,7 @@ func (p *Proxy) SynchronizeInstalledApps() error {
 	ctx, cancel := context.WithTimeout(context.Background(), config.RequestTimeout)
 	defer cancel()
 
-	mm := p.conf.MattermostAPI()
-	r := incoming.NewRequest(p.conf, utils.NewPluginLogger(mm), p.sessionService).WithCtx(ctx)
+	r := incoming.NewRequest(p.conf, p.sessionService).WithCtx(ctx)
```
hanzei, line 26:
```suggestion
	r := p.NewIncomingRequest().WithCtx(ctx)
```

server/proxy/synchronize.go
```diff
@@ -24,8 +23,7 @@ func (p *Proxy) SynchronizeInstalledApps() error {
 	ctx, cancel := context.WithTimeout(context.Background(), config.RequestTimeout)
 	defer cancel()
 
-	mm := p.conf.MattermostAPI()
-	r := incoming.NewRequest(p.conf, utils.NewPluginLogger(mm), p.sessionService).WithCtx(ctx)
```
hanzei, line 28: Should we pass a `Logger` to `SynchronizeInstalledApps`?

server/config/service.go
levb: `configLogger` is the logger to use while handling the configuration change itself, so it is appropriate that it uses the "old" configuration for itself. Also, FWIW, it is set to a "real" value only when called the first time in `OnActivate`, afterwards it's always `nil`.

All other config changes should be picked up by `newInitializedConfig`, or am I missing something else?

server/proxy/synchronize.go
levb: 1/5 it's fine as is, since it creates a new request with its own logger and uses it. The old explicit log initialization is already gone.

utils/plugin_logger.go
```diff
@@ -4,27 +4,43 @@
...
-		plog:         client.Log,
-	}
+func NewPluginLogger(mmapi *pluginapi.Client, confGetter LogConfigGetter) Logger {
 	return &logger{
-		SugaredLogger: zap.New(pc).Sugar(),
+		SugaredLogger: zap.New(&plugin{
+			mm:           mmapi,
+			LevelEnabler: zapcore.DebugLevel,
+			confGetter:   confGetter,
+		}).Sugar(),
```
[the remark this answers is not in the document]
levb: 1/5 There are lots of improvements as to _what_ we actually log, would prefer a separate PR.

utils/plugin_logger.go
```diff
@@ -77,5 +93,36 @@ func (p *plugin) Write(e zapcore.Entry, fields []zapcore.Field) error {
...
+
+	if logconf.IncludeJSON {
+		ccJSON := map[string]any{}
+		for i := 0; i < len(pairs); i += 2 {
+			ccJSON[pairs[i].(string)] = pairs[i+1]
+		}
+		if len(ccJSON) > 0 {
+			message += JSONBlock(ccJSON)
+		}
+	}
```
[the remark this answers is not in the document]
levb: 1/5 we should have another PR that ensures "minimum viable text" in all log messages, and meaningful JSON improvements; not here.

utils/plugin_logger.go
```diff
@@ -4,27 +4,43 @@
...
+	GetLogConfig() LogConfig
+}
+
 // plugin implements zapcore.Core interface to serve as a logging "backend" for
 // SugaredLogger, using the plugin API log methods.
 type plugin struct {
 	zapcore.LevelEnabler
-	plog   pluginapi.LogService
-	fields map[string]zapcore.Field
+	mm         *pluginapi.Client
```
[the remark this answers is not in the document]
levb: Updated; is this what you had in mind?

levb:
Added a draft docs PR for @asatkinson to polish.

javaguirre:
Screenshot 2022-12-05 at 12 40 16
Screenshot 2022-12-05 at 12 40 03

Not sure if I'm doing something wrong, but the command doesn't appear, I'm gonna try to use this along with the `apps settings` PR in case the developer mode is not set up correctly.

javaguirre:
It seems the QA is done for this one though, moving to another one.

server/config/service.go
hanzei: By changing the code to pass a `Logger` to `Reconfigure` the bug has been fixed. :+1:

javaguirre:
Need to test with other cases, but overall it's working well. I can use it, the channel configuration went well and I receive logs.

Screenshot 2022-12-05 at 13 15 02

neflyte:
What happens if I turn on logging with `--create-channel` twice in a row? Will I get two channels that have the same name? If I have two Apps and I turn on this debugging for both, do I also get two channels with the same name? If so, which channel maps to which App?

neflyte:
Also, how do I disable the logging once it's enabled? In a production scenario, I'd want to be able to turn it on, get the logs i need, and turn it back off.

hanzei:
@levb Heads up that there are merge conflicts to resolve

javaguirre:
> What happens if I turn on logging with `--create-channel` twice in a row? Will I get two channels that have the same name? If I have two Apps and I turn on this debugging for both, do I also get two channels with the same name? If so, which channel maps to which App?
> 

Now It only creates one channel and all apps write to that channel if logs are enabled. If you post `--create-channel` twice, it's the same channel and it's not created twice. 

> Perhaps change `--create-channel` to accept a channel name? A channel would be created if the parameter is non-empty.

You could create a channel manually and then use `--channel` to select it.

0/5 I don't think selecting a channel for creation is really needed, you can do it in two steps.

javaguirre:
> Also, how do I disable the logging once it's enabled? In a production scenario, I'd want to be able to turn it on, get the logs i need, and turn it back off.

Screenshot 2022-12-07 at 09 50 13

In order to use the Apps logging, you need to set `Developer mode` and `Testing Commands` to true. When you go back to Production mode we set these to false, and the logging will stop.

utils/plugin_logger.go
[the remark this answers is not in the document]
hanzei: Sure, submitted https://github.com/mattermost/mattermost-plugin-apps/pull/410

utils/plugin_logger.go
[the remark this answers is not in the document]
hanzei: Could you elaborate a bit on your idea? Most of our log messages are quite generic. The value comes from the context i.e. the fields.

hanzei:
@levb Heads up that I merged `master` into your branch. Please double check [`83d1308` (#394)](https://github.com/mattermost/mattermost-plugin-apps/pull/394/commits/83d13085a28c458ca07a3d91bd8ad78c04aff5e5).

hanzei (approved):

hanzei:
@javaguirre @DHaussermann The PR is ready for testing.

javaguirre:
I will test the following cases:

- [ ] Activate logs creating a new channel
- [ ] Activate logs selecting an existing channel
- [ ] Receive logs on a channel without the `--json` flag.
- [ ] Receive logs on a channel with the `--json` flag.
- [ ] Deactivating developer mode and/or testing commands with stop logging
- [ ] If we create a channel and it was created, it's not created twice

javaguirre:
If I select a team and I activate logs for an existing channel within that team, the logs are shown in a different team (the last one created probably if the channel has the same name) @levb @hanzei 

If I use `--create-channel` on a team channel, the logs are shown in the channel `DEBUG: Apps` of that same team as intended.

https://user-images.githubusercontent.com/488556/207268974-160f19af-8ed7-482f-a053-1c7266ca7261.mov

javaguirre (changes requested):
We need to fix this use case

https://github.com/mattermost/mattermost-plugin-apps/pull/394#issuecomment-1347948166

hanzei:
Taking a look, @javaguirre

levb:
@hanzei @javaguirre heads up: I am changing this a bit to combine with the Settings command.

hanzei:
Good catch, @javaguirre. :+1: That is indeed a bug in the framework: https://mattermost.atlassian.net/browse/MM-49043.

Given that  @levb is migrating the command into the settings form, it won't block merging this PR.

levb:
@javaguirre since your issue has been filed as a separate ticket, can you approve and unblock this PR?

javaguirre (approved):

Issue #396, opened 2022-11-21
Setting expand to "summary" results in no expansion for oauth2_user and oauth2_app


mickmister:
There is a breaking change in https://github.com/mattermost/mattermost-plugin-apps/pull/336, where we stopped allowing expand `summary` for some expandable objects. Some apps being developed were using this, and the apps are now breaking from this change. Note that this information is not documented here https://developers.mattermost.com/integrate/apps/structure/call-metadata/#expand-levels, so there's not a standard to adhere to, but we should try not to introduce changes that could break existing apps.

PR #397, opened 2022-11-21
Allow expand "summary" for oauth2_user and oauth2_app

mickmister:
#### Summary

There is a breaking change in https://github.com/mattermost/mattermost-plugin-apps/pull/336, where we stopped allowing expand `summary` for some expandable objects. Some apps being developed were using this, and the apps are now breaking from this change. Note that this information is not documented here https://developers.mattermost.com/integrate/apps/structure/call-metadata/#expand-levels, so there's not a standard to adhere to, but we should try not to introduce changes that could break existing apps.

This PR makes it so the `oauth2_app` and `oauth2_user` contexts accept the `summary` expand level

#### Ticket Link

Fixes https://github.com/mattermost/mattermost-plugin-apps/issues/396

### DIFF
--- a/server/proxy/expand.go
+++ b/server/proxy/expand.go
@@ -93,14 +93,12 @@ func (e *expander) expand(expand *apps.Expand) (*apps.Context, error) {
 		f              expandFunc
 		requestedLevel apps.ExpandLevel
 		expandableAs   []apps.ExpandLevel
-		err            error
 	}{
 		{
 			name:           "acting_user_access_token",
 			requestedLevel: expand.ActingUserAccessToken,
 			f:              e.expandActingUserAccessToken,
-			expandableAs:   []apps.ExpandLevel{apps.ExpandAll},
-			err:            utils.NewInvalidError(`invalid expand: "acting_user_access_token" must be "all" or empty`),
+			expandableAs:   []apps.ExpandLevel{apps.ExpandAll, apps.ExpandSummary},
 		}, {
 			name:           "acting_user",
 			requestedLevel: expand.ActingUser,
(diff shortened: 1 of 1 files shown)


hanzei (commented):
#336 introduced a couple of changes. Why this is one of them that should be reverted?

mickmister:
@hanzei A partnering org ran into issues with this specific change. More context here https://community.mattermost.com/core/pl/mpm61cztdibzzgn79gzytj3m4w

Are there breaking changes in that PR that would require releasing a `2.0` for the framework? As noted in the PR description, the previous behavior wasn't documented, but we did change how the API worked and caused issues with that specific change. 0/5 on if the individual apps should be patched, or if we should revert this specific change.

hanzei:
> Are there breaking changes in that PR that would require releasing a `2.0` for the framework?

I don't think so. Without any app that runs in production, I'm fine with doing breaking changes.

https://github.com/mattermost/mattermost-plugin-apps/pull/336 does include breaking changes that are worth the hassle. Some of them are listed in the PR description. This one, unfortunately, isn't. @levb Would you please update that PR description with the full list of breaking changes? We will need that for the release notes.

Coming back to this change: 2/5 that I like the current behaviour as it's'more strict and allows us to add a meaningful use of `summary` later.

mickmister:
@hanzei If there are known breaking changes, shouldn't we release a version 2? Since we released a version 1, we should stay true to semver and release another major version.

If we want to have an exception for this project because we assume there are no apps in production, I would say it depends on the severity of the breaking changes. This PR addresses a specific change that is affecting known apps in development, so it may affect apps being developed externally as well.

For these fields specifically `oauth2_user` and `oauth2_app`, I think it's unlikely they are being used in apps developed externally. The bot account functionality has been the focal point of usage, rather than user impersonation or mananging OAuth tokens for external services. I also think the usage of external OAuth apps is less likely than having user impersonation in the App.

So for this change, maybe it doesn't need a major version bump. There are other crucial changes in that PR you linked to. Whether they should constitute a breaking change, I'm not sure. We should gauge the impact on our known app developers. @dsharma522 Do you have any thoughts on this based on your experience with the framework?

mickmister:
@cwarnermm @neflyte This piece of the framework can be confusing, and it's behavior is not clearly documented here https://developers.mattermost.com/integrate/apps/structure/call-metadata/#expand-levels. We would want an explanation of each expand level for each object type. This would be displayed in a table, similar to the tables currently on that page.

The source of truth of this information is currently spread throughout this file https://github.com/mattermost/mattermost-plugin-apps/blob/master/server/proxy/expand.go. Maybe I can put this information in a spreadsheet to be used as a base for documentation?

It's important to note that this information is subject to change, so documentation could easily get out of date. But I'm hoping that documenting it will make the choices more concrete, and it forces us to adhere to the published documentation.

levb:
@mickmister, I agree with @hanzei that the current stricter behavior is intentional and we should not be changing it.

@hanzei It's be quite tedious for me to document every change now, partly because we did not have a consistent behavior nor tests before. 1/5 I'd argue this is the first time we have a "reference" implementation. If you think it is important for us to "release-note" changes from the previously undocumented behavior, I'll try to produce a comprehensive summary once I have access to a large screen next week.

neflyte:
@mickmister @levb @hanzei I would suggest taking an approach similar to the server team where they added GoDoc sufficient to be extracted - using golang reflection - and formatted to align with the docs site. With this approach, changes are automatically documented if the GoDoc has been added.

Thoughts?

mickmister:
@neflyte I like that idea a lot :+1:

mickmister:
Closing, as we're keeping the changes and working on proper documentation as discussed above

PR #400, opened 2022-11-23
Test app - Add num_bindings command to set number of bindings in different locations

mickmister:
#### Summary

This PR serves as a tool to test the functionality of the client-side functionality of the App framework. It adds a command `/test num_bindings (location) (number)` that allows you to set the number of bindings like so:

- `/test num_bindings post_menu 2` - two items in the post menu
- `/test num_bindings channel_header 0` - no items in the channel header / Apps bar
- `/test num_bindings channel_header -1` - normal set of test bindings

#### Ticket Link

Fixes https://github.com/mattermost/mattermost-plugin-apps/issues/401

### DIFF
--- a/test/app/command.go
+++ b/test/app/command.go
@@ -13,12 +13,50 @@ func commandBindings(cc apps.Context) []apps.Binding {
 			formCommandBinding(cc),
 			subscriptionCommandBinding("subscribe", Subscribe),
 			subscriptionCommandBinding("unsubscribe", Unsubscribe),
-
+			numBindingsCommandBinding(cc),
 			testCommandBinding(cc),
 		},
 	}
 
 	return []apps.Binding{b}
+}
+
+func numBindingsCommandBinding(cc apps.Context) apps.Binding {
+	return apps.Binding{
+		Label:       "num_bindings",
+		Description: "Choose how many bindings to show in different locations. Provide -1 to use the default options.",
+		Icon:        "icon.png",
+		Form: &apps.Form{
+			Submit: apps.NewCall(NumBindingsPath),
+			Fields: []apps.Field{
+				{
+					Name:                 "location",
+					Type:                 apps.FieldTypeStaticSelect,
+					Description:          "Location to change",
+					IsRequired:           true,
+					AutocompletePosition: 1,
+					SelectStaticOptions: []apps.SelectOption{
+						{
+							Label: "post_menu",
+							Value: "post_menu",
+						},
+						{
+							Label: "channel_header",
+							Value: "channel_header",
+						},
+					},
+				},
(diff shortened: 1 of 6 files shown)


hanzei (changes requested):
Could you elaborate a bit on what the use case is to limit the bindings by an arbitrary number? I see the use case of an on/off flag, but I'm wondering what value the arbitrary number provides.

mickmister:
@hanzei: @DHaussermann is currently testing some mobile functionality for the App framework https://github.com/mattermost/mattermost-mobile/pull/6735, which involves dynamic UI behavior in the post options menu, depending on how many bindings are present.

Being able to choose how many items are displayed there immensely helps with testing different devices. I've already gotten use out of this as it helped me find and fix a bug with this feature.

test/app/num_bindings.go
```diff
@@ -0,0 +1,36 @@
...
+	"strconv"
+
+	"github.com/gorilla/mux"
+	"github.com/pkg/errors"
+
+	"github.com/mattermost/mattermost-plugin-apps/apps"
+)
+
+var numPostMenuBindings int64 = -1
+var numChannelHeaderBindings int64 = -1
```
mickmister, line 14: Obviously not a fan of using globals like this, but I'm thinking the test app project should be re-written to use things like kv store to properly support this feature. In the meantime, this seemed fine since this is not a production application.

test/app/command.go
```diff
@@ -13,14 +13,52 @@ func commandBindings(cc apps.Context) []apps.Binding {
...
 		},
 	}
 
 	return []apps.Binding{b}
 }
 
+func numBindingsCommandBinding(_ apps.Context) apps.Binding {
+	return apps.Binding{
+		Label:       "num_bindings",
+		Description: "Choose how many bindings to show in different locations. Provide -1 to use the default options.",
```
hanzei, line 27: Can the description make it more clear that this is a temporary change?

hanzei (approved):
@mickmister Thanks for the explanation. If the command is useful for testing, I have no objections to merging this PR.

2/5 I would like for the setting to be persistent. The current `INCLUDE_INVALID=true` flag would also be a good one to migrate to KV.

levb:
> 2/5 I would like for the setting to be persistent. The current INCLUDE_INVALID=true flag would also be a good one to migrate to KV.

@hanzei et all: I agree, but let's perhaps strategize a little bit about the role and the functionality of the test app (vs examples vs tests). 0/5 with goapp we can produce specific test apps easily and it may be better to have != 1 test apps.

levb (approved):

levb:
A test-only PR, merging

PR #402, opened 2022-11-29
Change message when not subscribe to notification (#361)

### LINKED ISSUE #361: Test-App could use a friendly message when trying to remove a subscription that does not exist
For the Test-App, When running the command to remove a subscription that does not exist it would be nice to have a friendlier message. 

The current `not found` is confusing as it looks like an error has occurred.

image

javaguirre:
#### Summary

I was learning about the test app, fixes #361

### DIFF
--- a/server/appservices/subscriptions.go
+++ b/server/appservices/subscriptions.go
@@ -166,7 +166,7 @@ func (a *AppServices) unsubscribe(r *incoming.Request, ownerUserID string, e app
 		return modified, nil
 	}
 
-	return all, utils.ErrNotFound
+	return all, errors.New("You are not subscribed to this notification/s")
 }
 
 func (a *AppServices) hasPermissionToSubscribe(r *incoming.Request, sub apps.Subscription) func() error {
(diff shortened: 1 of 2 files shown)


server/appservices/subscriptions.go
```diff
@@ -166,7 +166,7 @@ func (a *AppServices) unsubscribe(r *incoming.Request, ownerUserID string, e app
 		return modified, nil
 	}
 
-	return all, utils.ErrNotFound
+	return all, errors.New("You are not subscribed to this notification/s")
```
hanzei, line 169: Why the `/s`? Does the command accept more than one subscription?

hanzei, line 169: 2/5 that we should still `errors.Wrap` `utils.ErrNotFound` so the upper layers can check for that error.

hanzei (changes requested):

hanzei:
And a test needs to be fixed

server/appservices/subscriptions.go
javaguirre: mmm, I don't understand what you mean, but I'll change it.

hanzei (approved):

server/appservices/subscriptions.go
levb: @javaguirre see `errors.Cause`

levb (approved):

PR #403, opened 2022-11-30
Add multiselect users and channels to test app form

mickmister:


mickmister:
This already exists in `/test form multiselect`, but I'm thinking this PR still adds value since it adds it to the "full" form

PR #413, opened 2022-12-12
[MM-49024] Remove v0.7 app manifest workaround

hanzei:
#### Summary
Compatibility with the app manifest of v0.7 is no longer needed. All apps have been updated to the new schema.

This PR removes the workaround that supported v0.7 manifests.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-49024


apps/manifest.go
```diff
@@ -146,29 +139,16 @@ type Manifest struct {
 // Thus, custom functions to encode/decode JSON, with backwards compatibility
 // support for App and Manifest.
 func DecodeCompatibleManifest(data []byte) (m *Manifest, err error) {
```
hanzei, line 141: I would like to rename this function (and the app one) to  `UnmarshalManifest` (and `UnmarshalApp`) as they are not longer about compatibility. Thoughts?

levb (approved):
👍 Thank you!

apps/manifest.go
hanzei: @levb What do you think?

levb: I thought about it too, as I was reviewing the code. 0/5 `Parse[Manifest|App]JSON`? But I don't think it's a must do.

PR #414, opened 2022-12-15
Moved cc log admin to /apps settings 

levb:
#### Summary
- Moved the log settings under `/apps settings`
- Also restored simple JSON marshaling of `StoredConfig` by using pointers

#### Ticket Link
none

#### Screenshots
image
image
image
image
image
image
image


javaguirre:
I will test the following cases:

- [ ] Activate logs creating a new channel
- [ ] Activate logs selecting an existing channel
- [ ] Receive logs on a channel without the --json flag.
- [ ] Receive logs on a channel with the --json flag.
- [ ] Deactivating developer mode and/or testing commands with stop logging
- [ ] If we create a channel and it was created, it's not created twice
- [ ] Setting a different --level will filter which logs to receive.
- [ ]  If we specify both --create-channel and --channel, it fails.
- [ ]  Logs should appear in a channel from the team where I executed the command.

javaguirre:
Not sure if this is a bug or not, but if I have a channel selected, I open the settings and try to select 'create a new channel', the option is not there, I have to persist the settings with the 'do not copy app logs to a channel', then open the settings again to select the 'create channel' selection (see video).

https://user-images.githubusercontent.com/488556/208077069-43e5cb28-fdac-4c99-b0ff-c0f43fd1b18e.mov

javaguirre:
If I disable the 'Developer Mode' in MM System Console, the apps settings won't check it's disabled and it will tell it's enabled, I think both should be in the same stage since we're talking about the same? Is this a different Developer Mode?

https://user-images.githubusercontent.com/488556/208078557-8977f21a-6309-44aa-883d-64161eebdee6.mov

levb:
@javaguirre the /apps settings command provides for overriding the system-wide "Dev mode" (dev+test), so even when it's off an admin can turn it on, for the apps plugin only. Ditto HTTP. So, when the override mode is on, the overrides are used; when it's off - the system-wide settings are. 

If the override mode in the apps plugin is off, the plugin should receive `OnConfigurationChanged` when you apply changes in the system console, and adjust accordingly. If it's not happening - it's a bug.

javaguirre:
> @javaguirre the /apps settings command provides for overriding the system-wide "Dev mode" (dev+test), so even when it's off an admin can turn it on, for the apps plugin only. Ditto HTTP. So, when the override mode is on, the overrides are used; when it's off - the system-wide settings are.
> 
> If the override mode in the apps plugin is off, the plugin should receive `OnConfigurationChanged` when you apply changes in the system console, and adjust accordingly. If it's not happening - it's a bug.

Thank you for the clarification! I understand now. 👍

javaguirre:
When I try to set the `--level` of the apps logs debug command, the autocomplete works, but when I submit the action it fails (attachment).

Screenshot 2022-12-16 at 11 32 51

levb:
@javaguirre that's weird, the `debug logs` command should be gone entirely... refresh your client? (bindings?)

javaguirre:
> @javaguirre that's weird, the `debug logs` command should be gone entirely... refresh your client? (bindings?)

I thought so, I executed a `make restart` so it builds the frontend too, the webapp notified me to refresh, but the bindings are still there, I might be doing something wrong, not sure if I have to clean something else, I thought it was enough with the `restart` command.

javaguirre:
When I select a level and open the settings again, the current level is not selected, I think that would be nice.

https://user-images.githubusercontent.com/488556/208082153-5d20d0c1-c905-4aae-91ee-6869ed2ee39e.mov

javaguirre:
This is no longer happening 👏 

https://github.com/mattermost/mattermost-plugin-apps/pull/394#issuecomment-1347948166

javaguirre (changes requested):
Let me know if you think some of the issues I found aren't a problem or if we want to fix those later in a different ticket. 👍

levb:
> the current level is not selected

@javaguirre ditto JSON, thx! Forgot...

levb:
@javaguirre I added the current value selection for Level and JSON

javaguirre:
> @javaguirre I added the current value selection for Level and JSON

I tested it and It works great 👍

javaguirre:
https://user-images.githubusercontent.com/488556/210335874-9064006e-d160-4d0c-9129-6bedb9470108.mov

When I enable the developer mode I can't select the channel for logging until I persist that change. I have to open the modal again and then it's available to select.

I was expecting to change both at the same time, is it possible?

javaguirre:
> @javaguirre the /apps settings command provides for overriding the system-wide "Dev mode" (dev+test), so even when it's off an admin can turn it on, for the apps plugin only. Ditto HTTP. So, when the override mode is on, the overrides are used; when it's off - the system-wide settings are.
> 
> If the override mode in the apps plugin is off, the plugin should receive `OnConfigurationChanged` when you apply changes in the system console, and adjust accordingly. If it's not happening - it's a bug.

Did you manage to fix this one?

javaguirre (changes requested):
- [X] Setting a different --level will filter which logs to receive.
- [X] If we specify both --create-channel and --channel, it fails. (Not relevant anymore)
- [X] Logs should appear in a channel from the team where I executed the command.

I tested the remaining ones and they work great, only found two issues this time:

- [ ] https://github.com/mattermost/mattermost-plugin-apps/pull/414#issuecomment-1369590402
- [ ] https://github.com/mattermost/mattermost-plugin-apps/pull/414#issuecomment-1369585445

Thank you Lev!

levb:
For the video - not currently fixable, filed https://mattermost.atlassian.net/browse/MM-49451

As for setting the dev mode - the apps plugin will not set the server's dev mode, only it's own. Sorry for the confusion.

Consider both resolved?

javaguirre (approved):
LGTM

levb:
@hanzei @mickmister Can I please get a 2nd review, I'd like to merge it soon now that 1.2 is out

server/config/service.go
```diff
@@ -85,18 +90,25 @@ func (s *service) newConfig() Config {
...
 	conf := s.newConfig()
+	conf.StoredConfig = newStoredConfig
 	newMattermostConfig := s.reloadMattermostConfig()
-	license := s.getMattermostLicense(log)
+
+	if conf.DeveloperModeOverride != nil {
+		conf.DeveloperMode = *conf.DeveloperModeOverride
+	} else {
+		conf.DeveloperMode = pluginapi.IsConfiguredForDevelopment(newMattermostConfig)
+	}
```
mickmister, line 103: Is it possible to revert the `DeveloperModeOverride`, i.e. the value made `nil` again? `override === false` shouldn't necessarily mean that we don't want developer mode on right?

server/plugin.go
```diff
@@ -169,9 +169,14 @@ func (p *Plugin) OnConfigurationChange() (err error) {
 	p.tracker.UpdateTracker(updatedTracker)
 
 	mm := pluginapi.NewClient(p.API, p.Driver)
-	cm := mm.Configuration.GetPluginConfig()
+	var sc config.StoredConfig
+	err := mm.Configuration.LoadPluginConfiguration(&sc)
+	if err != nil {
+		p.API.LogInfo("failed to load updated configuration", "error", err.Error())
```
mickmister, line 175: Should this be warn instead?

```suggestion
		p.API.LogWarn("failed to load updated configuration", "error", err.Error())
```

mickmister (approved):
LGTM

server/config/service.go
levb: Yes, and because DeveloperModeOverride is a stored config value, a change to it is propagated in OnConfigurationChanged, and `DeveloperMode` is set to the "server" value.

DHaussermann:
All config are functional.
It is a bit confusing that some fields must be set first. I was also a bit confused that you can't simply change the channel the logs get written to. Logging to a channel must be disable and re-enabled. 

@levb seeing as how the snags are minor and this is a debug tool - We can see if UX improvements are worth the effort later.

LGTM!

DHaussermann (approved):
Approved as per comment above.

server/config/config.go
```diff
@@ -33,29 +31,14 @@ type StoredConfig struct {
 	// manifest_<sha1(Manifest)>. Implementation in `store.Manifest`.
 	LocalManifests map[string]string `json:"local_manifests,omitempty"`
 
-	DeveloperMode bool `json:"developer_mode"`
-	AllowHTTPApps bool `json:"allow_http_apps"`
+	DeveloperModeOverride *bool `json:"developer_mode"`
+	AllowHTTPAppsOverride *bool `json:"allow_http_apps"`
```
hanzei, line 35: Due to the change to a pointer, the logging in https://github.com/mattermost/mattermost-plugin-apps/blob/461e9e66bb08cee1effcff459249072ce0c816ff/server/config/service.go#L272 is broken: `Storing configuration, 3 installed , 0 listed apps, developer mode 0xc000e41488, allow http apps 0xc000e41489`

hanzei: Fixed in https://github.com/mattermost/mattermost-plugin-apps/pull/440/

hanzei (commented):
What UX issues do the introduction of overrides in the settings modal solve? What downsides had the previous implementation?
Screenshot from 2023-01-31 11-14-37

hanzei:
@levb Heads up that this PR will need to update the docs that were originally added in https://github.com/mattermost/mattermost-developer-documentation/pull/1181

PR #417, opened 2022-12-19
MM-48137 "Join/Leave" notification refactor

levb:
Need to summarize


server/appservices/subscriptions.go
```diff
@@ -38,6 +38,34 @@ func (a *AppServices) Subscribe(r *incoming.Request, sub apps.Subscription) erro
 		return err
 	}
 
+	// TODO <>/<> MM-??? deprecate bot_joined|left_...
```
javaguirre, line 41: I guess we need to create a ticket for this and remove the comment?

server/appservices/subscriptions.go
```diff
@@ -166,6 +194,17 @@ func (a *AppServices) unsubscribe(r *incoming.Request, ownerUserID string, e app
 		return modified, nil
 	}
 
+	// TODO <>/<> MM-??? deprecate bot_joined|left_...
```
javaguirre, line 197: ditto

test/app/subscriptions.go
```diff
@@ -39,7 +39,27 @@ var allSubjects = map[apps.Subject]apps.Expand{
...
+		User:       apps.ExpandAll,
+		Team:       apps.ExpandAll,
+		TeamMember: apps.ExpandAll,
+	},
+	apps.SubjectSelfLeftTeam: {
+		User:       apps.ExpandAll,
+		Team:       apps.ExpandAll,
+		TeamMember: apps.ExpandAll,
+	},
+	// apps.SubjectSelfMentioned: {
```
javaguirre, line 62: could we remove the code? we can recover it with `git log -p`

javaguirre (changes requested):
LGTM, only a couple of comments.

server/appservices/subscriptions.go
levb: yes, I actually want to sprinkle the core with the ticket reference to make it easier to remove later.

test/app/subscriptions.go
levb: @mickmister @DHaussermann I am maintaining my view that functionality that is covered by restapitest should be excluded from the test app to minimize its complexity and maintenance. Why do we need subscriptions in the test app?

levb:
Going to combine the 2 subs PRs, clean up and request a fresh review

PR #418, opened 2022-12-20
Add tests for webhook paths

mickmister:
#### Summary

This PR is meant to test the functionality of the `OnRemoteWebhook` call provided in the manifest.


test/restapitest/webhook_test.go
```diff
@@ -0,0 +1,223 @@
...
+				return apps.NewTextResponse("success")
+			})
+
+			webhookSecret := getTestWebhookSecret(th, app)
+			err := sendWebhookPostRequest(th, webhookSecret, tc.reqPath)
+			if err != nil && (tc.called || err.Error() != "404 page not found") {
+				th.Fatal("received error for webhook request: " + err.Error())
+			}
+
+			var result bool
```
levb, line 214: Is this consistent with how we check for notifications? I'd like to make it a bit more robust, but as a separate PR.

levb (approved):

test/restapitest/webhook_test.go
mickmister: @levb I'm definitely open for improvement here. The difference between this and notifications is that this isn't always called, so we can't just block and wait for the receive from the channel. Not saying you're saying otherwise, just pointing it out

javaguirre (approved):
LGTM

levb:
@mickmister not sure why automerge label did not fire; @DHaussermann this is a test-only PR, so no QA needed.

PR #419, opened 2022-12-20
Remove ServiceNow from manifests and add Zendesk

mickmister:


assets/manifests.json
```diff
@@ -1,3 +1,3 @@
 {
-    "servicenow": "0.3.4"
+    "zendesk": "0.3.0"
```
hanzei, line 2: Does https://github.com/mattermost/mattermost-app-zendesk/pull/94 need to be merged before a new, compatible release can be cut?

hanzei (changes requested):

assets/manifests.json
mickmister: @hanzei Yes I'm just creating this as a placeholder

hanzei (approved):

PR #428, opened 2023-01-04
[MM-34723] OAuth2 User records must be stored encrypted

javaguirre:
#### Summary

Encrypting oauth2 user data on the KV store.

I have decided to generate an AES key on plugin activation if there wasn't any and store it in the StoredConfig. It was the best solution I came up with given what we have now.

I have created an Encrypter with an interface we can implement in the future if we want an external secret store, I added it as a dependency for the OAuth store following the dependency injection pattern, so then it was easy to implement testing using the [Fake pattern](https://www.educative.io/answers/what-is-faking-vs-mocking-vs-stubbing) to avoid mocking on the OAuth2Store test.

Let me know your thoughts.

#### Ticket Link

Fixes https://mattermost.atlassian.net/browse/MM-34723


javaguirre:
I have to fix an assertion in a test, but can you give your opinion? Thank you!

server/plugin.go
```diff
@@ -91,6 +91,23 @@ func (p *Plugin) OnActivate() (err error) {
...
+		if encErr != nil {
+			log.Errorf("Couldn't generate the encryption key")
+		}
+
+		sc.EncryptionKey = encKey
+		encErr = confService.StoreConfig(sc, log)
+		if encErr != nil {
+			log.Errorf("Couldn't store encryption key")
+		}
+	}
```
hanzei, line 109: 2/5 that `server/config.service.newInitializedConfig()` should take care of the default values

server/store/encrypter.go
```diff
@@ -0,0 +1,100 @@
...
+
+	"github.com/pkg/errors"
+)
+
+const DefaultAesKeySize = 32
+
+type Encrypter interface {
+	Encrypt(message []byte) ([]byte, error)
+	Decrypt(message []byte) ([]byte, error)
+}
```
hanzei, line 19: Nice use of an interface :+1:

hanzei (changes requested):
I have to double-check the encryption algorithm later, but the code looks solid :+1:

server/config/service.go
```diff
@@ -133,6 +136,17 @@ func (s *service) newInitializedConfig(newStoredConfig StoredConfig) (*Config, u
...
+	// Generate an encryption key on the fly
+	// to encrypt/decrypt oauth user data
+	if conf.EncryptionKey == nil {
+		encKey, encErr := GenerateEncryptionKey()
+		if encErr != nil {
+			log.Errorf("Couldn't generate the encryption key for OAuth user data encryption")
+		}
+
+		conf.EncryptionKey = encKey
+	}
```
hanzei, line 148: I don't see that the `EncryptionKey` actually gets stored in the config file if no other event like `/apps settings` triggered the store process.

The steps to reproduce are:
1. Start server with empty `EncryptionKey`
2. Store stuff in OAuth store
3. Restart server
4. Confirm no `EncryptionKey` is stored (That is the issue!)

server/store/encrypter.go
```diff
@@ -0,0 +1,88 @@
...
+	}
+
+	return src[:(length - unpadding)], nil
+}
+
+func (s *AESEncrypter) pad(src []byte) []byte {
+	padding := aes.BlockSize - len(src)%aes.BlockSize
+	padtext := bytes.Repeat([]byte{byte(padding)}, padding)
+	return append(src, padtext...)
+}
```
hanzei, line 40: Why do we need the padding? `crypto/cipher.Stream.XORKeyStream` doesn't have any length requirement to the `src`.

hanzei (changes requested):

server/store/encrypter.go
hanzei: I still don't understand where these additional characters come from. Does the encryption stream add them?

FWIW: The go example also doesn't use padding: https://pkg.go.dev/crypto/cipher#example-NewCFBDecrypter

hanzei (changes requested):

server/store/encrypter.go
javaguirre: I changed it to base64 + CFBEncrypter/Decrypter, much more stable and easy to debug.

server/store/oauth2.go
```diff
@@ -71,7 +73,12 @@ func (s *oauth2Store) SaveUser(appID apps.AppID, actingUserID string, data []byt
 		return err
 	}
 
-	_, err = s.conf.MattermostAPI().KV.Set(userkey, data)
+	dataEncrypted, err := s.encrypter.Encrypt(string(data))
+	if err != nil {
+		return err
+	}
```
hanzei, line 79: Let's wrap the error for more context:
```suggestion
	if err != nil {
		return errors.Wrap(err, "failed to encrypt oauth user data")
	}
```

server/store/oauth2.go
```diff
@@ -90,5 +97,16 @@ func (s *oauth2Store) GetUser(appID apps.AppID, actingUserID string) ([]byte, er
...
+	// Backwards compatibility, if the data is JSON or empty it means it's not encrypted
+	// so we return it as is
+	if data == nil || json.Valid(data) {
+		return data, nil
+	}
+
+	dataDecrypted, err := s.encrypter.Decrypt(string(data))
+	if err != nil {
+		return nil, err
+	}
```
hanzei, line 109:
```suggestion
	if err != nil {
		return errors.Wrap(err, "failed to decrypt oauth user data")
	}
```

hanzei (approved):
Nice :rocket:

PR #429, opened 2023-01-12
[MM-48800] Refresh bindings on activation

javaguirre:
#### Summary

Disabling and enabling the apps plugin does make Bindings disappear, so we need to refresh the bindings on activation.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-48800

### DIFF
--- a/server/plugin.go
+++ b/server/plugin.go
@@ -135,6 +135,7 @@ func (p *Plugin) OnActivate() (err error) {
 	log.Infof("activated")
 
 	p.conf.MattermostAPI().Frontend.PublishWebSocketEvent(config.WebSocketEventPluginEnabled, conf.GetPluginVersionInfo(), &model.WebsocketBroadcast{})
+	p.conf.MattermostAPI().Frontend.PublishWebSocketEvent(config.WebSocketEventRefreshBindings, map[string]interface{}{}, &model.WebsocketBroadcast{})
 
 	return nil
 }
--- a/server/plugin_test.go
+++ b/server/plugin_test.go
@@ -62,6 +62,7 @@ func TestOnActivate(t *testing.T) {
 	testAPI.On("RegisterCommand", mock.AnythingOfType("*model.Command")).Return(nil)
 
 	testAPI.On("PublishWebSocketEvent", "plugin_enabled", map[string]interface{}{"version": manifest.Version}, &model.WebsocketBroadcast{})
+	testAPI.On("PublishWebSocketEvent", "refresh_bindings", map[string]interface{}{}, &model.WebsocketBroadcast{})
 
 	err := p.OnActivate()
 	require.NoError(t, err)


levb (approved):
LGTM and works for `make deploy`

javaguirre:
> I'm still wondering what introduced the regression. Was there some client-side logic that took care of refreshing the bindings when the plugin gets enabled? We are sending a specific `plugin_enabled` event to the clients. Why do we need another one?

Not sure what changes are you requesting. Can you clarify? Thank you!

mickmister:
Does the client send a request to `bindings`, with the current state of the frontend and plugin? Maybe the plugin isn't ready to receive requests if this is the case? I would also like to see if this is the current behavior of the mobile app.

mickmister:
> We are sending a specific `plugin_enabled` event to the clients. Why do we need another one?

> Not sure what changes you are requesting. Can you clarify? Thank you!

@javaguirre I think he's saying, "We're already telling the frontend that the App framework plugin has just been enabled. The frontend should be smart enough to fetch new bindings when receiving that websocket message, instead of having the plugin explicitly send a separate `refresh bindings` websocket message."

javaguirre:
> > We are sending a specific `plugin_enabled` event to the clients. Why do we need another one?
> 
> > Not sure what changes you are requesting. Can you clarify? Thank you!
> 
> @javaguirre I think he's saying, "We're already telling the frontend that the App framework plugin has just been enabled. The frontend should be smart enough to fetch new bindings when receiving that websocket message, instead of having the plugin explicitly send a separate `refresh bindings` websocket message."

Great! I'll test that then, thank you!

javaguirre:
I'm testing this change https://github.com/mattermost/mattermost-webapp/pull/12036

If it gets accepted and solves the issue, we might need to close this PR.

javaguirre:
This wasn't straight forward to me, but when we deploy, at least using `make deploy` and the deploy ends, `plugin_enabled` is not triggered, but `plugin_statuses_changed`, does that sounds good to you? because in the code we are supposedly triggering `pluginenabled`.

https://user-images.githubusercontent.com/488556/213689097-e02b3e4a-263e-40d6-b65b-b7789a9af5ce.mov

Screenshot 2023-01-20 at 12 56 44

javaguirre:
I think I found the issue, and could be happening also with plugin_disabled. We are sending custom_com.mattermost.....plugin_enabled, but the frontend expects `plugin_enabled` so it will never trigger that event, does that sound good to you? am I missing some configuration? @levb 


https://user-images.githubusercontent.com/488556/213690122-efb5fe3d-29f3-41f3-9a74-2ec0756ed28b.mov

Screenshot 2023-01-20 at 13 01 41

Screenshot 2023-01-20 at 13 02 20

https://github.com/mattermost/mattermost-webapp/blob/master/actions/websocket_actions.jsx#L321

Part of the solution would be to send the socket action the webapp expects and then the change I made on the webapp to refresh bindings, let me know your opinions.

javaguirre:
Closing this, We're solving it at webapp level.

PR #430, opened 2023-01-12
[MM-40012] revoke access when app disabled

javaguirre:
#### Summary

This PR does two things:

- [Revokes all user sessions for a specific app when it's disabled](https://github.com/mattermost/mattermost-plugin-apps/pull/430/files#diff-10019993668a0be2f95c0a2a41f972a6098553d8cd4750840eecc8f61da850a2R99)
- [`store.Get` can't be used if the app is Disabled](https://community.mattermost.com/private-core/pl/pp1ot3y5o3ri7mips96sbicrph)

I'm not sure about the performance of this solution, I need to get more context, let me know if you see something that might be dangerous.

Let me know if you want to separate these two solutions on different PRs or if we don't want to implement one or the other yet.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-40012


server/session/session.go
```diff
@@ -174,6 +175,22 @@ func (s service) RevokeSessionsForApp(r *incoming.Request, appID apps.AppID) err
...
+	if err != nil {
+		return err
+	}
+
+	for _, userKey := range userKeys {
+		err := s.RevokeSessionsForUser(r, userKey)
+		if err != nil {
+			r.Log.WithError(err).Warnw("failed to revoke session for user")
+		}
+	}
```
hanzei, line 189: This will cause #OfUsers RPC requests per app deactivation. That seems like a lot to me. Should the session revocation happen on the MM server side?

hanzei (changes requested):
A performance concern before I jump into the details

server/session/session.go
javaguirre: yep, I was coming to that same conclusion, I'll check what can I do on the server to make this more performant

hanzei: Maybe it make even sense to add a `Disabled` state to the OAuth app server side and let the server handle the session revokation.

mickmister: @hanzei If we tell the server that this app is "disabled", we can then have the "soft revoke" the sessions. Meaning, if we re-enable the app then the sessions are back online. Just a thought. There are tradeoffs to that versus permanently revoking each session.

server/store/session.go
```diff
@@ -144,6 +146,33 @@ func (s sessionStore) ListForApp(appID apps.AppID) ([]*model.Session, error) {
 	return ret, nil
 }
 
+func (s sessionStore) ListUsersWithSessions(appID apps.AppID) ([]string, error) {
+	userKeys := make([]string, 0)
+
+	for i := 0; ; i++ {
```
mickmister, line 152: Since we're not using `i`

```suggestion
	for {
```

mickmister: What does this loop do for us? It seems like the exact same operation is done on each iteration

server/session/session.go
hanzei: Yep, that's what I also had in mind :+1: Let's traffic for the DB.

server/store/session.go
javaguirre: yep, that doesn't feel right.

server/session/session.go
javaguirre: 👍

hanzei: > so even adding a disabled flag we couldn't (soft)revoke the sessions in the mattermost-server, we need the list of sessions stored in the plugin-apps KV store

Doesn't the server already have the list of session tied to an OAuth App via https://github.com/mattermost/mattermost-plugin-apps/blob/f05a93d06104115edc6ea816710974fcdbe2ed4a/server/session/session.go#L96-L103?

javaguirre: I'm not sure, there is alot to cover to understand the session flow in the MM server. Thank you for the tip!

javaguirre:
Today I've been thinking I'm gonna close this PR and start from scratch. After talking this week to @hanzei this is what makes more sense IMO.

PR #432, opened 2023-01-24
[MM-49730] Enable bindings call when the OAuthEnabled system setting is false

javaguirre:
#### Summary

Check OAuth enabled system setting check on activation to avoid activation of the plugin when the setting is off.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-49730


server/plugin.go
```diff
@@ -91,6 +91,11 @@ func (p *Plugin) OnActivate() (err error) {
 	conf := p.conf.Get()
 	log.With(conf).Debugw("configured the plugin.")
 
+	oauthEnabled := confService.MattermostConfig().Config().ServiceSettings.EnableOAuthServiceProvider
+	if oauthEnabled == nil || !*oauthEnabled {
+		return errors.New("the system setting `Enable OAuth 2.0 Service Provider` needs to be enabled in order for the Apps plugin to work. Please go to /admin_console/integrations/integration_management and enable it")
+	}
```
hanzei, line 97: As I mentioned in the meeting yesterday, failing on deactivation is not the right path here 4/5. The apps plugin should act as part of the MM system and not as a plugin. Admins should not even be aware that it's a plugin. 

I would advocate for fixing the bug mentioned in https://mattermost.atlassian.net/browse/MM-49730?focusedCommentId=141272 instead.

hanzei (changes requested):

server/plugin.go
javaguirre: which bug are you referring to? Do you mean only reporting the error on the plugin management page?

hanzei: > Since https://github.com/mattermost/mattermost-plugin-apps/pull/316  `/apps ` does not show any longer if OAuth 2 apps are disabled. The check in https://github.com/mattermost/mattermost-plugin-apps/blob/461e9e66bb08cee1effcff459249072ce0c816ff/server/builtin/app.go#L253-L257  errors even for a /bindings call.

mickmister: @hanzei So you are saying the real solution here is to "allow" OAuth 2.0 to be disabled for the purposes of starting the plugin and running these commands. I think we also said we want to enable the setting by default right?

hanzei: >  So you are saying the real solution here is to "allow" OAuth 2.0 to be disabled for the purposes of starting the plugin and running these command

IMO, yes

>  I think we also said we want to enable the setting by default right?

Yes. I filled https://mattermost.atlassian.net/browse/MM-49991 for that.

javaguirre: So, what I can do now is:

- Assure that the app commands unrelated to OAuth can run even with OAuth 2.0 disabled (like `/bindings`), is that it? Anything else? @hanzei @mickmister

javaguirre: I understand now, I could replicate it.

javaguirre: I tested against [this commit](https://github.com/mattermost/mattermost-plugin-apps/pull/316) and it works fine (actually I tested this myself [on the ticket](https://mattermost.atlassian.net/browse/MM-42915)). It's a side effect of another later change.

Did we add some restrictions when there is an error on the bindings so they don't get fetched again? Was this after this change? I can see the `OAuth` error in the logs, but then when I use the `/apps` command it doesn't get executed and it doesn't show the error, It just ignores the request.

https://user-images.githubusercontent.com/488556/215485056-73ba7657-e416-4810-ac67-3fbd513f86de.mov

@hanzei @mickmister

javaguirre: I've tested a bit more and it seems it would work until you refresh the page, then `/apps` won't work anymore.

hanzei: Bindings only get refreshed on a channel change or refresh.

javaguirre: I didn't know commands were bindings too.

javaguirre:
@hanzei I think this is what we discussed, I'm working on how to create a test when oauth is disabled.

The test file doesn't have anything new, but I think it's now easier to add use cases, so I left it as is.

hanzei (approved):
Nice one :+1:

hanzei:
@javaguirre Should the PR title and description get updated based on the changes?

levb (approved):

javaguirre:
I'll create tests in a different PR, It's much trickier than I expected and I want this to be merged.

hanzei:
I've reached out to the DevOps team regarding the CI issue: https://community.mattermost.com/core/pl/zaiodus9kbfi7basx6oizg9hze

PR #434, opened 2023-01-25
Remove app id for top level bindings

mickmister:
#### Summary

This currently isn’t causing any functional issues, though there is an issue where the top level binding coming from an app will contain an app id if the app supplied, whereas it will be blank if none was provided. This inconsistent behavior, depending on if a given app provides that id in its top level bindings, has confused me for some time so I fixed the issue.

#### Ticket Link

Fixes https://mattermost.atlassian.net/browse/MM-49971


hanzei (changes requested):
Good catch :+1: 

Is it feasible to cover that fix with a test?

PR #439, opened 2023-01-30
[MM-50056] Rejects manifests with empty DisplayName

hanzei:
#### Summary
The install process later fails with `failed to create OAuth2 App: OAuthApp.IsValid: Invalid name`  if they displayName is empty. It should fail sooner and warn about the empty display name.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-50056


javaguirre (approved):

levb (approved):

javaguirre (changes requested):
I think it would be good to have a specific test where the DisplayName fails, since we are saying it should "fail" in the tests, but it could be failing because of a different reason, WDYT?

Sorry I changed my review, but It tickled me after.

This will require to refactor manifest testing a bit.

hanzei:
That's a good suggestion @javaguirre Updated the tests in [`7a549ae` (#439)](https://github.com/mattermost/mattermost-plugin-apps/pull/439/commits/7a549aec328bd8ea5006b778b498a6031244a11a)

javaguirre (approved):

PR #440, opened 2023-01-30
[MM-50054] Fix panic on server with invalid configuration

hanzei:
#### Summary
If the server config is invalid, e.g. because `SITE_URL` is not set, the logger returned from `config.MakeService` is `nil`. In that case, the plugin logger should be used instead.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-50054

### DIFF
--- a/server/config/service.go
+++ b/server/config/service.go
@@ -52,7 +52,7 @@ type service struct {
 	mattermostConfig *model.Config
 }
 
-func MakeService(mm *pluginapi.Client, pliginManifest model.Manifest, botUserID string, telemetry *telemetry.Telemetry, i18nBundle *i18n.Bundle) (Service, utils.Logger, error) {
+func MakeService(mm *pluginapi.Client, pliginManifest model.Manifest, botUserID string, telemetry *telemetry.Telemetry, i18nBundle *i18n.Bundle, log utils.Logger) (Service, error) {
 	s := &service{
 		pluginManifest: pliginManifest,
 		botUserID:      botUserID,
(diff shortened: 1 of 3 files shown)


server/config/service.go
```diff
@@ -90,8 +90,7 @@ func (s *service) newConfig() Config {
 	}
 }
 
-func (s *service) newInitializedConfig(newStoredConfig StoredConfig) (*Config, utils.Logger, error) {
-	log := s.NewBaseLogger()
+func (s *service) newInitializedConfig(newStoredConfig StoredConfig, log utils.Logger) (*Config, error) {
```
hanzei, line 93: Before this change, the logger created here would be returned and used in `OnActivate()` causing a lot of messages to not be channel-logged as this logger doesn't have the stored config.

server/config/service.go
```diff
@@ -247,8 +246,15 @@ func (s *service) getMattermostLicense(log utils.Logger) *model.License {
 }
 
 func (s *service) Reconfigure(newStoredConfig StoredConfig, verbose bool, services ...Configurable) error {
-	// Use the old settings to log the processing of the new config.
-	clone, log, err := s.newInitializedConfig(newStoredConfig)
```
hanzei, line 251: This fixes another issue: Even if `verbose` was `false` `newInitializedConfig` would still log it's messages on every configuration change.

server/config/service.go
```diff
@@ -269,8 +272,12 @@ func (s *service) Reconfigure(newStoredConfig StoredConfig, verbose bool, servic
...
 
 func (s *service) StoreConfig(sc StoredConfig, log utils.Logger) error {
-	log.Debugf("Storing configuration, %v installed , %v listed apps, developer mode %v, allow http apps %v",
-		len(sc.InstalledApps), len(sc.LocalManifests), sc.DeveloperModeOverride, sc.AllowHTTPAppsOverride)
+	log.Debugf("Storing configuration: %d installed, %d listed apps, developer mode %t, allow http apps %t",
+		len(sc.InstalledApps),
+		len(sc.LocalManifests),
+		sc.DeveloperModeOverride != nil && *sc.DeveloperModeOverride,
+		sc.AllowHTTPAppsOverride != nil && *sc.AllowHTTPAppsOverride,
+	)
```
hanzei, line 280: Fixes https://github.com/mattermost/mattermost-plugin-apps/pull/414/files#r1088937047

levb (commented):
LGTM, TY!

levb (approved):

javaguirre (approved):

PR #405, opened 2022-12-02
MM-48793, MM-48137: improved bot_joined|left_channel|team

levb:
#### Summary
The `bot_joined|left_...` notifications had several issues making them dysfunctional for the use-cases. This PR adds new subjects `self_joined|left_...` and also provides backwards compatibility for the `bot_` subjects. 1/5 I'd prefer to remove the bot-specific subjects altogether, but not sure if it'd break any existing apps.

- Added `self_joined|left_channel|team` subscription subjects, available to all users not just the bot. These are globally scoped since they are already limited to the subscribing user.
- Changed Expand to bypass permission checks for these events since they are limited to the member of the team|channel.
- Replaced `server/mmclient` with `server/proxy/ExpandGetter`
- Deprecated `bot_joined|left_channel|team`
  - Previously stored (global?) subscriptions work.
  - New `bot_...` subscriptions are stored as `self_...`
- A small logging improvement in `goapp`

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-48137
https://mattermost.atlassian.net/browse/MM-48793

PR #443, opened 2023-01-30
MM-48793, MM-48137: improved bot_joined|left_channel|team

levb:
#### Replaces #405 
Upon a conversation with @hanzei we decided to change the semantics of `UserJoined|Left...` rather than introducing new subjects.

#### Summary
1. Makes `bot_(joined|left)_(channel|team)` global, and ensures reliable attribute expansion.
2. Makes `user_(joined|left)_(channel|team)` either accept the entity ID in which case they fire for all users joining/leaving the specified entity; or without the entity ID fire only for the subscriber, if joined/left any relevant entity.
3. Changed Expand to bypass permission checks for these events since they are limited to the member of the team|channel.
4. Replaced `server/mmclient` with `server/proxy/ExpandGetter`
5. A small logging improvement in `goapp`

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-48137
https://mattermost.atlassian.net/browse/MM-48793

#### TODO
- [x] finish the tests
- [x] fix the test app


apps/subscription.go
```diff
@@ -22,48 +22,48 @@ const (
...
 	SubjectUserCreated Subject = "user_created"
 
-	// SubjectUserJoinedChannel, SubjectUserLeftChannel watch the specified
-	// channel for users joining and leaving it.
+	// SubjectUserJoinedChannel, SubjectUserLeftChannel behave differently
+	// depending on the provided parameters. If no channel is specified, the
+	// subscription is system-wide, and the event is fired for every channel the
+	// subscriber joins or leaves. If a channel is specified (and the user has
+	// access to the channel), the event is fired for all users joining or
+	// leaving the channel.
```
hanzei, line 30: I'm not 100% a fan of the duality of the subject, but the previous `Self` subject was even more confusing. I guess it's fine.

If anyone has better ideas on how to structure the subject, please comment!

hanzei (approved):

apps/subscription.go
javaguirre: I think it would be nice to have a difference between `SubjectUserCreated` and `SystemSubjectUserCreated` or something like that. It could be a good idea to create a Jira ticket to explore this more in depth IMO.

server/proxy/notify.go
```diff
@@ -15,167 +17,199 @@ import (
...
 func (p *Proxy) NotifyUserCreated(userID string) {
-	p.notify(nil,
+	p.notifyAll(
 		apps.Event{
 			Subject: apps.SubjectUserCreated,
 		},
 		apps.UserAgentContext{
 			UserID: userID,
 		},
+		nil, // no special filtering, notify all subscriptions mathcing the event.
```
javaguirre, line 27: I think we'd need a Notification struct for the future so we can remove these comments and have less parameters on some of the methods.

server/proxy/notify.go
```diff
@@ -15,167 +17,199 @@ import (
...
+		newExpandSelfGetter(mm, user, member, nil, channel),
 	)
-}
 
-// NotifyUserJoinedTeam handles plugin's UserHasJoinedTeam callback. It emits
-// "user_joined_team" and "bot_joined_team" notifications to subscribed apps.
-func (p *Proxy) NotifyUserJoinedTeam(teamID, userID string) {
-	p.notifyUserTeam(teamID, userID, true, "NotifyUserJoinedTeam")
+	// Notify on the deprecated bot_joined|left_channel subscriptions.
+	if user.IsBot {
```
javaguirre, line 88: We can save an indentation here with `if !user.IsBot`

javaguirre (approved):
👍 I have some comments, but for the future.

hanzei:
@mickmister Would you mind also taking a look at the PR?

.golangci.yml
```diff
@@ -49,5 +49,14 @@ issues:
...
         - unparam
+        - staticcheck
+    - path: /
+      linters:
+        - staticcheck
+      text: "BotJoined"
+    - path: /
+      linters:
+        - staticcheck
+      text: "BotLeft"
```
mickmister, line 60: Wondering why we need this

mickmister (approved):

hanzei:
@levb How can @DHaussermann test this PR?

levb:
@hanzei @DHaussermann the PR is covered with REST API tests; I _think_ I fixed the test app, too.

hanzei:
@levb I will take a look at the test app later and check if I can get it to work

.golangci.yml
hanzei:
```
server/proxy/notify.go:90:13: SA1019: apps.SubjectBotJoinedChannel is deprecated: use SubjectUserJoinedChannel, etc. with no entity ID instead. (staticcheck)
		subject = apps.SubjectBotJoinedChannel
		          ^
server/appservices/subscriptions.go:193:8: SA1019: apps.SubjectBotJoinedChannel is deprecated: use SubjectUserJoinedChannel, etc. with no entity ID instead. (staticcheck)
		case apps.SubjectBotJoinedChannel,
		     ^
test/app/subscriptions.go:36:2: SA1019: apps.SubjectBotJoinedChannel is deprecated: use SubjectUserJoinedChannel, etc. with no entity ID instead. (staticcheck)
	apps.SubjectBotJoinedChannel,
	^
```

server/proxy/notify.go
hanzei: I'm not sure if that would make the code easier to read. We might add additional code in the future that handles more cases.

hanzei (approved):
I've tested the changes with both the `bot_(joined|left)_(channel|team)` and `user_(joined|left)_(channel|team)` using the test app. Expandtions for teams and channel now work again. LGTM :rocket:

Issue #469, opened 2023-04-03
Remove reference to rudder key in ci workflow, keep in cd workflow


mickmister:
Using the prod Rudderstack telemetry key in CI can skew our telemetry numbers. We should make it so we use the development key instead. The production builds of the plugin produced by the `cd.yml` should still contain the prod key.

PR #470, opened 2023-04-03
Remove reference to prod rudder key in ci workflow, keep in cd workflow

mickmister:
#### Summary

Using the prod Rudderstack telemetry key in CI can skew our telemetry numbers. The production builds of the plugin produced by the `cd.yml` should still contain the prod key. This PR We makes it so we use the development key instead for the `ci.yml` workflow.

## Ticket Link

Fixes https://github.com/mattermost/mattermost-plugin-apps/issues/469

### DIFF
--- a/.github/workflows/cd.yaml
+++ b/.github/workflows/cd.yaml
@@ -15,6 +15,7 @@ on:
 env:
   TERM: xterm
   GO_VERSION: 1.19.6
+  MM_RUDDER_PLUGINS_PROD:  ${{ secrets.MM_RUDDER_PLUGINS_PROD }}
 
 jobs:
   build:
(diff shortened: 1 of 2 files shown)


trilopin (approved):
Nice!

toninis (approved):
👍🏼

toninis (changes requested):
You need to add the environmental variable on the CD pipeline from the secrets in order to accessible . 
It is not accessible by default 
```yaml
env:
  MM_RUDDER_PLUGINS_PROD:  ${{ secrets.MM_RUDDER_PLUGINS_PROD }}
```

Since with https://github.com/mattermost/mattermost-plugin-apps/pull/468 you expect it to exist

toninis (approved):

hanzei (approved):

PR #472, opened 2023-05-16
[MM-51931] Don't send channel_created event for DMs

hanzei:
#### Summary
If a newly created channel is a DM, the `ChannelHasBeenCreated` hook gets called with an empty `TeamID`. This leads to an error down the stack. Now, `ChannelHasBeenCreated` gets ignored for DMs.

I did try adding a test, but the only symptom of the bug is a log message (https://github.com/mattermost/mattermost-plugin-apps/blob/22741daa176f919160193831d3bd2f113aa369c8/server/proxy/notify.go#L225-L229) which is hard to test for.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-51931

### DIFF
--- a/server/proxy/notify.go
+++ b/server/proxy/notify.go
@@ -195,6 +195,12 @@ func (p *Proxy) NotifyUserTeam(member *model.TeamMember, actor *model.User, join
 // NotifyChannelCreated handles plugin's ChannelHasBeenCreated callback. It emits
 // "channel_created" notifications to subscribed apps.
 func (p *Proxy) NotifyChannelCreated(teamID, channelID string) {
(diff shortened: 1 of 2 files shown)


mickmister (approved):

hanzei:
@DHaussermann I think the best way to test this PR is to check the logs for the error message linked in the ticket and confirm it does not longer show up

DHaussermann (approved):
Tested and passed
- Tested prod release and found a simple repo step of creating a DM channel. Opening a new DM with a user would log the error
- Upgraded to this PR resolved the issue 
- Brief regression testing showed nothing unusual in the logs

 LGTM!

PR #480, opened 2023-09-09
[MM-53015] Allow the user_agent field to be included in context

kyeongsoosoo:
#### Summary

When the `InvokeCall` method is executed, `AppContext` is assigned an expanded context.
https://github.com/mattermost/mattermost-plugin-apps/blob/89a1e4c1fd9e784a52129266e7021b72fc813100/server/proxy/invoke_call.go#L185-L190

I found that the `user_agent` field disappears because `expandContext` method initializes the `UserAgentContext` when it runs.
I changed the `user_agent` and `track_as_submit` fields to keep their original values on initialization.
https://github.com/mattermost/mattermost-plugin-apps/blob/89a1e4c1fd9e784a52129266e7021b72fc813100/server/proxy/expand.go#L216-L217

#### Ticket Link

Fixes https://github.com/mattermost/mattermost/issues/23621
JIRA https://mattermost.atlassian.net/browse/MM-53015


mickmister (approved):

hanzei (approved):
Thanks for the contribution and also for adding tests :+1:

